PR #44, opened 2016-05-26
upgrade to latest fabric8 version which has ingress support

jstrachan:
upgrade to latest fabric8 version which has ingress support


jstrachan:
[merge]

PR #62, opened 2016-06-01
fix broken dependencies lets avoid possible bad combinations of kubernetes-api and kubernetes-client versions; as we can let kubernetes-client be brought in via transient dependencies

jstrachan:
fix broken dependencies
lets avoid possible bad combinations of kubernetes-api and kubernetes-client versions; as we can let kubernetes-client be brought in via transient dependencies


rhuss:
[merge]

PR #84, opened 2016-06-03
Refactor enricher configuration.

rhuss:
Instead of using String keys sprinkled around all over the code,
typesafe enum are used now. These are defined as inner classes in the
enricher, and provide an overview as well as type safety. Default
values are supported, too.

By accident it forces a certain naming convention. It could be that
this is to strict, but we can relax the constraints later, too.


jstrachan:
[merge]

PR #85, opened 2016-06-03
fixes #46 so that we default to Deployment now; though you can disable it and it'll use RS; which you can disable too if need be

### LINKED ISSUE #46: for folks not using yaml files we should provide a configuration option on the maven plugin to specify if Deployment/DeploymentConfig/ReplicaSet/ReplicationController should be generated
we should default to Deployment OOTB I would think really as its the k8s standard and has rolling upgrades baked in etc

jstrachan:
fixes #46 so that we default to Deployment now; though you can disable it and it'll use RS; which you can disable too if need be


jstrachan:
[merge]

PR #89, opened 2016-06-03
lets allow the customizer to be combined with custom build configuration if a configuration is set. Also lets add typesafe Config support for customizer like we do for enhancer

jstrachan:
lets allow the customizer to be combined with custom build configuration if a configuration is set. Also lets add typesafe Config support for customizer like we do for enhancer


rhuss:
[merge]

PR #90, opened 2016-06-03
lets expose the docker user, name and label as maven properties so we can refer to them inside custom <build> configurations

jstrachan:
lets expose the docker user, name and label as maven properties so we can refer to them inside custom <build> configurations


jstrachan:
[merge]

PR #91, opened 2016-06-03
Refactored location of resource files

rhuss:
Now all generated resource files are stored within META-INF/fabric8

* Two file "kubernetes.yml" and "openshift.yml"
* Two directories "kubernetes" and "openshift" holding the individual resource items.


rhuss:
[merge]

PR #92, opened 2016-06-03
lets avoid using the image name docker property by default (we could maybe enable it with a setting later?)

jstrachan:
lets avoid using the image name docker property by default (we could maybe enable it with a setting later?)


jstrachan:
[merge]

PR #93, opened 2016-06-04
Refactor config handling again.

rhuss:
Make it look a bit smarter. Also moved enricher name intitialization to the super constructor.


rhuss:
[merge]

PR #94, opened 2016-06-04
Further refactoring

rhuss:
* Cleaned up support package to extenvvar package. Not sure whether it's still needed (its about loadingn external json schema files). Can be safely deleted now.
* Harmonized class lookups in IconEnricher, slimmed down EnricherContext a bit. The context should not grow without reason to provide enrichers the minimal environment they need to operate


rhuss:
[merge]

PR #95, opened 2016-06-04
Add a new phase for enrichers

rhuss:
So enrichers can

* addDefaultResources() when there are resources missing in a KubernetesList
* adapt() at the end of the processing to do some fine tuning on the generated list (not yet used yet)

Maybe one should detangle the 'enriching' and the 'defaulting' into two separate interfaces. But lets wait for now how this evolves.


rhuss:
[merge]

PR #97, opened 2016-06-04
fixes #96 so that we can define some maven properties for docker user, image name part and label that can be reused on custom side car container configuration builds and kubernetes manifests. Also made the use of 'latest' for snapshots configurable

### LINKED ISSUE #96: when creating sidecars (e.g. funktion using nodejs) having maven properties for the user, name and label of the docker image is handy
so that you can use the same user and label for your sidecar containers. We generally can't just use `groupId` and `artifactId` as-is since they need to be modified slightly to be valid docker names. So lets expose those properties so folks can use those properties safely in sidecar image names

jstrachan:
fixes #96 so that we can define some maven properties for docker user, image name part and label that can be reused on custom side car container configuration builds and kubernetes manifests. Also made the use of 'latest' for snapshots configurable

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/DockerUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/DockerUtil.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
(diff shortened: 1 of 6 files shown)


jstrachan:
[merge]

PR #98, opened 2016-06-04
fixes #77 to allow timestamp docker labels to be used for snapshot builds

### LINKED ISSUE #77: when building a docker image with a SNAPSHOT version, it would be nice to replace "1.0-SNAPSHOT" with a timestamp in docker
if you use `1.0-SNAPSHOT` as the docker image name, then deploy it with `mvn fabric8:deploy` then the second time you build you often get docker failures due to the image label being in use.

it'd be nice if OOTB the plugin was smarter & defaulted the docker label to a timestamp label if the `${project.version}` ends with `SNAPSHOT`; so that the same timestamp is exposed in the kubernetes resource as is created via `fabric8:build`

jstrachan:
fixes #77 to allow timestamp docker labels to be used for snapshot builds


rhuss:
[merge]

PR #100, opened 2016-06-06
added liveness and readiness check generation for spring boot if the actuator is enabled

jstrachan:
added liveness and readiness check generation for spring boot if the actuator is enabled


jstrachan:
[merge]

PR #102, opened 2016-06-06
fixes #101 to use the actual docker label if its different from project.version

### LINKED ISSUE #101: lets use the docker label used in the generated kubernetes resources rather than the project.version if they are different
this is particularly important when using timestamp labels

jstrachan:
fixes #101 to use the actual docker label if its different from project.version


jstrachan:
[merge]

PR #104, opened 2016-06-08
lets let the alias be configured for the spring boot customizer

jstrachan:
lets let the alias be configured for the spring boot customizer


rhuss:
[merge]

PR #113, opened 2016-06-09
fixes #112 to link to the site URL for the mvn pom based on the current version

### LINKED ISSUE #112: add support for the `fabric8.io/docs-url` annotation to link to the generated site url based on the maven pom's distributionManagement element
which if its a kubernetes service, resolve to the external URL. Then replace any `${project.artifactId}` or `${project.verison}` values so even if the site is hosted outside of kubernetes it works fine

jstrachan:
fixes #112 to link to the site URL for the mvn pom based on the current version


jstrachan:
[merge]

PR #117, opened 2016-06-10
upgrade fabric8, add offline mode for enricher, use relative grafana link and reuse the fabric8 annotation constants

jstrachan:
upgrade fabric8, add offline mode for enricher, use relative grafana link and reuse the fabric8 annotation constants


jstrachan:
[merge]

PR #123, opened 2016-06-21
first spike at a a fabric8:watch mojo. Needs some work mind ;)

jstrachan:
first spike at a a fabric8:watch mojo. Needs some work mind ;)


jstrachan:
[merge]

PR #127, opened 2016-06-21
lets make the default to use timestamp for docker images for snapshot projects; otherwise `mvn install` will barf if you have already run `mvn fabric8:deploy`

jstrachan:
lets make the default to use timestamp for docker images for snapshot projects; otherwise `mvn install` will barf if you have already run `mvn fabric8:deploy`


jstrachan:
[merge]

PR #128, opened 2016-06-22
lets try auto-update fabric8 + docker-maven-plugin versions as we release

jstrachan:
lets try auto-update fabric8 + docker-maven-plugin versions as we release


jstrachan:
[merge]

PR #130, opened 2016-06-22
lets default to use service type LoadBalancer as it means services can be accessed from outside of the cluster which is usually want folks want. Its easy to change the default too to something else

jstrachan:
lets default to use service type LoadBalancer as it means services can be accessed from outside of the cluster which is usually want folks want. Its easy to change the default too to something else

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/enricher/DefaultServiceEnricher.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/enricher/DefaultServiceEnricher.java
@@ -57,7 +57,7 @@ public class DefaultServiceEnricher extends BaseEnricher {
         headless {{ d = "false"; }},
 
         // Type of the service (LoadBalancer, NodePort, ...)
-        type;
+        type {{ d = "LoadBalancer"; }};
 
         public String def() { return d; } protected String d;
     }


jstrachan:
[merge]

rhuss:
But isn't a LoadBalancer only available in managed environments like GKE ? What happens when a there is no LoadBalancer facility available ? Is it silently ignore or does it croak ?

jimmidyson:
Silently ignored.

I'm not sure it should be a default because I don't believe it is something that everyone wants by defaut, especially when you're running in a cloud env where you might have to pay for ELB instances/rules. I would also expect only certain services to be accessed externally, the majority of a (large) deployment are likely to be in-cluster only, but that might be worth discussing with Kubernetes/OpenShift users a bit more to see what reqs they have.

jstrachan:
out of all our examples, quickstarts and apps there's a tiny percentage of them which don't want external access; its very very common to make REST APIs and web apps that are intended to be accessed via a browser. 

Whatever value we choose its wrong for some folks. I'd rather the default be what most folks want (e.g. 95% of all our examples & quickstarts); then for those 5% of cases folks don't want to use a pesky web browser to work with their code, they can set this one flag ;)

jstrachan:
Right now we automate the creation of Ingress or Routes if its LoadBalancer. Many environments just ignore LoadBalancer (e.g. vanilla Kubernetes & OpenShift). Its only GKE which does use it to make it accessible. For more background see: https://github.com/openshift/origin/issues/2026

jimmidyson:
I'm not sure that our examples are the best thing to base real-world usage on :) Setting a default like this now becomes more difficult to revert in future without breaking users setups. I'd prefer to do more asking around to validate this is the right default setting, perhaps on the Kubernetes mailing list? The fact that it's not a default in the Kubernetes API, which has many sane defaults now, points me towards not using it as a default in our plugin, but that may be because they're more conservative, not sure.

jstrachan:
We're making a microservices platform; for folks to make microservices. Folks usually use REST and browsers to work with them. So making things actually work as developers expect them to (via their browser); seems like a sensible approach to me.

Its trivial single property in a pom.xml to change though if folks wanna disable it. Or folks can write the service YAML template if they prefer to opt out of the default generated Service.

jimmidyson:
In a microservices architecture it is certainly not usual to expose _every_ service to  cluster external clients. I can't give hard numbers but from what I've seen the majority of microservices provide a few externally accessed services that route/load balance/orchestrate across multiple cluster-internal services.

As, IMO, a great example of a microservice architecture, Feed Henry has a few services externally accessible which route to the internal services, of which there can be a lot. This is heavily skewed to the side of internal microservices that wouldn't require an external load balancer to be set up, only a few requiring that. This is quite a regular architecture from what I've seen.

rhuss:
I agrree with @jimmidyson that a default should be very carefully chosen, since this changing defaults later is the most awkward thing /wrt to backwards compatibility. If there is no good default, then maybe its better to avoid a default at all ? (afaik 'Type' is not mandatory for a service).

@jstrachan yes, but Microservices talk to each other, only a minority speak to the 'outside' (e.g. the UI stuff). Its also a matter of security, trust relationships etc. And for testing you can always use 'kubectl proxy'. Also if we think about production envs, it doesn't make sense to expose all Mircroservices per default. 

Personally I tend to be a bit more conservative (i.e. would prefer no default).

rhuss:
This boild down, whether we should generate our resource descriptors in a generic fashion, applicable for development and production (and then production will dictate some restrcitions), or whether we should focus on a single use case. E.g. for development, production needs to create then their own descriptors.

However we could introduce something like 'styles', where e.g. a 'development style' would choose default more appropriate by a development workflow. However, since the descriptors are also considered to be (immutable) artefacts, we add an extra level of complexity here.

It is an architectural decision what to expose as a service and what. For a default we should be sure that really 90% or so of all  development an production uses cases needs this. And we should big warning signs, that the service get exposed by default. pretty sure many will oversee this.

Better safe than sorry ?

rhuss:
Another suggestion: I think for customizer, which are very opinionated per se (e.g. selecting the base image for you e.g.), its perfectly to make a default decision. So maybe we can put this in for the spring-boot customizer ? This is meant to be a starting point anyway where we make the decisions (which later then can be overwritten).

For the `DefaultServiceEnricher` this is a bit to hard, since we would add a `LoadBalancer` even when the use didn't specify any type. Only when he specifies a different type its not used.

PR #140, opened 2016-07-08
lets use the fabric8-maven-dependences project to update all the quickstarts

jstrachan:
lets use the fabric8-maven-dependences project to update all the quickstarts


jstrachan:
[merge]

PR #141, opened 2016-07-08
simplify the PR configuration so its a little clearer how things differ across projects (e.g. parent pom)

jstrachan:


jstrachan:
[merge]

PR #156, opened 2016-07-12
Refactored git-enricher to build-enricher (as it also enriches the jenkins url)

rhuss:
Still open:

* When to switch on 'online mode' (which currently can be done with a global property fabric8.online) ?
* Rename GOGS_JENKINS_USER to GOGS_USER since the jenkins annotation doesnt require a user part ?


rhuss:
LGTM

rhuss:
@jstrachan @rawlingsj if you have some time to spend, could you please give a short ok/nok to this PR's implication that if applied, the Jenkins CD would need to set a `-Dfabric8.online` when running the builds in order to extract live data like the gogs or jenkins service URL. See also #157 for a more thorough discussion on this. thx ;-) 

I would love to merge this in to avoid constantly rebasing while left open.

rhuss:
Going to merge this PR with online still being the default. Need this merged to avoid merge conflicts on the stuff i'm starting to work now.

rhuss:
LGTM

rhuss:
LGTM

rhuss:
[merge]

jstrachan:
I wonder if we can find a more automatic way to know a build is being run inside Jenkins - e.g. via a Jenkins environment variable?

rhuss:
For our opinionated enrichers this is probably a good idea. Is there are well known env var which is always set when running in a CD build ?

jstrachan:
There is when not using jenkins and/or kubernetes pipeline plugins; haven't had chance to check what env vars are visible inside a kubernetes pipeline block yet. e.g. BUILD_ID
https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-JenkinsSetEnvironmentVariables

jstrachan:
I don't have a working fabric8 right now; still fighting with OAuth and minishift; but you could try something like this...

println "BUILD_ID = ${env.BUILD_ID}"

inside the `kubernetes.pod... { }` block in a Jenkinsfile and see if that prints anything out?

jstrachan:
I just confirmed that inside a Jenkins pipeline build, outside and inside the `node {}` block as well as inside a kubernetes build pod, the env vars are valid for `BUILD_ID` and `BUILD_NUMBER` so checking either of those values would be a nice way to default to a CI/CD build or not. Folks can explicitly enable it via configuration/property otherwise we default to on if those env vars are set

rhuss:
Good, idea, lets do it that way. Hopefully there are not many use case with running `fabric8:resource` in a CI outside of a Jenkins context.

Maybe for the default we could also check whether a login is possible. If `fabric8.online` is not set explicitly, the use online enrichment only when `BUILD_ID` is set **and** a connection could be made to a k8s cluster. Otherwise, siltently don't do the enrichment. One could always make it more concrete with setting the `fabric8.online` property, where the build will then fail if a connection cannot be made (and online mode has been switched on).

Issue #157, opened 2016-07-12
When to switch on / off online mode for enrichers ?


rhuss:
I did a refactoring of the GitEnricher since he constantly throws errors because trying to contact a running Kubernetes Cluster.

IMO the default should be offline, which could be switched on by a CD. The reason is, that enriching the resource descriptors with service urls coming from random clusters where one is currently logged in, and which end up in an jar-artefact, too, is not really good for versioning and reproducability (you can't recreate an artefact later if you are not running in the very same environment).

It's ok when running in a CD in order to store meta data about the current build because this artefact gets pushed around the pipeline (even then, one need to give up the dream that a single GIt tag is enough to reproducably recreate a piece of software). 

So my idea is to start in offline mode and let the CD switch it on with a property. This is currently possible by adding a `-Dfabric8.online` to the build scripts (if #156 gets merged in). Of course this can be set within the `pom.xml`, too for a permanent setting.

I introduced an `AbstractLiveEnricher` which does all the on / offline handling and is supposed to be the single entry point for contacting a cluster. 

I also renamed the GitEnricher to a BuildEnricher (since it adds also Jenkins meta data). 

Is this ok that we start offline by default ? This also avoids nasty errors and pauses when people start from scratch.

nicolaferraro:
+1: I just came up with that (closed) PR for changing the log messages because people were saying "I cannot just execute the `mvn spring-boot:run` to see the quickstart running locally, it requires some kubernetes stuff".

IMHO, the f-m-p should not even try to contact the local kubernetes cluster, unless it is going to deploy the app to a cluster.

rhuss:
I wonder whether there is fabric8 specific CD environment variable, which we could check for an online mode ? `BUILD_ID` is probably not good enough as it is set by **any** Jenkins build.

@rawlingsj maybe you have an idea how we could uniquely detect from within the plugin whether it is running in a CD run so that K8s service are accesible and we can extract the proper URL ?

rawlingsj:
I think `${env.JOB_NAME}` or `${env.BUILD_ID}` should be ok

rhuss:
@rawlingsj Yeah, but isn't `BUILD_ID` set for **any** Jenkins environment, whether it is running in Kubernetes or not ? The problem is, when running *online* then the enricher tries to detect the Service URL for Gogs (also a fabric8 special thing) and Jenkins.

[...]

PR #161, opened 2016-07-13
fixed log messages

nicolaferraro:
I just removed some forgotten log warnings (they can alert users more than necessary).


nicolaferraro:
I've seen @ro14nd refactored the code.

rhuss:
Yeah, sorry. Just waiting about some discussion whether to switch off the online-mode by default (and only enable it if `fabric8.online` is set which can be easily done within a CD). IMO by default the creation for the resource descriptors should reach out to a running cluster for querying information since this can make the build unreproducible when running against a different cluster.  

See #156 for a discussion about this.

rhuss:
and also #157

PR #162, opened 2016-07-14
lgtm.co integration

rhuss:


jimmidyson:
An alternative to the MAINTAINERS file is to create a Maintainers team at the org level. What do you think?

rhuss:
Both are ok, I guess. MAINTAINERS is good because you quickly find the people behind the project, not so good because of spam (but well, we are naked anyway).

Maintainer team on the org level is easier to change and add new maintainers.

tbh, I'm not sure ;-). any bias from your side ?

jimmidyson:
Would prefer org team personally, don't like the idea of keeping MAINTAINERS files in sync across multiple repos (& orgs...).

rhuss:
Ok, so lets do it ;-). I will remove the MAINTAINERS file now.

rhuss:
At the end the merge must then done manually, right ? Either by pushing the button or by adding a '[merge]' comment ?

jimmidyson:
Yeah LGTM just adds a required status check so PRs can't be merged until added (it's a very simple idea/service). Admins can override it & merge if need be but it just adds another requirement to review.

PR #164, opened 2016-07-14
Add dependency enricher to help bundle resources from dependencies

jimmidyson:


enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+        dependencyArtifacts.add(artifact.getFile());
+      }
+    }
+  }
+
+  @Override
+  public void adapt(KubernetesListBuilder builder) {
+    for (File artifact : dependencyArtifacts) {
+      try {
+        URL url = new URL("jar:file:" + artifact.getAbsolutePath() + "!" + DEPENDENCY_KUBERNETES_YAML);
```
rhuss: This is a perfectly valid way to read the resource file. I wonder though, whether we should harmonize it with the way how the `fabric8-enricher` and `fabric8-generator` descriptors are loaded in [PluginServiceFactory](https://github.com/rhuss/fabric8-maven-plugin/blob/9f70907d008479decae6bc2d33dfc861065c632c/core/src/main/java/io/fabric8/maven/core/util/PluginServiceFactory.java#L93-L100) where this happens over a direct classpath lookup. Don't whether there might be any difference, but at least it would fail consistently ;-)

Of course, a classpath lookup would always include the transitive dependencies, so going over files (except for path separator issue, which already lurking around ;-) like here is maybe the better way ?

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  protected boolean isIncludeTransitive() {
+    return Configs.asBoolean(getConfig(Config.includeTransitive));
+  }
+
+  // ====================================================================================================
+
+  // Available configuration keys
+  private enum Config implements Configs.Key {
```
rhuss: Minor: The other enrichers have their config at the top so one quickly can see the options on a first glance. To avoid to much visual boilerplate they are also compacted and formatted a bit unusual. Maybe we should move this at least up to the top, too, or we move the config from the other enrichers to the bottom, so that our enrichers have a common 'style'

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
jimmidyson: Are dependency artifacts in the classpath at this pointin the lifecycle? Didn't know they were. I prefer the reading from filesystem anyway tbh as it allows us to not include transitive deps if desired.

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
jimmidyson: This is intellij formatting... putting inner classes/enums at the bottom of the file. Don't mind which way round you prefer tbh so will move to top & format as per other files.

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
rhuss: Ok, lets leave it like this. Just a bit unsure when `file.getAbsolutePath()` returns a Windows path with `\` whether they work as intended in an URL or whether they should be converted to `/`.  I dont know how hard it would be to write a unit test, so we could verify this when building on Windows.

We might consider to take this over later for the enricher and generator lookup, too.

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
rhuss: Thanks ! ;-)

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
jimmidyson: That's a good point. I'll investigate more.

jstrachan: you can annotate the mojo to say that dependencies should be resolved IIRC so that you can do the classpath scanning stuff.
```java
@Mojo(requiresDependencyResolution = ResolutionScope.COMPILE, ...
```

jimmidyson: Seems like `file.toURI().toURL()` should work well cross platform (I hope!). Changed to this,

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
jimmidyson: Done.

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
```diff
@@ -0,0 +1,88 @@
...
+        dependencyArtifacts.add(artifact.getFile());
+      }
+    }
+  }
+
+  @Override
+  public void adapt(KubernetesListBuilder builder) {
+    for (File artifact : dependencyArtifacts) {
+      try {
+        URL url = new URL("jar:" + artifact.toURI().toURL() + "!" + DEPENDENCY_KUBERNETES_YAML);
```
rhuss: thanks ;-) (strangely, when doing a rebase with forced push, all comments are lost, too. Github should at least keep them as reference)

rhuss:
lgtm

rhuss:
[merge]

enricher/dependency/src/main/java/io/fabric8/maven/enricher/dependency/DependencyEnricher.java
jimmidyson: I normally don't squash & force push until review is complete but thought this was going to be fine anyway - next time won't squash/force push until review complete.

PR #179, opened 2016-07-23
Fixed issue with build mode and did some minor refavtorings

rhuss:
* 'fabric8.build.mode' is now 'fabric8.mode' as it can be reused in other circumstances, too
* 'fabric8.recreate' is an enum for specifying what should be recreated.


rhuss:
LGTM

rhuss:
[merge]

PR #180, opened 2016-07-23
Tuned BuildEnricher with respect to offline mode

rhuss:
TODO: There should be a more specific datum to decide whether the build is running in an Fabric8 CD environment or not.
BUILD_ID is to generic as it is set for every Jenkins build.


rhuss:
LGTM

rhuss:
[merge]

PR #198, opened 2016-07-29
Addition Refactorings

rhuss:
* Renamed "java-app" to "java-exec" so that it matches the java:exec goal of the maven-java-plugin
* Cleaned up generator names to not contain "."


rhuss:
LGTM

rhuss:
[merge]

Issue #146, opened 2016-07-09
customizer to auto-detect executable jars OOTB


jstrachan:
if we have an executable jar we should default the docker command accordingly without the user having to add any explicit configuration. Also if we have a vanilla executable jar (without Spring Boot or WF Swarm) we should enable the docker image being created; using artifact-with-dependencies etc.

e.g. this example https://github.com/fabric8-quickstarts/vertx-simple should not require the main class to be provided in the fabric8-maven-plugin configuration since it can be deduced/used via the executable jar.

This issue is similar to #143 and this example https://github.com/fabric8-quickstarts/swarm-camel though with WF Swarm its also a fat jar too

To deal with executable jars versus Spring Boot versus WF Swarm; we may need a way to prioritise customerisers; e.g. Spring Boot customisation before 'executable jar' customisation?

rhuss:
Technically you can already prioritise customisers / enrichers by adding an order in the `fabric8-default-customizers` manfifest. It's just a matter of the order.

We could also add a sort of `break` method in the Customizer interface, which stops looking into customizer further down. E.g. when a spring boot customizer has detected to run then it could stop any other customization (which could break things).

The plain java app customizer could detect the main  class:

* Checking the source for a `public void static main()` and when it founds a unique one, take this.
* Check for a `maven-execute-pluging` config and use a main class from there

For `spring-boot` and `widlfly-swarm` we should not need to extract a main class at all, as the base image should be able to start this without the need of a main class.

btw, I really like to move 'customizer' to 'starter' ;-)

rhuss:
I added support for auto-detection of a Main-Class in `target/classes`.

However, this wont found MainClasses from dependencies, like `org.apache.camel.cdi.Main`.
I wonder whether we should add some sort of white-list to detect these classes, too ? Or we could even add a dedicated **Camel** detector.

jstrachan:
Yeah, thats not a bad idea really. Though I guess its not a biggie to have custom configuration for the main in cases where we're not producing an executable jar

rhuss:
Ok. My next steps here are:

* Auto-detect 'fat' jars. E.g. by checking for the maven-shade-plugin or even examining the jars below `target/` by reading their manifest. This would also adapt the assembly configuration to include this jar, which would be nice because it will work with wildfly-swarm, too (wildfly-swarm is a bit unorthodox as it doesn't add its repackaged jar to the reactor, like e.g. spring-boot)

[...]

PR #204, opened 2016-08-01
. #146 : Add detection of a single MainClass

rhuss:
This works currently for flat classpath apps. Only the classes in target/classes are examined.
I wonder whether certain lib jars should be examined, too.


rhuss:
LGTM

rhuss:
lgtm

rhuss:
[merge]

PR #205, opened 2016-08-01
Fix fabric8:push with generated image configs (like in the quickstarts)

rhuss:
Still open is the handling with the configuration properties where they are all prefixed with 'docker.' which would be niced to be replaced with 'fabric8.'


rhuss:
lgtm [merge]

PR #209, opened 2016-08-02
Add S2I Binary Builds + auto detection of the build mode

rhuss:
* Support for S2I binary builds
* fabric8.mode is by default now "auto" which tries to detect the environment
* sprng-boot's assembly fixed
* fabric8.build.mode can be now "s2i" (default) or "docker" (for a binary docker build)


rhuss:
LGTM [merge]

PR #226, opened 2016-08-08
Allow for a default value for `fabric8.build.recreate` when no value given

rhuss:
It that case everythin is recreated.


rhuss:
lgtm [merge]

PR #234, opened 2016-08-09
fixes #233 and adds a spring-boot-with-yaml sample to demonstrate it in action

### LINKED ISSUE #233: support enrichment of Deployment's PodTemplate metadata while reusing the default, say, spring boot generator
we've a use case for hystrix where we want to specify labels for the podTemplate in a Deployment; but inherit all the generated metadata for the PodSpec etc. Right now that doesn't work for Deployments with missing PodSpec

jstrachan:


jstrachan:
lgtm [merge]

PR #242, opened 2016-08-10
Worked on documentation, refactored enricher modules

rhuss:
There are now two groups of enricher which we deliver

* `fmp-` : Standard enrichers useful for creating default objects and annotations
* `f8-` : fabric8 specific enrichments used in the Fabric8 consiole and Fabric8 CD


rhuss:
lgtm [merge]

PR #247, opened 2016-08-11
fixes #246

### LINKED ISSUE #246: create helm goal to create a helm chart for the current maven project

jstrachan:


plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+        File templatesDir = new File(outputDir, "templates");
+        templatesDir.mkdirs();
+        try {
+            Files.copy(sourceDir, templatesDir);
+        } catch (IOException e) {
+            throw new MojoExecutionException("Failed to copy manifest files from " + sourceDir + " to chart templates directory: " + templatesDir + ". Reason: " + e, e);
+        }
+
+        File outputChartFile = new File(outputDir, "Chart" + HELM_YAML_EXTENSION);
```
rhuss: Is it the normal convention to uppercase 'Chart.yaml' ?

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
```diff
@@ -0,0 +1,248 @@
...
+        Chart chart = createChart();
+        try {
+            KubernetesHelper.saveYaml(chart, outputChartFile);
+        } catch (IOException e) {
+            throw new MojoExecutionException("Failed to save chart " + outputChartFile + ". Reason: " + e, e);
+        }
+
+        File basedir = null;
+        if (project != null) {
```
rhuss: will never be null. And if so, an NPE would be better than silently ignoring it. I just would leave out the check.

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
```diff
@@ -0,0 +1,248 @@
...
+            basedir = project.getBasedir();
+            if (basedir != null) {
```
rhuss: dito

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
```diff
@@ -0,0 +1,248 @@
...
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    protected static File copyTextFile(File sourceDir, FilenameFilter filter, File outFile) throws IOException {
+        File[] files = sourceDir.listFiles(filter);
+        if (files != null && files.length == 1) {
```
rhuss: what if there are more than one README or LICENSE ? Shouldnt we pick one ? At least we should add a warning. The return value is ignored,  btw.

rhuss:
If you dont mind, I would merge it and hackup the minor issues. a small relief from writing docs and talks ;-)

rhuss:
lgtm

rhuss:
[merge]

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
jstrachan: yes - see https://github.com/kubernetes/helm/blob/master/docs/charts.md#the-chart-file-structure

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
jstrachan: it can be null if you run a maven goal in a folder that has no pom.xml. That may sound odd to a maven user, but mvn goals can be used inside (say) a nodejs project even if there's no pom.xml

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
jstrachan: yeah, might be worth adding a warning

plugin/src/main/java/io/fabric8/maven/plugin/HelmMojo.java
rhuss: If you call `mvn fabric8:helm` in a directory without pom.xml, then it will croak, that it cant find this plugin. It will even croak when there _is_ a pom.xml but not fabric8-maven-plugin configured. Even if you use a full qualified plugin name, this wont work because by default the Mojo property "requiresProject"  is true. You have to set this explicitly to `false` to allow a Mojo goal to run without a project.

So imo a project in our case will always be set if the plugin comes until this point.

jimmidyson: You *could* run it using fully qualified plugin gid:aid:version:goal but who's gonna do that if they're not using maven?

rhuss: Yes, you could, but it will stop before running the Mojo because this Mojo declares that it *requires* a pom.xml. See also `requiresProject` in http://maven.apache.org/ref/3.2.1/maven-plugin-api/plugin.html which is by default `true`. If you want to allow a goal to be run standalone (without a pom.xml and hence without a `project`) you have to set this to false in your `@Mojo`. The generate `HelpMojo`  does this.

PR #268, opened 2016-08-17
fixes #267 to add a `fabric8:distro` goal for making distros of kubernetes manifests

### LINKED ISSUE #267: create a distro maven goal
gofabric8 uses distros (zips of kubernetes manifests) to be able to download and install all the various OpenShift Templates / ConfigMaps for apps so we can easily run them in the UI. Long term we'll probably move to helm, but to support gofabric8 and to make an easy to consume zip of kubernetes manifests; lets allow projects to combine all the kubernetes / openshift yaml files into a simple zip file.

Then in the distro maven project we can execute goal `fabric8:distro` and it'll walk the dependency tree and make a templates.zip file containing the main packages and individual microservices. The main difference from the fabric8-maven-plugin 2.x approach will be: 

* we include both `openshift` and `kubernetes` folders in the root directory
* we use YAML files instead of JSON

other than that it should be easy to patch tools like gofabric8 to use the new distros

jstrachan:


jstrachan:
lgtm [merge]

PR #270, opened 2016-08-18
try use project classpath for enrichers when enabled...  for https://…

christian-posta:
…github.com/fabric8io/fabric8-maven-plugin/issues/269

interested in a review from @rhuss and @jstrachan 

still WIP

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/ClassUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/ClassUtil.java
@@ -18,9 +18,13 @@ package io.fabric8.maven.core.util;
 
 import java.io.*;
 import java.net.URL;
+import java.net.URLClassLoader;
 import java.util.*;
 
+import io.fabric8.maven.docker.util.Logger;
 import javassist.*;
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.maven.project.MavenProject;
 
 /**
  * @author roland
@@ -29,13 +33,33 @@ import javassist.*;
 public class ClassUtil {
 
     public static Set<String> getResources(String resource) throws IOException {
+        return getResources(resource, null);
+    }
+
+    public static Set<String> getResources(String resource, List<ClassLoader> additionalClassLoaders) throws IOException {
+
+        ClassLoader[] classLoaders = mergeClassLoaders(additionalClassLoaders);
+
         Set<String> ret = new HashSet<>();
-        for (ClassLoader cl : getClassLoaders()) {
+        for (ClassLoader cl : classLoaders) {
             Enumeration<URL> urlEnum = cl.getResources(resource);
             ret.addAll(extractUrlAsStringsFromEnumeration(urlEnum));
         }
         return ret;
     }
+
+    private static ClassLoader[] mergeClassLoaders(List<ClassLoader> additionalClassLoaders) {
+        ClassLoader[] classLoaders;
+
+        if (additionalClassLoaders != null && !additionalClassLoaders.isEmpty()) {
+            classLoaders = ArrayUtils.addAll(getClassLoaders(), additionalClassLoaders.toArray(new ClassLoader[additionalClassLoaders.size()]));
+        }
+        else {
+            classLoaders = getClassLoaders();
+        }
+        return classLoaders;
+    }
+
 
     private static ClassLoader[] getClassLoaders() {
         return new ClassLoader[] {
@@ -52,9 +76,11 @@ public class ClassUtil {
         return ret;
     }
 
-    public static <T> Class<T> classForName(String className) {
+
+    public static <T> Class<T> classForName(String className, List<ClassLoader> additionalClassLoaders) {
+        ClassLoader[] classLoaders = mergeClassLoaders(additionalClassLoaders);
         Set<ClassLoader> tried = new HashSet<>();
-        for (ClassLoader loader : getClassLoaders()) {
+        for (ClassLoader loader : classLoaders) {
             // Go up the classloader stack to eventually find the server class. Sometimes the WebAppClassLoader
             // hide the server classes loaded by the parent class loader.
             while (loader != null) {
@@ -69,6 +95,7 @@ public class ClassUtil {
         }
         return null;
     }
+
 
     /**
      * Find all classes below a certain directory which contain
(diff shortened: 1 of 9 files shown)


christian-posta:
Still need to update docs. Should probably add this functionality for the Generators too, yah?

christian-posta:
k.. added the code for generators. will do the docs in a different PR; will try fill in the section about developing enrichers and the enricher API (5.2)

@rhuss thoughts?

christian-posta:
bump @rhuss

rhuss:
@christian-posta  ;-) will have a look early tomorrow.

core/src/main/java/io/fabric8/maven/core/util/ClassUtil.java
```diff
@@ -140,4 +167,27 @@ private static String convertToClass(String name, String prefix) {
         ret = ret.substring(0, name.length() - ".class".length());
         return ret.substring(prefix.length());
     }
+
+
+    public static URLClassLoader createProjectClassLoader(final MavenProject project, Logger log) {
+
+        try {
+
+            List<URL> compileJars = new ArrayList<>();
```
rhuss: would use already an array of URLs which would avoid the conversion later.

core/src/main/java/io/fabric8/maven/core/util/ClassUtil.java
```diff
@@ -140,4 +167,27 @@ private static String convertToClass(String name, String prefix) {
...
+
+            for (String element : project.getCompileClasspathElements()) {
+                compileJars.add(new File(element).toURI().toURL());
+            }
+
+            return new URLClassLoader(compileJars.toArray(new URL[compileJars.size()]),
+                    PluginServiceFactory.class.getClassLoader());
+
+        } catch (Exception e) {
```
rhuss: What kind of exception could occur here ? Should we really ignore it (i.e. only warn about) or shouldn't we break the build here because its supposed to be an internal error.

core/src/main/java/io/fabric8/maven/core/util/ClassUtil.java
```diff
@@ -140,4 +167,27 @@ private static String convertToClass(String name, String prefix) {
...
+            log.warn("Instructed to use project classpath, but cannot. Continuing build if we can: ", e);
```
rhuss: log.warn expects a String.format() specifier plus extra args:

```
log.warn("... %s ...", e.getMessage())
```

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
```diff
@@ -109,6 +108,12 @@
     private OpenShiftBuildStrategy buildStrategy = OpenShiftBuildStrategy.s2i;
 
     /**
+     * Should we use the project's compmile-time classpath to scan for additional enrichers/generators?
+     */
+    @Parameter(property = "fabric8.useProjectClasspath", defaultValue = "false")
+    private boolean useProjectClasspath = false;
```
rhuss: I would do it unconditionally since tbh I dont see a use case where we should not use the project classpath. In order to exclude / include enrichers one could still use a dedicated configuration for this. I'm a fan of adding config options only if absolutely needed (number of configuration options will explode anyway). Also save some documentation ;-)

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -81,6 +81,12 @@
     private File resourceDir;
 
     /**
+     * Should we use the project's compmile-time classpath to scan for additional enrichers/generators?
+     */
+    @Parameter(property = "fabric8.useProjectClasspath", defaultValue = "false")
+    private boolean useProjectClasspath = false;
```
rhuss: See above, wouldn't make it configurable but enable it always (as this was in the design anyway).

plugin/src/main/java/io/fabric8/maven/plugin/enricher/EnricherManager.java
```diff
@@ -54,6 +55,10 @@ public EnricherManager(EnricherContext enricherContext) {
         PluginServiceFactory<EnricherContext> pluginFactory = new PluginServiceFactory<>(enricherContext);
 
 
+        if (enricherContext.isUseProjectClasspath()) {
+            pluginFactory.addAdditionalClassLoader(ClassUtil.createProjectClassLoader(enricherContext.getProject(), enricherContext.getLog()));
```
rhuss: If we make the addition of the project classpath classloader unconditional then I would also add this class loader to the constructor of PluginServiceFactory to make the list of class loaders to use immutable after creation of the object.

rhuss:
@christian-posta Thanks a lot !

I wonder a bit though, how this interferes with building the real application as the custom plugin configurations are then also visible on the application classpath.

Maybe its better to add the enricher / generator dependencies directly to the plugin configuration as described in http://maven.apache.org/guides/mini/guide-maven-classloading.html ? (tbh, I didn't know that its possible to add dependencies to a plugin config). I think, that's the most elegant solution (if it works).

rhuss:
E.g.

```xml
 <plugin>
       <groupId>io.fabric8</groupId>
       <version>3.1-SNAPSHOT</version>
       <dependencies>
            <dependency>
                <groupId>myorg</groupId>
                <artifactId>my-custom-enricher</artifactId>
                <version>1.7.1</version>
            </dependency>
       </dependencies>
       ...
</plugin>
```

jimmidyson:
Agree. We shouldn't have build dependencies on project classpath - they can be specified as plugin dependencies.

rhuss:
@christian-posta is the solution to add the dependency to the plugin configuration acceptable for you ? If so, I would close this PR.

christian-posta:
@rhuss @jimmidyson no, not really.

What i'd like is to be able to include 3rd party dependencies for my project that implicitly know how to add kubernetes-specipic metadata if needed via enrichers and have the f-m-p automatically pick that up. For example, I want to include "spring-cloud-kubernetes-hystrix" for my project to provide hysteria functionality. But as part of that, a k8s service needs to be labeled with "hsytrix.enabled=true" for it to work properly and be picked up by the turbine dashboard. I already include the spring-cloud-kubernetes-hystrix artifact, i'd rather the f-m-p automatically adapt based on what enrichers it finds? Obviously I could add the labels manually myself, or explicitly add the dependency to the build plugin class path, but it just feels weird and is brittle; i originally brought the spring-cloud-kuberentes-hystrix dependency as a compile-time dependency for my project. Would be nice if f-m-p scanned that class path too.

rhuss:
@christian-posta I see your point, we should support both. So having a config option to switch this on and off makes sense then. A developer then can the declare that he want his code to be inspected, too. 

The question is whether we want to enable by default or not. Probably `false` is good default value in order to avoid unexpected surprises. 

Just thought also whether it makes sense to only inspected the classes of the current project, and not its dependencies by adding only `target/classes` to the classloader ?

rhuss:
@christian-posta ok, just re-read your use case, for which it makes sense to add the whole dependency.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
rhuss: fine with it now (see discussion below).

rhuss:
@christian-posta if you don't mind I would merge the PR now and then fix my comments (if there are no objections from your side).

rhuss:
rebased this and merged with a different PR

PR #276, opened 2016-08-19
unless explicitly configured via <configuration> or property, lets default to a sane deploy timeout for DeploymentConfigs that can work on minishift with slow wifi

jstrachan:


jstrachan:
lgtm [merge]

jstrachan:
[merge]

PR #296, opened 2016-09-01
Fixes #295, Default securityContext and replicas settings were not copied over

### LINKED ISSUE #295: Adding a src/main/fabric8/deployment.yml which only configures resource limits breaks default config
I tried adding a `src/main/fabric8/deployment.yml` which only configures the resource limits.  Here's the example:

https://github.com/chirino/spring-boot-camel/blob/resource-limits/src/main/fabric8/deployment.yml

The kubernetes.yml that gets generated when compared to what it generated before adding the file is missing the default `securityContext` and `replicas` settings.

chirino:
Configuration merging seems a bit brittle.  This is the second occasion
where a default setting was not properly merged in.  I’ve also added a
reflective method to help merge over simple fields to try to get this
robust.


jstrachan:
LGTM [merge]

PR #302, opened 2016-09-02
first spike of fabric8:setup

jstrachan:


doc/src/main/asciidoc/inc/_fabric8-setup.adoc
```diff
@@ -0,0 +1,38 @@
...
+[source, sh]
+----
+mvn io.fabric8:fabric8-maven-plugin:3.1.19:setup
+----
+
+Now your pom.xml should be modified to use the fabric8 maven plugin. So you should be able to now run the following:
+
+[source, sh]
+----
+mvn farbic8:run
```
rhuss: typo

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -246,4 +251,56 @@ public static void createArchive(File sourceDir, File destinationFile, JarArchiv
...
+                for (Map.Entry<Object, Object> entry : entries) {
+                    Object key = entry.getKey();
+                    Object value = entry.getValue();
+                    if (key != null && value != null) {
+                        groupArtifactVersionMap.put(key.toString(), value.toString());
+                    }
+                }
+            }
+        }
+        return groupArtifactVersionMap;
```
rhuss: Why not using the properties directly ? it's a string-string typed map anyway.

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -246,4 +251,56 @@ public static void createArchive(File sourceDir, File destinationFile, JarArchiv
...
+        if (version == null) {
+            LOG.warn("Could not find the version for groupId: " + groupId + " artifactId: " + artifactId + " in: " + map);
+        }
+        return version;
+    }
+
+    /**
+     * Returns the version from the list of pre-configured versions of common groupId/artifact pairs
+     */
+    public static String getVersion(String groupId, String artifactId, String defaultVersion) {
```
rhuss: for simplicity reasons I wouldn't allow a `defaultVersion`. If the version could not be found in the properties instead an internal error exception should be thrown to show prominently that something internally is wrong. I think its fair to assume that the version from `versions.properties` must be *always* loadable.

That also avoids to be forced to update the version in multiple places.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,268 @@
...
+            }
+        }
+    }
+
+    private boolean createOrUpdateFabric8MavenPlugin(Document doc) {
+        boolean updated = false;
+        String latestVersion = MavenUtil.getVersion(PLUGIN_GROUPID, PLUGIN_ARTIFACTID, PLUGIN_DEFAULT_VERSION);
+
+        // lets check that there's a property defined
+        String currentVersion = project.getProperties().getProperty(FABRIC8_MAVEN_PLUGIN_VERSION_PROPERTY);
```
rhuss: imo its better to add this property **only** when there is not yet a f-m-p configured because it could be the case that there is already a setup of f-m-p which does not use this very property (but e.g. a pluginManagement from a parent pom, or the property is defined in the parent pom).

Instead I would put this into `findOrAddPlugin()` when the plugin is actually added.

Adding a property this way is imo critical anyway, since its *non-local* in the sense that it happens in a different place in the pom.xml where the plugin is added. So when manually reverting the change the properties can be easily forgotten. tbh, I would add the version literally in the plugin config and let the use do the extraction in props if needed.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,268 @@
...
+            doc = XmlUtils.parseDoc(pom);
+        } catch (Exception e) {
+            getLog().error("Failed to parse pom " + pom + ". " + e, e);
+            throw new MojoExecutionException(e.getMessage(), e);
+        }
+
+        if (createOrUpdateFabric8MavenPlugin(doc)) {
+            updated = true;
+        }
+        if (updated) {
```
rhuss: would remove the tree lines above.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,268 @@
...
+            getLog().info("Updating the pom " + pom);
+            try {
+                DomHelper.save(doc, pom);
```
rhuss: don't forget the backup if requested ;-)

rhuss:
looks good ! (except some minor things, see comments). Really nice for a quickstart ....

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -246,4 +251,56 @@ public static void createArchive(File sourceDir, File destinationFile, JarArchiv
...
+            answer = defaultVersion;
+        }
+        return answer;
+    }
+
+    protected static Map<String, String> getGroupArtifactVersionMap() {
+        if (groupArtifactVersionMap == null) {
+            groupArtifactVersionMap = new HashMap<>();
+
+            InputStream in = MavenUtil.class.getResourceAsStream("versions.properties");
```
rhuss: I wonder whether it wouldn't be better to pick up the version we need from `META-INF/maven/${groupId}/${artifactId}/pom.properties` ? This is alway generated by Maven so we don't need any extra custom version properties. See also http://stackoverflow.com/questions/3697449/retrieve-version-from-maven-pom-xml-in-code

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
jstrachan: agreed - fixed

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
jstrachan: done!

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
jstrachan: yeah -using the pom.properties and failing if its not found is simpler

doc/src/main/asciidoc/inc/_fabric8-setup.adoc
jstrachan: fixed

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: done

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: whoops ;)

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: I was hoping we could use the `fabric8:setup` goal to upgrade older versions of the plugin (as updating the plugin version and using the latest executions etc). 

e.g. imagine if a project was using 3.1.1 then we introduced 3.1.29 which now uses the `helm` execution too. If we don't update the version, the project is broken as the new execution isn't available in the old version.

I think I'd rather the `fabric8:setup` add unnecessary versions which folks can remove (due to wishing to define them in a parent pom or via pluginDependency or whatever), than break builds.

As a compromise I've added a flag to disable updating of the plugin version if the plugin is already used. e.g.:
```
mvn io.fabric8:fabric8-maven-plugin:3.1-SNAPSHOT:setup  -DupdateVersion=false
```

will not update the version if its on an older version (or the version is defined somewhere else than the plugin element)

rhuss: but why not avoiding the property at all and update the `<version>...</version>` directly ?

tbh I thought about `fabric8:setup` as a one shot task for quick ramp up, but not as round-trip tool like forge. There are tons of things which could go wrong when updating or manipulating an existing setup. I.e. adding a new execution might be useful but could also be not wanted. And I don't think we should shield the user from touching the pom.xml because its not possible anyway and could raise false hope (tbh, I also dont think it should be the scope of this plugin).

A good update documentation is imo enough (as the user is already sold in using f-m-p when doing an update).

tl;dr - imo we shouldn't strive for supporting plugin updates but for one-shot setups only (with a no-op when there is already a f-m-p configured). But I'm open for discussions here.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+public class SetupMojo extends AbstractFabric8Mojo {
+
+    public static final String PLUGIN_GROUPID = "io.fabric8";
+    public static final String PLUGIN_ARTIFACTID = "fabric8-maven-plugin";
+    public static final String FABRIC8_MAVEN_PLUGIN_VERSION_PROPERTY = "fabric8.maven.plugin.version";
+
+    /**
+     * Should we update the version if the fabric8-maven-plugin is already used on a project?
+     */
+    @Parameter(property = "updateVersion", defaultValue = "true")
```
rhuss: Please prefix properties with 'fabric8.' as we do in the other properties, too.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+    private boolean updateVersion;
+
+    /**
+     * Whether we should use a version property for the plugin
+     */
+    @Parameter(property = "useVersionProperty")
+    private boolean useVersionProperty;
```
rhuss: Maybe we could use this as a String and name it `versionProperty` and, if the string is `null` then we insert the version literally. Also add a prefix `fabric8.` to the sys-property name.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Controls whether a backup pom should be created (default is true).
+     */
+    @Parameter(property = "generateBackupPoms")
```
rhuss: Please prefix with `fabric8.`

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+    private boolean generateBackupPoms;
+
+    @Parameter(name = "backupPomFileName", defaultValue = "${basedir}/pom.xml-backup")
```
rhuss: Please prefix with `fabric8.`

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+        } catch (IOException e) {
+            throw new MojoExecutionException(e.getMessage(), e);
+        }
+
+        // lets check that there's a property defined
+        String currentVersion = project.getProperties().getProperty(FABRIC8_MAVEN_PLUGIN_VERSION_PROPERTY);
+        String versionExpression = "${" + FABRIC8_MAVEN_PLUGIN_VERSION_PROPERTY + "}";
+        Element configuration = null;
+        Element fmpPlugin = findPlugin(doc, PLUGIN_GROUPID, PLUGIN_ARTIFACTID);
+        boolean pluginInserted = fmpPlugin == null;
```
rhuss: Wouldn't this be better called 'pluginMissing' ?

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+                        propertyElement.setTextContent(latestVersion);
+                        updated = true;
+                    }
+                }
+            } else {
+                versionExpression = latestVersion;
+            }
+        }
+
+        if (fmpPlugin == null) {
```
rhuss: if (pluginMissing) ....

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+            fmpPlugin = findOrAddPlugin(doc, PLUGIN_GROUPID, PLUGIN_ARTIFACTID, versionExpression, configuration);
```
rhuss: afais, `configuration` is always `null`.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+            updated = true;
+        } else {
+            if (updateVersion || pluginInserted) {
```
rhuss: 'pluginInserted' (aka 'pluginMissing') cant be the the case here anymore as it already has been checked in the if() above.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+                String version = DomHelper.firstChildTextContent(fmpPlugin, "version");
+                if (version == null || !version.equals(versionExpression)) {
+                    Element versionElement = DomHelper.firstChild(fmpPlugin, "version");
+                    if (versionElement == null) {
```
rhuss: Still not sure whether we unconditionally should add a version in this case, since the version could also be specified in a `<pluginManagement>` section. Adding the version here would override the version specified in the pluginManagement section and might lead to unexpected behaviour.

It would be much better to update the `<pluginManagement>` 's version if existing.

And if we want to be really clever we should check also for parent pom's and boms (for both, a direct plugin declaration and the `<pluginManagement>`)

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+                            appendAfterLastElement(fmpPlugin, textNode);
+                        }
+                        addChildAfter(textNode, "version", versionExpression);
+                    } else {
+                        versionElement.setTextContent(versionExpression);
+                    }
+                    updated = true;
+                }
+            }
+            if (configuration != null) {
```
rhuss: Don't see where configuration is ever set (but could be wrong since I'm not in the IDE currently).

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,295 @@
...
+                Element oldConfig = firstChild(fmpPlugin, "configuration");
+                DomHelper.detach(oldConfig);
+                fmpPlugin.appendChild(configuration);
+                if (oldConfig == null) {
+                    fmpPlugin.appendChild(doc.createTextNode("\n      "));
+                }
+            }
+        }
+        if (fmpPlugin != null) {
+            Element executions = firstChild(fmpPlugin, "executions");
```
rhuss: Should executions always be added with our default ? It could be that a user has a custom `<executions>` setting (e.g. binding our mojos to different phases). This would be unconditionally overwritten here.

plugin/src/main/resources/io/fabric8/maven/versions.properties
```diff
@@ -0,0 +1,21 @@
+#
```
rhuss: File not needed anymore.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: this is a top level goal though, its never used within a regular mvn execution on a project - so figured keeping the CLI simpler was a better idea?

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: though I'd rather use the canonical name we use for maven properties by default. We've started down the road of making it easier to keep upgrading projects with the latest plugins & dependencies via the [fabric8-maven-dependencies project](https://github.com/fabric8io/fabric8-maven-dependencies) - would prefer to default to the correct property name.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: see above - not sure this stand alone goal to install the fabric8 plugin needs prefixing?

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: OK, lets only add executions the first time we run this goal. That does mean we can't ever upgrade the executions though as we add new goals (we've added 'helm' recently for example)

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
rhuss: I see your point. I'm somewhat in between simplicity (with possible clashes to other plugins, you still remember `docker.image` ;-) ?) and consistency (we _always_ prefix with `docker.` and `fabric8.` for our stuff). 

But tbh, personally I never would use an update because I would be afraid that things are messed up. Just thought about ppl using profiles, too. Do we update these as well ? Sorry, for me _update_ still feels like Pandora's box.

However, if you really insist on going this route, maybe we could set the `update` property by default to false (to avoid accidentally mess up of their pom.xml) ?

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
rhuss: We still can default to the property name `fabric8.maven.plugin.version` if we put that as the default value (if we decide to be that the default that the property is added).

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,281 @@
...
+                    }
+                }
+            } else {
+                versionExpression = latestVersion;
+            }
+        }
+
+        if (pluginMissing) {
+            fmpPlugin = findOrAddPlugin(doc, PLUGIN_GROUPID, PLUGIN_ARTIFACTID, versionExpression);
+            updated = true;
```
rhuss: `pluginMissing`  should be set to `false` here since it is still referred to below (and after that point its not missing anymore)

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
```diff
@@ -0,0 +1,281 @@
...
+        }
+        if (updateVersion || pluginMissing) {
```
rhuss: Up to here plugin should be already inserted, so pluginMissing should be always false.

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: the purpose of the flag is to track if it was missing - i.e. if we should add the executions. If we set it to false then we'll never ever add the executions. I added this logic based on your previous comment about only adding executions when we first add the plugin

plugin/src/main/java/io/fabric8/maven/plugin/SetupMojo.java
jstrachan: as in the comment above - the purpose of the pluginMissing value is to track if we actually added the plugin in this run - in which case we should add executions + set the version.

Maybe we should switch back to my original variable name - pluginInserted to track if we inserted the plugin or not

jstrachan:
@rhuss you happy with this PR? would love a release soon...

jstrachan:
@rhuss gonna merge this one for now - we can always keep polishing later - its damn useful as is!
LTGM [merge]

rhuss:
Will have a look this morning

James Strachan <notifications@github.com> schrieb am Do., 8. Sep. 2016,
09:58:

rhuss:
no problem, will have an eye on it later ;-)

PR #305, opened 2016-09-06
lets default to not creating Route / Ingress resources as we now use the exposecontroller to decouple that logic

jstrachan:


jstrachan:
[merge]

jstrachan:
LGTM

jstrachan:
[merge]

PR #313, opened 2016-09-07
Seems like k8s will restart a pod if it misses it's readiness-checks too, so Karaf will need a long delay before checking this probe.

chirino:


chirino:
[lgtm] [merge]

jimmidyson:
Hmm kubernetes shouldn't restart based on readiness checks afaik. Are you sure there's not something else going on here?

chirino:
Your right, I was just just testing with the latest snapshot.  Latest snapshot is fine!

PR #316, opened 2016-09-08
lets refactor the helm-index goal so we can also add a manifests-index goal to create an index of all kubernetes and openshift manifests

jstrachan:


jstrachan:
LTGM [merge]

jstrachan:
[merge]

PR #322, opened 2016-09-08
lets improve manifest-index so we can add some bootstrap to allow showing/hiding of the download links or the command line to install an app

jstrachan:


jstrachan:
lgtm [merge]

PR #324, opened 2016-09-09
helm chart.tgz files should put the Chart.yaml inside a folder - see https://github.com/kubernetes/helm/pull/1172

jstrachan:


jstrachan:
LGTM [merge]

Issue #333, opened 2016-09-09
`fabric8:build` should check if there is no docker host, then if gofabric8 is on the PATH use `gofabric8 docker-env`


jstrachan:
if someone clones a fabric8 enabled project and doesn't install docker or kubernetes, they'll have no `dockerHost` maven property in their `~/.m2/settings.xml` or have `$DOCKER_HOST`

However now we have `mvn fabric8:install` we've java code that can easily check for `gofabric8` on the PATH and download it if required (maybe with a user prompt).

Also if we find `gofabric8` there's a new command soon: `gofabric8 docker-env` which will figure out the docker env vars for minikube or minishift (maybe others later on).

So if there's no DOCKER_HOST the code could run `gofabric8 docker-env` to find the env vars required.

I had a quick look at implementing this - looks like docker-maven-plugin has a private method 
https://github.com/fabric8io/docker-maven-plugin/blob/7e1af5b66871631d4ab3403c282d55f9c6c520d3/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java#L298

it would be nice to make that protected so we can override it in the fabric8:build goal to auto-detect docker via gofabric8 instead.

jstrachan:
FWIW this one change would be really handy for me - as I often switch & restart minishift + minikube so a given shell usually doesn't know how to talk to the current kubernetes/openshift clusters's docker daemon at any point in time. I'd prefer all my shells to have no DOCKER_HOST by default and to lazily evaluate them via `gofabric8 docker-env`.

Its also tempting to alias `docker` to be a bash script to `eval $(docker-env}` then run the real docker binary ;)

rhuss:
I will have a look today. Shouldn't be that hard to make it protected in docker-maven-plugin.

jstrachan:
Awesome! We have `gofabric8 docker-env` now, so we should be able to easily check if gofabric8 is on the PATH, that DOCKER_HOST is missing and if so parse the results to setup the various DOCKER env stuff like DOCKER_HOST et al.

Maybe if there's no gofabric8 on the PATH and no DOCKER_HOST / dockerHost property we could also display some text after the warning about no DOCKER_HOST along the lines of 

```
no gofabric8 binary was found on the PATH which can setup your environment automatically. Try running: `mvn fabric8:install` to setup your local machine!
```

jstrachan:
[...]

PR #355, opened 2016-09-15
. #333 - Initial implementation for using `gofabric8 docker-env` as a fallback.

rhuss:
* Not yet done: Call of 'fabric8:install' in case there is now gofabric8 in the path. Currently the goal bails out.
* Requires 0.15-SNAPSHOT of d-m-p
* Also still open: `gofabric8 docker-env` asks for user + password if not connected, but this shouldnt be the case (an empty string should be printed). See  https://github.com/fabric8io/gofabric8/issues/175 for this issue.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -15,7 +15,8 @@
   ~ implied.  See the License for the specific language governing
   ~ permissions and limitations under the License.
   -->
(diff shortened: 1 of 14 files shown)


rhuss:
lgtm [merge]

rhuss:
lgtm

jstrachan:
@rhuss would love a release once the snapshot of d-m-p is removed!

rhuss:
@jstrachan will do it today

Issue #349, opened 2016-09-15
running `mvn fabric8:run` a second time should kick in a rolling upgrade to the newly built image..


jstrachan:
... so we maybe need to default to timestamp image labels like this?
https://github.com/fabric8io/funktion/blob/master/starter-parent/pom.xml#L108-L111

I've seem times the build failed too as the `1.0-SNAPSHOT` image was in use

jstrachan:
not quite there yet - the timestamp only gets re-generated when you do
```
mvn clean fabric8:run
```
it does not update for 
```
mvn fabric8:run
```

PR #356, opened 2016-09-15
part of #349 partial fix - works fine now for `mvn clean fabric8:run` though 2 subsequent run commands don't re-generate the timestamp

jstrachan:


jstrachan:
lgtm [merge]

PR #363, opened 2016-09-16
Only update app specific resource descriptors.

### LINKED ISSUE #359: Distinguish between resource templates to apply to the build image and others
Currently, when we found skeleton resources in `src/main/fabric8` we enrich them all without differentiation whether it should apply to the current app (that which is build) or a other resources which should not be enriched with build information. like database services and others.

Examples are:

* All service templates found are enriched with the ports of the image configuration found
* All rc templates found are enriched with the configured image names and put into the container specs.

My suggestion is to check if no name for a  resource in `src/main/fabric8`  is set and only enrich this with the build information for default service and default controller (and the name from the configuration). E.g. we would store those objects ins `svc.yml` or `rc.yml` whereas other resources are called like `couchdb-svc.yml` or `couchdb-rc.yml`.

rhuss:
Heads up, this might be a breaking (but necessary) change: In order to not messe up resource objects like database which are part of an application but not the application itself and which does not e.g. an update of the image spec, **only descriptors with no name or with the artifact name of the project* are enriched with Enricher.addMissingResources() (i.e. only those are feed into the enrichers).

This fixes #359.

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
@@ -63,18 +63,27 @@ public class KubernetesResourceUtil {
      * can be adapted later.
      *
      * @param apiVersion the api version to use
+     * @param apiExtensionsVersion the extension version to use
+     * @param defaultName the default name to use when none is given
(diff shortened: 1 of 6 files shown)


rhuss:
lgtm [merge]

Issue #341, opened 2016-09-13
better support for service exposing


jstrachan:
We need better docs on how to expose services easily. 

Up to now lots of our quickstarts have used `type: LoadBalancer` to expose services. We need to update those to use the label `expose: true` instead so that things get annotated nicely and if folks are running router / ingress controller they get a route/ingress.

e.g. by adding something like this to `src/main/fabric8/foo-service.yml`

``` yaml
metadata:
  labels:
      expose: true
```

Its worth adding a section to the docs on all this too as a its a little confusing

kameshsampath:
@jstrachan  this also needs to have router running as part of the minishift startup to have the service exposed and accessible via the route url like `myservice.demo.<ip>.xip.io`

jstrachan:
@kameshsampath yeah - if the router is running `expose: true` will create a Route; if there's the ingress controller running on kubernetes we'll create an Ingress otherwise we default to nodePorts locally. On a cloud provider you can also use `type: LoadBalancer` to get an external IP address. 

This complexity is why we created the exposecontroller microservice to take care of all this:
https://github.com/fabric8io/exposecontroller/

jstrachan:
the expose services docs should reference this too: https://blog.fabric8.io/accessing-your-applications-on-kubernetes-or-openshift-3e69be0ad3bc#.cphg55mzp

kameshsampath:
Thanks @jstrachan

jstrachan:
there's improvements on #409 to expose by default in the fabric8 profile

kameshsampath:
thanks @jstrachan , i see that coming with latest 3.1-SNAPHOT plugin

jstrachan:
I'll leave this open for now - we should be able to remove all the type:LoadBalancer stuff from quickstarts as the default profile now exposes services; we can then use the `internal-microservice` profile to disable it again

PR #384, opened 2016-09-20
Fix for #341, added an expose config item on services which defaults to true and is used to set the expose label.

chirino:


core/src/main/java/io/fabric8/maven/core/config/ServiceConfig.java
```diff
@@ -39,6 +39,10 @@
     @Parameter
     private boolean headless = false;
 
+    // If the expose annotation is added to the service
```
jimmidyson: expose label, not annotation

core/src/main/java/io/fabric8/maven/core/config/ServiceConfig.java
```diff
@@ -79,6 +87,11 @@ public Builder headless(boolean headless) {
             return this;
         }
 
+        public Builder expose(boolean headless) {
```
jimmidyson: Paramater name should be `expose`.

core/src/main/java/io/fabric8/maven/core/handler/ServiceHandler.java
```diff
@@ -99,7 +100,7 @@ public Service getService(ServiceConfig service) {
         return ret;
     }
 
-    private Map<String, String> getPrometheusAnnotations(ServiceConfig service) {
+    private Map<String, String> getAnnotations(ServiceConfig service) {
```
jimmidyson: Why rename this method?

chirino: This method could add other annotation besides prometheus ones.

chirino:
Updated and squashed based on comments.

jimmidyson:
LGTM [merge]

PR #392, opened 2016-09-21
fixes #389 to allow existing projects to be imported into fabric8/openshift either existing git repos or create new ones on the fly and link to the console page to view or customise the pipeline

### LINKED ISSUE #389: lets have a `mvn fabric8:import` goal to make it easy to import projects into fabric8
which then creates the `BuildConfig` from the git repository of the local project. If there is no git repository then also create a new gogs repository and push the code there. 

This goal should also prompt the user for user/password for https based git protocols; or SSH keypairs for git/ssh protocol so that we can pre-populate the secrets easily for reuse by the console

jstrachan:


jstrachan:
lgtm [merge]

PR #408, opened 2016-09-26
Convert Docker Host URL properly

### LINKED ISSUE #333: `fabric8:build` should check if there is no docker host, then if gofabric8 is on the PATH use `gofabric8 docker-env`

rhuss:
Fixes #333

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/Gofabric8Util.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/Gofabric8Util.java
@@ -22,6 +22,7 @@ import java.util.*;
 
 import io.fabric8.maven.docker.access.DockerConnectionDetector;
 import io.fabric8.maven.docker.access.util.EnvCommand;
+import io.fabric8.maven.docker.util.EnvUtil;
 import io.fabric8.maven.docker.util.Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


rhuss:
lgtm [merge]

PR #411, opened 2016-09-26
fixes #409 so that we expose services by default in the fabric8 profile; or folks can use the `internal-microservice` profile to exclude the expose enricher

### LINKED ISSUE #409: should we expose services by default in the `fabric8` profile via expose label then let folks switch profile for internal microservices

jstrachan:


jstrachan:
lgtm [merge]

PR #416, opened 2016-09-26
fixes #396 to allow debug mode to be enabled via a maven or system property to speed up the `fabric8:debug` goal

### LINKED ISSUE #396: have a canonical maven property to enable debug mode when building?
so we have the [fabric8:debug](https://maven.fabric8.io/#fabric8:debug) which is great; it modifies the current running Deployment to add the debug environment variable.

However some developers in `pre-commit` flow may want to stay in debug mode; so that at any time they can run `mvn fabric8:debug` and not have to wait to update the Deployment & restart the pod.

i.e. enable the debug settings on a regular build.

So someone may want to enable debug by adding a profile to set a maven property which would mean that the `fabric8:resource` goal would add the debug environment variable by default in a regular build - meaning the `mvn fabric8:debug` goal would only have to ensure the pod is running and do the debug port proxy.

i.e. it'd make it a bit faster for folks to debug on demand if in developer mode, developers left debug enabled all the time via maven profiles etc.

Note we don't expect folks to use debug mode once code is committed; this is purely a local developer setting

jstrachan:


jstrachan:
lgtm [merge]

Issue #223, opened 2016-08-05
Move prometheus service annotations to an enricher


rhuss:
Currently they are added hardcoded directly in the ServiceHandler, which should be independent of any specific additions.

PR #419, opened 2016-09-26
fixes #418 so that we can enable spring-boot generator for nodejs funktion builds; while having a separate custom docker image configuration as well e.g. like this https://github.com/fabric8io/funktion/blob/master/starter-parent/starter-parent-sidecar/pom.xml#L59

### LINKED ISSUE #418: cannot use spring-boot generator with a custom image configuration
e.g. in nodejs funktion when we use springboot generator for the trigger and then use an image for the funktor image.

Looks like a regression

jstrachan:


jstrachan:
lgtm [merge]

generator/api/src/main/java/io/fabric8/maven/generator/api/support/BaseGenerator.java
```diff
@@ -121,8 +122,8 @@ protected String getAlias() {
     }
 
     protected boolean shouldAddDefaultImage(List<ImageConfiguration> configs) {
-        String merge =  getConfigWithSystemFallbackAndDefault(Config.alias, "fabric8.generator.alias", "false");
-        return !containsBuildConfiguration(configs) || Boolean.parseBoolean(merge);
+        String merge =  getConfigWithSystemFallbackAndDefault(Config.alias, "fabric8.generator.alias", "");
+        return !containsBuildConfiguration(configs) || Strings.isNotBlank(merge);
     }
```
rhuss: Sorry, don't get the diff. What exactly is the purpose of this change ?

jstrachan: the check for whether to invoke the base generator or spring boot generator was comparing `fabric8.generator.alias` to be true/false; when its the name of the docker image alias name.

We used this configuration a while back to explicity allow spring boot generator and a custom docker image build to coexist together.

i.e. it was a bad boolean condition - `fabric8.generator.alias` is never true.

We could use some other explicit configuration to enable the spring-boot generator if there's a `<configuration><image>` tag for a separate image and there's a spring boot app?

jstrachan: I just added https://github.com/fabric8io/fabric8-maven-plugin/pull/420 to use a more explicit boolean condition; using the generator property `enabled` on java generator / spring boot generator to force the generator to be enabled even if there's already a docker image configured

rhuss: thanks, looks better. And now I get it, too ;-) Being explicit makes much more sense. There are still some spots which need to be moved into the generators / enrichers, will tackle that this week (e.g. #223).

thx ;-)

PR #420, opened 2016-09-26
for #418 lets use a more explicit configuration option to force the spring-boot / java generators to run even if there is a current image configuration

jstrachan:


jstrachan:
lgtm [merge]

PR #429, opened 2016-09-27
fixes #428

### LINKED ISSUE #428: we should generate an ImageStream in the openshift.yml in openshift mode
so that we can deploy to another namespace via `oc apply -f target/classes/META-INF/fabric8/openshift.yml` and similarly things like arquillian and jenkins pipeline can work too

jstrachan:


plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
```diff
@@ -241,7 +293,191 @@ private void executeOpenShiftBuild(ServiceHub hub, ImageConfiguration imageConfi
...
+            is.setSpec(spec);
+        }
+        List<TagReference> tags = spec.getTags();
+        if (tags == null) {
+            tags = new ArrayList<>();
+            spec.setTags(tags);
+        }
+        TagReference tag = null;
+        if (tags.isEmpty()) {
+            tag = new TagReferenceBuilder().build();
```
rhuss: Isn't there a `new TageReference()` ?

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
```diff
@@ -241,7 +293,191 @@ private void executeOpenShiftBuild(ServiceHub hub, ImageConfiguration imageConfi
...
+            KubernetesResourceUtil.handleKubernetesClientException(e, this.log);
+        } catch (MojoExecutionException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new MojoExecutionException(e.getMessage(), e);
+        }
+
+    }
+
+    private boolean updateImageStreamTag(OpenShiftClient client, ImageConfiguration imageConfig, ImageStream is, String buildConfigName) throws MojoExecutionException {
```
rhuss: would call it differently since you are updating an `ImageStream` , not an `ImageStreamTag`.

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
```diff
@@ -241,7 +293,191 @@ private void executeOpenShiftBuild(ServiceHub hub, ImageConfiguration imageConfi
...
+        }
+        if (!Objects.equals(name, from.getName())) {
+            from.setName(name);
+            answer = true;
+        }
+        if (answer) {
+            log.info("Updated ImageStream " + buildConfigName + " to namespace: " + namespace + " name: " + name);
+        }
+        return answer;
+    }
```
rhuss: Isn't it necessary to write back the `ImageStream` to update it in the k8s cluster ?

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
```diff
@@ -241,7 +293,191 @@ private void executeOpenShiftBuild(ServiceHub hub, ImageConfiguration imageConfi
...
+                            updated = true;
+                        }
+                    }
+                }
+            }
+            if (updated) {
+                // lets store the entities again!
+                KubernetesList entity = new KubernetesListBuilder().withItems(new ArrayList<>(entities)).build();
+                File resourceFileBase = new File(this.targetDir, ResourceClassifier.OPENSHIFT.getValue());
+                AbstractResourceMojo.writeResourcesIndividualAndComposite(entity, resourceFileBase, ResourceFileType.yaml, log);
```
rhuss: I am bit ambivalent here, since the `BuildMojo` actually does *not* write resource descriptors, that's the task of the `DeployMojo`. Yes, it does create `BuildConfig` and `ImageStream` objects but only live for doing the build (not as part of the deployment). 

Not sure whether this can be kept this way, however I would really love to leave the writing of resources to disk to the `ResourceMojo` (separation of concerns). If its only for the watch functionality, isn't a live update sufficient ?

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
```diff
@@ -241,7 +293,191 @@ private void executeOpenShiftBuild(ServiceHub hub, ImageConfiguration imageConfi
...
+    }
+
+    private boolean isBuildCompleted(Watcher.Action action, Build build) {
+        BuildStatus buildStatus = build.getStatus();
+        if (buildStatus != null) {
+            String status = buildStatus.getPhase();
+            if (Strings.isNotBlank(status)) {
+                if (!Objects.equals(status, lastBuildStatus)) {
+                    lastBuildStatus = status;
+                    log.info("Build " + getName(build) + " status: " + status);
```
rhuss: Maybe we could move this to log level *verbose* (if this is called often)

rhuss: Also recommend to use the "%s" form as it can bork if the status or name contains characters building an invalid format.

E.g.

```java
log.verbose("Build %s : %s",getName(build),status);
```

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -236,6 +239,18 @@ public void executeInternal() throws MojoExecutionException, MojoFailureExceptio
...
         platformMode = clusterAccess.resolvePlatformMode(mode, log);
 
+        if (isOpenShiftMode()) {
+            Properties properties = project.getProperties();
+            if (!properties.contains(DOCKER_IMAGE_USER)) {
+                String namespace = clusterAccess.getNamespace();
+                log.info("Using docker image name of namespace: " + namespace);
+                properties.setProperty(DOCKER_IMAGE_USER, namespace);
+            }
+            if (!properties.contains(PlatformMode.FABRIC8_EFFECTIVE_PLATFORM_MODE)) {
```
rhuss: If I understand right, `FABRIC8_EFFECTIVE_PLATFORM_MODE` is used only as an internal marker for storing the detected mode when the platform mode is `auto` ?

rhuss: Same for `DOCKER_IMAGE_USER` ?

Maybe we can mark internal properties used for storing cross-Mojo state with some prefix like "internal.fabric8" ...

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -453,6 +472,26 @@ private KubernetesList convertToOpenShiftResources(KubernetesList resources) {
                 }
             }
         }
+        if (isOpenShiftMode()) {
+            for (ImageConfiguration image : this.resolvedImages) {
+                String name = image.getName();
+                int idx = name.lastIndexOf(':');
+                if (idx < 0) {
+                    log.warn("No ':' in image name: " + name);
```
rhuss: Why is it worth a warning when no tag is provided in an image name ? imo its a valid use case to have no tag (which implies `latest`)

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -453,6 +472,26 @@ private KubernetesList convertToOpenShiftResources(KubernetesList resources) {
...
+                    continue;
+                }
+                String label = name.substring(idx + 1);
```
rhuss: Tip: For parsing image names you could also use the helper class [ImageName](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/util/ImageName.java) from d-m-p

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
jstrachan: BTW we can only update the ImageStream after the BuildMojo - its the BuildMojo that creates the ImageStream tags - so we can't know what the tags are at the ResourceMojo stage.

The DeployMojo takes the current resources on disk and applies them to kubernetes/openshift; but we also release these resources which need to be post processed after a BuildMojo with the tags. e.g. if you do a release, the `openshift.yml` from a maven repo should be able to be applied via `oc apply -f` - so we need to ensure that `mvn install` will include properly tagged ImageStream resources.

I understand its not clean - but I don't see any other way unfortunately. We must post process the resources after a build.

Right now the ResourceMojo is fast so we're cool with that being part of the compile lifecycle; whereas BuildMojo tends to be slow and we only want to do that when we're positive we've updated the code and need a new docker image. We could move the ResourceMojo so its always after a BuildMojo; the worry though is often you iterate on the resources while getting things right (e.g. volume mapping or whatnot) so being able to run that goal independently is very handy given the slowness of a docker build

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
jstrachan: went with updateTagsInImageStreams / updateTagsInImageStream - PR coming soon...

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
jstrachan: yeah  -left over from a refactor where I was adding logic into the builder ;) removed...

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
jstrachan: the purpose of the updated ImageStream is not for the current build namespace; its to be able to install this image in a different namespace/cluster; so there's no need to update the current namespace at all - rather its a way of getting a reference to where the image stream is so we can tag it into different namespaces

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
jstrachan: yes - FABRIC8_EFFECTIVE_PLATFORM_MODE is just a way to detect the effective platform across various utility functions and mojos. 

DOCKER_IMAGE_USER is used by docker maven plugin to default the group (that would need fixing in d-m-p)

I've gone with `fabric8.internal....` to its clear its an internal flag not something a user should use

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
jstrachan: if we do a s2i build and can't find a tag it means we can't ever deploy that s2I image stream into any other namespace ever. So its  pretty serious issue. We get a tag even for using snapshot / latest.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
jstrachan: ah thanks! will use that...

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
rhuss: I see you point, but there are also other reasons which might speak against this:

* One reason why we put the `ResourceMojo` before the `BuildMojo` was that the generated resource descriptors are part of the jar file. If now the `BuildMojo` is required to update the descriptors *after* the build, the included descriptors are not valid anymore.

* Another thing is, isn't the `ImageStreamTag` *installation* specific ? So, does it make sense to distribute these in released artifacts at all (if e.g. the imagestream tag would be different on the production side). From my understanding, the `ResourceMojo` creates installation independent resource descriptors, which are universally applicable (and not only for the particular setup). These are included in the artifact jar and image itself. The `BuildMojo` creates images or other build related objects. The `DeployMojo` applies the universal resources and may add and mangle installation specific objects like `ImageStreamTags`.

Point 1 is a bit an egg-hen situation. I have no problem to give up on distributing the resource descriptors as part of the jar (we would need to change quite a bit for the devops stuff of how we combine descriptors to larger apps, though). 

Another idea would be to distinguish the resource descriptors between 'installation independent' (these are then released and distributed over e.g. a Maven repo) and 'installation dependen' (these we would keep locally)

imo its not so much about speed (its also important, of course), but how and what we deliver as part of the artifacts.

jstrachan: We need to distribute the descriptors as part of the jar. The issue is that the `ResourceMojo` doesn't generate complete resources; its a combination of both Mojos which does. Just like `process-resources` and `compile` make the jar contents in typical java apps. 

Maybe its easier to digest if the `BuildMojo` is the mojo which appends the `ImageStream` resources? (If it were not for the combining of resources into uber-yamls, then the `BuildMojo` would just be generating its own resources in addition to those created by `ResourceMojo` - then ordering is less important really too. To simplify things I'm tempted to refactor the logic along these lines; that `ResourceMojo` generates most resources; then `BuildMojo` appends `ImageStreams` in openshift mode; then they are logically combined - I'd also like to make them order agnostic (so re-running `ResourceMojo` in isolation won't zap all the `ImageStream` resources generated from the previous `BuildMojo` goal).

On point 2; the ImageTag is how to install the image stream into other namespaces for single cluster openshift environments. e.g. if folks have 1 cluster, they build/release in one namespace then reuse those same resources to install in any other namespace. I'm hoping we can figure out how to also make this ImageTag multi-cluster aware (I'm hoping its just a case of also including the URL of the internal openshift docker registry too) - in which case they will be installation agnostic (provided the cluster that built the image stays up). Not 100% sure as of yet on the multi-cluster nature - but its my hope we can make them multi-cluster-aware somehow. Otherwise its time to drop ImageStreams completely as not fit for purpose and switch tacks and use regular docker registries again ;). 

To say that another way; I don't think open source projects would never use OpenShift S2I builders or ImageStreams; we'd just use docker hub instead and use real docker registries. But when developing on premise releases using OpenShift and S2I then you do want to put the ImageTag in your releases so that you can use `oc apply` to install the released resources & image in any namespace

jstrachan: BTW the latest patch to the PR now cleanly separates things; `ResourceMojo` does all resources apart from `ImageStream` related stuff. Then `BuildMojo` does the `ImageStream` stuff. Plus either goal can be invoked in any order as they both combine their output logically together

rhuss: > To simplify things I'm tempted to refactor the logic along these lines; that ResourceMojo generates most resources; then BuildMojo appends ImageStreams in openshift mode; then they are logically combined - I'd also like to make them order agnostic (so re-running ResourceMojo in isolation won't zap all the ImageStream resources generated from the previous BuildMojo goal).

That's a very good idea, so that we can run both goals independently if they write into different files. For the uber jar (which we don't need for running `fabric8:deploy`) we could use an own `fabric8:package`. Then the analogy to the Maven lifecycyle would be complete:

* *resource* --> `fabric8:resource`
* *compile* --> `fabric8:build`
* *package* --> `fabric8:package`

`fabric8:package` would then be responsible for packing up things like creating the uber yaml. That way we can let `fabric8:resource` and `fabric8:build` write in to separate files, `fabric8:package` would then combine them (and if required, take care of the order).

BTW, there is one thing were ImageStream are superior than images in registries: With refering to image in registries we also bind the resource objects (pod spec) to a specific registry, which would be the same as say, you need to refer to a maven repo from within a war.xml in an war. This means you refer to meta information which might change (e.g. when you move your app image to another registry you need to update the resource descriptor within the app). 

With `ImageStreams` (not *ImageStreamTag* !) you abstract away the registry (its the internal registry which can delegate to other registries, too), so in principle this is better for an universal resource descriptor. Not for ImageStreamTag refering to a very specific image via a checksum. Maybe we can even work more with ImageStreams themself ? (not so sure what I'm saying here ;-)

rhuss: I don't think that multi-cluster via OpenShift registries will be possible, since the OpenShift internal registry then would need to be exposed externally which is not the normal use case for OpenShift (and probably forbidden in many situations). Also authentication is hard since you would need to log in via a token to access the registry from external, not so possible in automatic setups like refering from a resource descriptor.

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
rhuss: thx :)

plugin/src/main/java/io/fabric8/maven/plugin/BuildMojo.java
rhuss: Understand, its only for updating the resource descriptor files.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
rhuss: Ok, let me check the DOCKER_IMAGE_USER thing ...

jstrachan: we could also let f-m-p just set the image user directly via a setter on the ImageNameFormatter class (I think thats the one thats using it) - to avoid using maven properties to configure it

rhuss:
Yeah, think we found a good solution (separating descriptor files).

lgtm

rhuss:
[merge]

Issue #430, opened 2016-09-28
sourceStrategy.from.name: Invalid value: "fabric8/s2i-java"


chirino:
Seems fabric8/s2i-java is not a valid image stream name.  When run against OpenShift, the following:
    
    git clone git://github.com/fabric8-quickstarts/spring-boot-camel.git 
    cd spring-boot-camel
    mvn clean install -Dfabric8.build.recreate=all fabric8:deploy  -Dfabric8.mode=auto

fails with:

[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.1.53:build (default) on project spring-boot-camel: Execution default of goal io.fabric8:fabric8-maven-plugin:3.1.53:build failed: Failure executing: POST at: https://192.168.64.2:8443/oapi/v1/namespaces/test/buildconfigs. Message: BuildConfig "spring-boot-camel" is invalid: spec.strategy.sourceStrategy.from.name: Invalid value: "fabric8/s2i-java": ImageStreamTag name contains invalid syntax. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=spec.strategy.sourceStrategy.from.name, message=Invalid value: "fabric8/s2i-java": ImageStreamTag name contains invalid syntax, reason=FieldValueInvalid, additionalProperties={})], group=null, kind=BuildConfig, name=spring-boot-camel, retryAfterSeconds=null, additionalProperties={}), kind=Status, message=BuildConfig "spring-boot-camel" is invalid: spec.strategy.sourceStrategy.from.name: Invalid value: "fabric8/s2i-java": ImageStreamTag name contains invalid syntax, metadata=ListMeta(resourceVersion=null, selfLink=null, additionalProperties={}), reason=Invalid, status=Failure, additionalProperties={}). -> [Help 1]

rhuss:
Previously, the from was always considered to be a Docker image name, for which fabric8/s2i-java is perfectly valid.

However when you change this to an `ImageStream`, it must be a [simple name](https://github.com/openshift/origin/blob/48f6a2807d7b2920b5301152cbe2e8168bb7d4bf/pkg/image/api/validation/validation.go#L37) without [registry part](https://github.com/openshift/origin/blob/48f6a2807d7b2920b5301152cbe2e8168bb7d4bf/pkg/api/helpers.go#L13).

[...]

PR #431, opened 2016-09-28
#430 fix regressions in recent changes to custom BuildStrategy config…

jstrachan:
…urations so that things still work for regular docker based fabric8/s2i-java builds


jstrachan:
lgtm [merge]

PR #435, opened 2016-09-28
fix for issue #430 reverting to docker image by default instead of im…

chirino:
…age streams for s2i binary builds.


jstrachan:
lgtm [merge] - hoping for no NPE this time ;)

PR #436, opened 2016-09-28
fixes #422 so we tail the s2i binary build and fail the build early if it fails or is cancelled

### LINKED ISSUE #422: openshift s2i binary build should tail the build log and wait until its complete by default
so that if folks do `mvn fabric8:run` then we wait for the S2I binary build to complete first

jstrachan:


jstrachan:
lgtm [merge]

PR #448, opened 2016-09-29
lets move helm before `verify` to avoid breaking releases when the helm.tgz doesnt get signed

jstrachan:


jstrachan:
lgtm [merge]

PR #453, opened 2016-09-30
Issue 149

kameshsampath:
This PR has support WAR generator plain old webapp

* set default docker-image i.e. `fabric8/tomcat-8`
* ability for the users to provide the app server configuration via generator config
* add one simple samples for tomcat(default), wildfly and jetty
* auto detect the application server based on some magic files i.e. application server specific files


rhuss:
Thanks a lot ! I will have review later today or over the weekend.

kameshsampath:
Thanks!

Get Outlook for iOS




On Fri, Sep 30, 2016 at 3:42 PM +0530, "Roland Huß" <notifications@github.com> wrote:





Thanks a lot ! I will have review later today or over the weekend.

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+                .alias(getAlias())
+                .buildConfig(buildBuilder.build());
+        configs.add(imageBuilder.build());
+
+        return configs;
+    }
+
+    protected List<String> extractPorts() {
+        List<String> answer = new ArrayList<>();
+        addPortIfValid(answer, getConfig(Config.webPort));
```
rhuss: The WebPort could be also chosen based on the selected base image since it is the Servlet container which exposes the port.

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+        addPortIfValid(answer, getConfig(Config.jolokiaPort));
```
rhuss: guess the wildfly image doesn't contain Jolokia, so a base image based port addition as mentioned above makes sense here, too.

samples/simple-jetty-webapp/pom.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
rhuss: Minor things: We prefer spaces over tabs (and for XML indentation with 2-spaces)

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+
+    @Override
+    protected String getFrom() {
+        String from = super.getFrom();
+
+        if (from == null) {
+
+            if (isJetty()) {
```
rhuss: Instead of many `if (is...())` I would prefer an `AppServerDetector` interface with methods

```java
boolean isApplicable()
String getFrom()
String getCommand()
String getDeploymentDir()
```

Then you could create a list of these detectors (either in the generator or a helper class) and iterate over it until the first `isApplicable()` returns true. All the other parameters can then get from the found detector

Advantages:

* Easier to extend for future appservers (you only need to add new implementation of this interface to the list instead of finding and changing all `if(...)`
* Detection is only called once and not with every `isTomcat()`

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+
+        return cmd;
+    }
+
+    private String getDefaultBaseDir() {
+
+        String defaultBaseDir = getConfig(Config.deploymentDir);
+
+        if(isWildFly()
+                && "/deployments".equals(defaultBaseDir)){
```
rhuss: This will fail if the `deploymentDir` is set explicitely to "/deployments". A reason for this might be that a user do not want to rely on default values. He then would be surprised that for widlfly this would be changed to "/opt/jboss/wildfly/standalone/deployments" although he set it to /deyploments. Also, if we later might want to change the default value, then we can easily forget the spot here, which would needed to be updated, too.

It would be better here to check, whether the config is really set (so not with `getConfig()` but checking directly). On the other hand, I question the value of this config variable anyway, since it must "fit" to the base image. Maybe we should only evaluate it, when the "from" is also changed manually ?

generator/webapp/src/test/java/io.fabric8.maven.generator.webapp/AppServerAutoDetectionTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.Test;
+
+import java.io.File;
+
+import static junit.framework.TestCase.*;
+
+/**
+ * @author kameshs
+ */
+public class AppServerAutoDetectionTest {
```
rhuss: Thanks for the test ;-) !

generator/webapp/src/test/java/io.fabric8.maven.generator.webapp/AppServerAutoDetectionTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Test
+    public void testIsWildFly() throws Exception {
+        String tmpDir = System.getProperty("java.io.tmpdir");
```
rhuss: An alternative would be to use `File.createTempFile()` or `Files.createTempDirectory()`. The advantage is that it gets cleaned up automatically after the test.

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
kameshsampath: @rhuss can you please tell me more on this point "Detection is only called once and not with every isTomcat()" ? I am not getting what you mean here

rhuss: I simply mean that the 'hasWildFlyFiles()' etc. in AppServerDetector are called multiple times for each `isWildfly()` call in the generator (three times if I count right). Using the approach above would do the check on the files only once (in `isApplicable()`)

generator/webapp/src/test/java/io.fabric8.maven.generator.webapp/AppServerAutoDetectionTest.java
rhuss: It's Java 7 ;-) (and btw, we still target Java 7 with this plugin(

samples/simple-jetty-webapp/pom.xml
rhuss: For XML I use this setting:

image

generator/webapp/src/test/java/io.fabric8.maven.generator.webapp/AppServerAutoDetectionTest.java
kameshsampath: @rhuss my bad ..  i never explored those in Java 7 ...

rhuss: hey, no problem at all ;-) just a heads up (because I found it useful from time to time)

kameshsampath:
thanks @rhuss.  thats how we learn :) .. am learning a lot after started to work with you guys in form simple/small PR’s..

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
kameshsampath: @rhuss am not sure am getting this point completely.  can you please give me more details -  typically when you say "WebPort could also be chosen based on ..."

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
kameshsampath: @rhuss  - bit confused here .i understood that when developer sets this config for baseDir manually to "/deployments" then we are forcing the plugin to set to /opt/deployments, which I agree its wrong . when you say you want to check config you mean to do via Maven Model --> f8 plugin --> properties ?  some more details help here..

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
rhuss: I mean the AppServerDetector which kicks in knows about the base image (if no `from` is set explicitely). So it also knows which ports its exposes. Its 8080 by default but eg. Jolokia is only exposed by our base images, not by wildfly.

so it would be cool if the `AppServerDetector` would have a method `getPorts()` or so which is called by the generator and added to the list of ports.

kameshsampath:
Makes sense @rhuss, let me add that method to the interface
Get Outlook for iOS




On Tue, Oct 4, 2016 at 9:36 PM +0530, "Roland Huß" <notifications@github.com> wrote:





rhuss commented on this pull request.



> +                .ports(extractPorts())
+                .cmd(getDockerRunCommand())
+                .env(getEnv());
+        addLatestTagIfSnapshot(buildBuilder);
+        imageBuilder
+                .name(getImageName())
+                .alias(getAlias())
+                .buildConfig(buildBuilder.build());
+        configs.add(imageBuilder.build());
+
+        return configs;
+    }
+
+    protected List<String> extractPorts() {
+        List<String> answer = new ArrayList<>();
+        addPortIfValid(answer, getConfig(Config.webPort));

I mean the AppServerDetector which kicks in knows about the base image (if no `from` is set explicitely). So it also knows which ports its exposes. Its 8080 by default but eg. Jolokia is only exposed by our base images, not by wildfly.

so it would be cool if the `AppServerDetector` would have a method `getPorts()` or so which is called by the generator and added to the list of ports.

--
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub:
https://github.com/fabric8io/fabric8-maven-plugin/pull/453

rhuss:
Superseeded by #488

PR #461, opened 2016-09-30
fix for issue #430 reverting to docker image by default instead of image streams for s2i binary builds.

chirino:


rhuss:
lgtm [merge]

PR #470, opened 2016-10-01
we need to update the S2I BC whenever the version tag changes to ensu…

jstrachan:
…re we generate the correct versioned tags


jstrachan:
lgtm [merge]

Issue #466, opened 2016-10-01
fmp:cluster-start fails on osx - Error executing minikube


davsclaus:
Instead of minishift I am trying with minikube, and it fails at
```
davsclaus:/Users/davsclaus/Documents/workspace/fabric8-hello (master)/$ mvn io.fabric8:fabric8-maven-plugin:3.1.59:cluster-start
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building Maven Stub Project (No POM) 1
[INFO] ------------------------------------------------------------------------
[INFO]
[INFO] --- fabric8-maven-plugin:3.1.59:cluster-start (default-cli) @ standalone-pom ---
[WARNING] F8> Note that the fabric8 folder /Users/davsclaus/.fabric8/bin is not on the PATH!
[WARNING] F8> Please add the following to your ~/.bashrc:
[WARNING] F8>   export PATH=$PATH:/Users/davsclaus/.fabric8/bin
Would you like to add this line to your ~/.bashrc now? (Y/n): Y
[INFO] F8> Updated /Users/davsclaus/.bashrc. Please type the following command to update your current shell:
[INFO] F8>   source ~/.bashrc
[INFO] gofabric8> fabric8 recommends OSX users use the xhyve driver
[INFO] gofabric8> xhyve driver already installed
[INFO] gofabric8> Downloading https://storage.googleapis.com/minikube/releases/v0.10.0/minikube-darwin-amd64...Downloaded minikube
[INFO] gofabric8> Downloading https://storage.googleapis.com/kubernetes-release/release/v1.3.8/bin/darwin/amd64/kubectl...Downloaded kubectl
[INFO] gofabric8>
[INFO] F8> running: gofabric8 start --batch --console
[INFO] gofabric8> fabric8 recommends OSX users use the xhyve driver
[INFO] gofabric8> xhyve driver already installed
[INFO] gofabric8> Downloading https://storage.googleapis.com/minikube/releases/v0.10.0/minikube-darwin-amd64...Downloaded minikube
[INFO] gofabric8> Downloading https://storage.googleapis.com/kubernetes-release/release/v1.3.8/bin/darwin/amd64/kubectl...Downloaded kubectl
[INFO] gofabric8> Unable to get status exec: "minikube": executable file not found in $PATH
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[...]

PR #472, opened 2016-10-01
lets improve cluste-start for #466 #465

jstrachan:
no need to run `gofabric8 install` as gofabric8 start does that anyway and avoids the double download or download of minikube when using minishift


jstrachan:
lgtm [merge]

PR #474, opened 2016-10-01
Update pom property fabric8.maven.plugin.version to 3.1.62

rawlingsj:


rawlingsj:
[merge]

rawlingsj:
lgtm

rawlingsj:
what is this new lgtm comment? it breaks the CD pipelines as mentioned on IRC when it was talked about.

if changes are made to the way pull requests are merged then changes are needed to the cd pipelines.  each fabric8 project is built and released using the same scripts so if a change to pull requests is made thats fine but that needs to be applied to all projects and pipelines.

jimmidyson:
iirc lgtm has been on this repo for a while since @rhuss added it. We were trying to get to a point of actually having code reviews to get some stability in there but as people are continuing to self merge it seems pointless.

Either way, we should disable lgtm & move to github code review approval anyway. That needs to be added to CD pipelines regardless as a lot of people are moving to this since it's release so whether we use it or not, it will be important for community.

rawlingsj:
That's all totally fine but my point is if this happens then it needs to be coordinated and applied to all fabric8 projects.  It also needs to be planned, agreed and prioritised as there's an effort to do this and change to the way we all work.  It's not as simple as applying a new rule to one project.

jimmidyson:
> planned, agreed and prioritised

Oh yeah like all the other changes :-b

BTW lgtm was discussed on IRC before it was added & has been followed by the active maintainers on this repo, as you can see by looking back at PRs. Anyway, let's discuss the required changes on the issue you raised.

PR #477, opened 2016-10-03
lets detect spring fat jars better and work with flat classpath or fa…

jstrachan:
…t jars easier - and make it easier to just enable fat jars via configuration

fixes https://github.com/fabric8io/fabric8/issues/6417


jstrachan:
lgtm [merge]

Issue #483, opened 2016-10-03
gofabric8 version check


jimmidyson:
As discussed in #482, the current method of doing the version check is fragile & a borked gofabric8 release (as happened today) means that everyone's `mvn fabric8:cluster-start` stops working too. The problem is that the current way to check for the latest version is to read `https://raw.githubusercontent.com/fabric8io/gofabric8/master/version/VERSION`. That file is on a development branch (master) & indicates the current dev version, not latest released version.

We need to change to a different method for checking for updated versions. In [minishift](https://github.com/fabric8io/minishift), we use github releases API. This works great, but due to rate limiting on unauthenticated requests sometimes causes updates to fail. A simple oauth token to add to requests can get around that, but you might consider that too much.

rhuss:
My suggestion is to use a property in `pom.xml` for the current stable gofabric8 version. This property then can be updated for each gofabric8 version by our CD as we do for other properties, too.

If there are no objections I'll change this and put it into a `<version.gofabric8>` property in the parent pom.xml

jimmidyson:
So you're suggesting a new release every time there's a gofabric8 release? I'm personally not too keen on that as I can see there being times when the plugin isn't in a state to be released & a new version of gofabric8 is ready to be consumed.

rhuss:
I mean that an application should declare its dependencies clearly. So when f-m-p depends on gofabric8 it should declare the version it depends on (because, considering we would have tests, this is the version we tested against). The situation we just has was, that a "new" version of gofabric8 (i.e. something changed on the gofabric8 side) broke fabric8-maven-plugin. I know this could be hardened by using real releases from the GitHub API, but it still leaves open the question, whether an application should be defensive by defining its dependencies or trust every new shiny release of gofabric8.

If we consider to always use the newest release, why don't we use always SNAPSHOT for all our projects and depend on them ?

It's really depends on your point of view. I'm about to send a PR for doing it 'conservatively' since I'm a fan of stating dependencies explicitly, but it depends on the discussions here, whether we want it to apply. BTW, all our other dependencies e.g. on fabric8 or kubernetes0client itself works by providing explicit versions in the pom.xml

If there is a new release of gofabric8 which is required by f-m-p for some fix or new features, I would release a new version of f-m-p.

rohanKanojia:
Since gofabric8 is not supported anymore, not sure if this issue is still valid. Closing this, please reopen if it still looks valid to you

Issue #482, opened 2016-10-03
cluster-start fails on linux (fedora)


sterpico:
Having the following plugin configuration:
`<plugin>
  <groupId>io.fabric8</groupId>
	<artifactId>fabric8-maven-plugin</artifactId>
	<version>3.1.58</version>
	<executions>
	<execution>
	 <id>fmp</id>
		<goals>
			<goal>resource</goal>
			<goal>helm</goal>
			<goal>build</goal>
		</goals>
	</execution>
	</executions>
</plugin>`
and issuing mvn fabric8:cluster-start, generates the following error:
 ...
[INFO] --- fabric8-maven-plugin:3.1.58:cluster-start (default-cli) @ kubernetes-fabric8 ---
[INFO] F8> Downloading version 0.4.77 of gofabric8 to /home/sterpico/.fabric8/bin/gofabric8 
...
[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.1.58:cluster-start (default-cli) on project kubernetes-fabric8:
Failed to open URL: https://github.com/fabric8io/gofabric8/releases/download/v0.4.77/gofabric8-linux-amd64. java.io.FileNotFoundException:
https://github.com/fabric8io/gofabric8/releases/download/v0.4.77/gofabric8-linux-amd64 -> [Help 1]

I manually installed gofabric8 version 0.4.76 (latest at the time of writing), however the plugin always attempts to resolve this version instead. The issue could be similar to the windows one, however is there a way to instruct it to work with this gofabric8 version?

anthonykulis:
following. i get the same issue on mac

sterpico:
By browsing through the source code, the call to https://raw.githubusercontent.com/fabric8io/gofabric8/master/version/VERSION returns 0.4.77 instead of 0.4.76, so the plugin attempts to resolve that version instead.

diesim:
Same issue all around no matter OS. This is also the same as the windows one posted earlier.

jimmidyson:
Looks like a borked release - will fix up soon.

jimmidyson:
Didn't know that was how the version was discovered... We'll change that
soon.

jimmidyson:
This should be fixed now that 0.4.77 of gofabric8 is released. I'll open a new issue to discuss how the plugin is doing the latest version check.

PR #484, opened 2016-10-03
. #483 Set versions of gofabric8 to use in pom.xml

rhuss:
This fixes problems, when unconsolidated changes happen outside of fabric8-maven-plugin. See #482 for such an scenario.


rhuss:
Doesn't make much sense probably anyway. so lets close it.

PR #486, opened 2016-10-04
OSFUSE-327: f-m-p Karaf generator should get the version info of the images via build properties

chirino:
Also put all image version in the parent/pom.xml for easy updating.

### DIFF
--- a/generator/api/pom.xml
+++ b/generator/api/pom.xml
@@ -32,10 +32,10 @@
 
   <properties>
     <!-- Default images used in Java based generators -->
-    <generator.java.docker.upstream>fabric8/java-alpine-openjdk8-jdk:1.2.0</generator.java.docker.upstream>
-    <generator.java.s2i.upstream>fabric8/s2i-java:1.3.4</generator.java.s2i.upstream>
-    <generator.java.docker.redhat>jboss-fuse-6/fis-java-openshift:2.0</generator.java.docker.redhat>
-    <generator.java.s2i.redhat>jboss-fuse-6/fis-java-openshift:2.0</generator.java.s2i.redhat>
+    <generator.java.docker.upstream>fabric8/java-alpine-openjdk8-jdk:${generator.java.docker.upstream.version}</generator.java.docker.upstream>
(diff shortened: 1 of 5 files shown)


generator/api/pom.xml
```diff
@@ -32,10 +32,10 @@
 
   <properties>
     <!-- Default images used in Java based generators -->
-    <generator.java.docker.upstream>fabric8/java-alpine-openjdk8-jdk:1.2.0</generator.java.docker.upstream>
-    <generator.java.s2i.upstream>fabric8/s2i-java:1.3.4</generator.java.s2i.upstream>
-    <generator.java.docker.redhat>jboss-fuse-6/fis-java-openshift:2.0</generator.java.docker.redhat>
-    <generator.java.s2i.redhat>jboss-fuse-6/fis-java-openshift:2.0</generator.java.s2i.redhat>
+    <generator.java.docker.upstream>fabric8/java-alpine-openjdk8-jdk:${generator.java.docker.upstream.version}</generator.java.docker.upstream>
```
rhuss: The question is which we want to extract the version number separately ? We could move the whole properties here into the parent.xml, too. tbh I dont see yet the value of this extra level of indirection.

generator/karaf/src/main/java/io/fabric8/maven/generator/karaf/KarafGenerator.java
```diff
@@ -30,8 +32,28 @@
...
+    }
+
+    private static String getBuildSetting(String key) {
+        String val = BUILD_SETTINGS.getProperty(key);
+        if (val == null) {
+            throw new IllegalArgumentException("Cannot retrieve default value " + key + " from " + SETTINGS_RESOURCE_FILE);
+        }
+        return val;
+    }
+
```
rhuss: Wonder whether we should combine this lookup with the one of the JavaRunGenerator s, too. But we can do this always later, too.

generator/api/pom.xml
chirino: The version is the part the changes most often as part of release builds.  Wanted a simple property that holds that value so the CD automation can easily update it.

rhuss:
going to merge this and unify the property handling.

thanks !

rhuss:
Cosed in favor of #496

PR #496, opened 2016-10-05
version properties for every generator.

rhuss:
They are defined in the parent's pom.xml and could be updated by CD.


rhuss:
lgtm [merge]

PR #498, opened 2016-10-05
Make WebAppGenerator stateless

rhuss:
This e.g. avoid an issue with scanning directories even when this generator is not
ative.


rhuss:
lgtm [merge]

Issue #503, opened 2016-10-06
provide a goal to watch spring boot containers for local development changes & reload on the fly


jstrachan:

rhuss:
"I think we should move now the spring-boot-generator to fabric8-maven-core, too. makes more sense as the core is now prominently coupled to spring-boot anyway. So we can avoid this unnecessary abstraction."

sorry, not my day.

This feature definitely goes absolutly in the right direction for providing an awesome developer experience. And we should have this. But not hardcoded as a goal sticking to a specific technology. This carries all this generator / enricher separation ad absurdum. I wouldn't mind if we open a spring-boot-kubernetes-plugin which is only about spring-boot, but for a general purpose plugin this is far too specific for a goal. 

Do we plan to introduce 'fabric8:watch-widlfly-swarm', 'fabric8:watch-tomcat', ... so that each developer needs to remember whether its current microservice project (out of thousand others) is running spring-boot, widlfy-swarm, ... ?

My suggestion is to do this under the umbrella `fabric8:watch`, with maybe a first hard coded detection. Later then something like 'enricher' and 'generator', like 'watcher' which are then specific to each tech.

Isn't it all about developer experience ? why should a developer need to specify `watch-spring-boot` when the plugin could detect it itself ?

jstrachan:
yeah - it'd be nice to just have a `fabric8:watch` and it figure out the right thing to do!

for app server like stuff like tomcat/jetty/wildfly/karaf we should be able to just reuse the current `fabric8:watch` goal really; so I'm not expecting a load of new commands.

jstrachan:
all fixed now - see https://maven.fabric8.io/#fabric8:watch-spring-boot

Issue #301, opened 2016-09-02
Neither spring nor karaf should be referenced in core's ContainerHandler


rhuss:
Adding specific health checks for systems like spring-boot and karaf must not be part of the cor module but of a specific enricher or generator which detects its environment and add this kind of stuff.

I.e. https://github.com/rhuss/fabric8-maven-plugin/blob/master/core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java#L101-L107 should be removed in favour of adding this functionality to the spring-boot and karaf enricher, respectively.

rhuss:
changed already

PR #509, opened 2016-10-07
Reading spring-boot configuration also from application.yml

nicolaferraro:
This allow setting the correct port for health checks even if the `management.port` is set on a `application.yml` file (only `application.properties` was supported before).


core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -180,10 +184,17 @@ public static boolean hasClass(MavenProject project, String ... classNames) {
     }
 
     /**
-     * Returns the spring boot `application.properties` or an empty properties object if not found
+     * Returns the spring boot configuration (supports `application.properties` and `application.yml`)
+     * or an empty properties object if not found
      */
     public static Properties getSpringBootApplicationProperties(MavenProject project) {
-        return getPropertiesFile(project, "application.properties");
+        URLClassLoader compileClassLoader = getCompileClassLoader(project);
```
rhuss: Not your fault, but more a reminder to myself: This is the wrong class for spring-boot handling, In fact its even the wrong module. All specific tech stuff (spring-boot, wildfly-swarm, karaf, tomcat, ...) must be in generators and enrichers which nicely encapsulates these tech. See also #301 and comments to #503 .

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -180,10 +184,17 @@ public static boolean hasClass(MavenProject project, String ... classNames) {
...
+        URL ymlResource = compileClassLoader.findResource("application.yml");
+        URL propertiesResource = compileClassLoader.findResource("application.properties");
+
+        Properties props = getPropertiesFromYamlResource(ymlResource);
+        props.putAll(getPropertiesResource(propertiesResource));
```
rhuss: So iiur `application.properties` take precedence over `application.yaml`. Is this also the spring-boot behaviour when starting up ?

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -204,6 +222,78 @@ public static Properties getPropertiesFile(MavenProject project, String properti
...
+        if (resource != null) {
+            try (InputStream yamlStream = resource.openStream()) {
+                Yaml yaml = new Yaml();
+                @SuppressWarnings("unchecked")
+                SortedMap<String, Object> source = yaml.loadAs(yamlStream, SortedMap.class);
+                Properties properties = new Properties();
+                properties.putAll(getFlattenedMap(source));
+                return properties;
+            } catch (Exception e) {
+                LOG.error("Failed to load yaml properties from: " + resource + ". " + e, e);
```
rhuss: The general question is, whether we should really only log the error or wouldn't it be better for this utility class to propagate a IOException to its called. Logging in utility classes is an antipattern anyway, so I would avoid that. (I know this has been there previously already, but would like to clean it now anyway :)

So I would simply remove the catch block and let the exception propagate.

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
```diff
@@ -204,6 +222,78 @@ public static Properties getPropertiesFile(MavenProject project, String properti
...
+            else if (value instanceof Collection) {
+                Collection<Object> collection = (Collection<Object>) value;
+                int count = 0;
+                for (Object object : collection) {
+                    buildFlattenedMap(result,
+                            Collections.singletonMap("[" + (count++) + "]", object), key);
+                }
+            }
+            else {
+                result.put(key, (value != null ? value : ""));
```
rhuss: Question is whether we would add an null value as empty string or simply skip it, if null. have no strong opinion on that, though.

core/src/test/java/io/fabric8/maven/core/util/MavenUtilTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Checking the behaviour of utility methods.
+ */
+public class MavenUtilTest {
+
+
+    @Test
+    public void testYamlToPropertiesParsing() {
```
rhuss: Some more tests would be awesome, maybe with a dedicated test file (not reusing an exisiting one, as it makes tests dependant on each other).

Something like multiple arrays after each other, or null values, ...

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
nicolaferraro: Yeah, I tried to stick with the existing code, but I'm ok to try a refactoring :)

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
nicolaferraro: That's what spring boot does...

nicolaferraro:
I'll make some more changes and update the PR.

core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
nicolaferraro: Yep, verified. Although this should not happen in practice.

nicolaferraro:
I moved every karaf and spring-boot stuff to their own enricher. The code looks much clearer now :)

plugin/pom.xml
```diff
@@ -76,6 +76,16 @@
 
     <dependency>
       <groupId>io.fabric8</groupId>
+      <artifactId>fabric8-maven-enricher-karaf</artifactId>
```
rhuss: question is whether we maybe should have a single 'fabric8-maven-enricher-healthcheck' module containing both enrichers. They still can be switched on individually and since its part of the f-m-p plugin it doesnt make much sense to have separate modules so that only one might be included as a dependency. 

In general I avoid to introduce to many modules, because they typically increase the build time and release considerably and then there is a maintenance overhead. More specific, I'd try to avoid modules with a single class. Was there (in a 70 modules projects) and it was not much fun ;-)

nicolaferraro: In general I would create a separate module if that let me organize things better. For example, I was working on a spring-boot quickstart and noticed that yml properties are ignored when building the health checks. So I went to the "enrichers" directory expecting to find a spring-boot module... At the end I found where it was by searching the string "8081" and finding it in a comment :)

But you're right about the modules, it was almost impossible to build last Apache Camel after we have introduced other 200+ modules (also because of me :D). So also including them into standard or create a "stack" / "technology" module where things are separated into packages (not by type of enricher, but by technology stack) would be a good option. What do you think?

enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private Container getContainer(Deployment deployment) {
+        if (deployment.getSpec() != null &&
+                deployment.getSpec().getTemplate() != null &&
+                deployment.getSpec().getTemplate().getSpec() != null &&
+                deployment.getSpec().getTemplate().getSpec().getContainers() != null &&
+                deployment.getSpec().getTemplate().getSpec().getContainers().size() > 0) {
+
+            List<Container> containers = deployment.getSpec().getTemplate().getSpec().getContainers();
+            return containers.get(containers.size() - 1); // last one
```
nicolaferraro: I've also a doubt about this part. I did it so because the original "enricher" updated only the last image.
But I'm not aware of the cases where the f-m-p can build more than 1 container and put all them in a deployment, and why only the last one should be enriched. Thoughts?

plugin/pom.xml
rhuss: Let's start a `fabric8-maven-enricher-healthcheck` module and see how it feels. I agreed its easier to find spring-boot related code when the modules are called like that, on the other hand one could argue, when looking for healt-check it's easier when there is a heatlcheck module. It all depends on the perspective ;-)

We can still later move or split stuff. E.g. when we decide to have the tech as main focus, we could also combine the spring-boot generator, the spring-boot healthcheck and maybe later the spring-boot "watcher" into a dedicated 'spring-boot' module.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
rhuss: In deed, f-m-p is supposed to deal with multiple containers from multiples images within a pod. However, currently the 'generators' (which are supposed to create default images) kick in by default only if there is no other image configured. 

Maybe it would make sense to establish some sort of 'communication' between generators and enrichers, eg. by setting a certain marker in the image configuration.

Then the algo could be:

* If there is only one image / container in the deployment config and we are a spring-boot project, lets add the health check.
* If there are multiple containers, lets check for this 'marker' and add health checks to each of the marked containers.

The question of course is, what could be this "marker" ? A label or annotation is probably the easiest thing, however the question how this can be added to a `Deployment` config. Maybe by the enricher which adds this `Deployment` ? (if done manually, one need to add this annotation oneself in the configuration).

Still in brainstorming mode ;-) ...

nicolaferraro: Yes, now I understand better the point. But I don't get why enrichers should be bound to generators. I mean, if I don't want to use Alpine and use a different O.S., or another JVM, I'd provide my own image but probably I still want to have health checks installed in my spring-boot or vertx application.

If the images has built-in health checks I can specify them in the configuration (or if it's a well known image, an enricher can create them). I can also deactivate health-check enrichers using profiles if I don't need them... Maybe I don't have the whole picture.

In the meantime I'll refactor to create the healthcheck module (I'm ok with it...). We can later decide what's best adding in case of user specified images. I just think that defaulting to create health checks is ok, since they can be deactivated (and the contrary is not true afaik).

rhuss: You are probably right, it should be the enricher which runs after the generator who is completely responsible for the health checks (at the end is a k8s / openshift concept. However, ill be funny how to deal with the new [Docker Healthchecks](https://docs.docker.com/engine/reference/builder/#/healthcheck) ;-)

Also, not every image is defined by a `Generator`. However the problem what I see is, that an `Enricher` might not have enough information about the images it contains.

So I wonder whether we should extend the `Enricher` API to also get hold of the `ImageConfiguration`s defining the image ? From this the image itself can be deeply inspected like e.g the Docker `HEALTHCHECKS`, exposed ports etc. 

Another idea: To have a generic health-check enricher which checks for an Image Label (or the mentioned Docker HEALTH_CHECKS). This generic health-check enricher is independent of spring-boot, wildfly-swarm etc.

Then:

* spring-boot and widlfly-swarm generator could define the specific health check info (Docker Health Check or Labels or Envs)
* Generic images can create the appropriate info on their own by creating their own labels in the format that the generic enricher understands.

Maybe we can settle on Docker HEALTH_CHECK as this 'marker' as is sufficient generic ? For this to work:

* d-m-p must be improved to support health checks in `ImageConfiguration` (and must take care about backwards compatibility as Docker health checks are quite new).
* `Enricher` must be enhanced to get the list of ImageConfigurations.
* A generic `HealthCheckEnricher` must be written.

That way a health check enricher would only kick in if an image defines a specific health-check, so it's also quite flexible (and close the original app) and might be useful for the image per-se, even when not used in a K8s / OpenShift context.

Still thinking loud a bit. One problem could be that, Docker health_checks are so new that the image format is not supported by K8s or OpenShift. If so, we can still work with a similar label.

nicolaferraro: Thinking about your proposal.. So the reason why you would like to move the logic of health checks from enrichers to generators is because health checks are becoming a lower level concept, because they are going to be embedded in containers. And I agree.

First of all we should check if the new image format is backward compatible with old "runtimes". If it is, there should be no problems for adding support for health checks in d-m-p, since it should be aware of the "builder" version. The new health check string could have been implemented as a label and let the engine do the rest (just supposing, did not check). The best form of communication between generator and enrichers is letting enrichers inspect images, so they can create probes even for images with bundled health checks, also outside the range of fabric8. We should check how to do it.

The second problem is that docker health checks seem almost equivalent to kubernetes exec liveness probes. They don't have the readiness probe concept, and don't have a http get equivalent. The fabric8 labels should fill these gaps.

So, in order to proceed, given that only one standard enricher is expected to survive the natural selection, I would merge the two health check enrichers into eg. the fabric8 enricher project and fix the spring-boot issue that was the subject of this PR :)
(It is blocking some activities).

I'll start to work on a new model for health checks, tracking things in a specific issue. Wdyt?

rhuss: Sounds good ! 

Having a single generic healthcheck features would be really, but it is by no way normative. If we encounter any problems, we should go back to specific health check enrichments. Nothing is worse then over optimised abstractions which at the end blows up anyway. 

But from my feeling I really think the image generator (or the one specifying the image) knows already how the health checks should look like. 

You are right, w.r.t to readiness probes there is no such concept in Docker (that also why d-m-p uses so called `<wait>` configs to check the startup of a container when using `docker:run` (not used in f-m-p)). But as you said, labels could fill the gap. 

It would be awesome if you could work on this. For the momemnt I'm quite happy with this PR and would like to merge it since it is very valuable and usable already. So we can work on top of this.

ok for you ?

nicolaferraro: Sure, I just started doing some tests..

rhuss:
@nicolaferraro will finally integrate the PR as soon as I'm (which is tomorrow)

nicolaferraro:
@rhuss ping :)

nicolaferraro:
opening a new pr

PR #512, opened 2016-10-07
lets detect watch mode and let the spring boot generator automatically generate the remote token

jstrachan:


generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -52,6 +60,40 @@ public boolean isApplicable(List<ImageConfiguration> configs) {
     }
 
     @Override
+    public List<ImageConfiguration> customize(List<ImageConfiguration> configs) throws MojoExecutionException {
+        // if enabled lets generate a spring devtools token
+        String flag = getProjectProperty("fabric8.spring-devtools.generateToken", null);
```
rhuss: Where is this property coming from, i.e who is setting this ?

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -52,6 +60,40 @@ public boolean isApplicable(List<ImageConfiguration> configs) {
...
+        }
+        return super.customize(configs);
+    }
+
+    private void generateSpringDevToolsToken() throws MojoExecutionException {
+        Properties properties = MavenUtil.getSpringBootApplicationProperties(getProject());
+        String remoteSecret = properties.getProperty(DEV_TOOLS_REMOTE_SECRET);
+        if (Strings.isNullOrEmpty(remoteSecret)) {
+            String newToken = UUID.randomUUID().toString();
+            log.info("Generating the spring devtools token: " + newToken);
```
rhuss: Since this info is normally not of much value for the daily usage (correct me when I'm wrong)  maybe its better to move this to level 'verbose', i.e. `log.verbose()` which can be enabled with `-Dfabric8.verbose`. Less noisy than debug, and helps in finding errors.

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -52,6 +60,40 @@ public boolean isApplicable(List<ImageConfiguration> configs) {
...
+
+            File file = new File(getProject().getBasedir(), "target/classes/application.properties");
+            file.getParentFile().mkdirs();
+
+            String text = "# lets configure the spring devtools remote secret\nspring.devtools.remote.secret=" + newToken + "\n";
+
+            if (file.exists()) {
```
rhuss: Shouldn't we also check whether this property not already exists in the file and update it in case ?

plugin/src/main/java/io/fabric8/maven/plugin/PushMojo.java
```diff
@@ -95,8 +100,8 @@ protected String getLogPrefix() {
                 generator != null ?
                     generator :
                     ProfileUtil.extractProcesssorConfiguration(ProfileUtil.GENERATOR_CONFIG, profile, resourceDir);
-            return GeneratorManager.generate(configs, generatorConcfig, project, log, mode, buildStrategy, false);
-        } catch (IOException e) {
+            return GeneratorManager.generate(configs, generatorConcfig, project, session, goalFinder, log, mode, buildStrategy, false);
+        } catch (Exception e) {
```
rhuss: Here we could leave the IOException if customize() would throw the IOException as described above.

Alternatively I would catch `IOException | MavenExecutionException`, not plain `Exception`.

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -52,6 +60,40 @@ public boolean isApplicable(List<ImageConfiguration> configs) {
...
+                text = "\n" + text;
+            }
+            try (FileWriter writer = new FileWriter(file, true)) {
+                writer.append(text);
+            } catch (IOException e) {
```
rhuss: I would simply throw the IOException here (adapting the customize() interface). That way we dont need to add later a catch on `Exception` and simply can leave the catch on `IOException`.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -581,8 +576,8 @@ private HasMetadata convertKubernetesItemToOpenShift(HasMetadata item) {
                 @Override
                 public List<ImageConfiguration> customizeConfig(List<ImageConfiguration> configs) {
                     try {
-                        return GeneratorManager.generate(configs, extractGeneratorConfig(), project, log, mode, buildStrategy, useProjectClasspath);
-                    } catch (IOException e) {
+                        return GeneratorManager.generate(configs, extractGeneratorConfig(), project, session, goalFinder, log, mode, buildStrategy, useProjectClasspath);
+                    } catch (Exception e) {
```
rhuss: see above

generator/api/src/main/java/io/fabric8/maven/generator/api/MavenGeneratorContext.java
```diff
@@ -28,15 +31,18 @@
...
+    private final MavenSession session;
+    private final GoalFinder goalFinder;
     private final ProcessorConfig config;
-
     private final Logger log;
     private final PlatformMode mode;
     private final OpenShiftBuildStrategy strategy;
 
-    public MavenGeneratorContext(MavenProject project, ProcessorConfig generatorConfig, Logger log,
+    public MavenGeneratorContext(MavenProject project, MavenSession session, GoalFinder goalFinder, ProcessorConfig generatorConfig, Logger log,
```
rhuss: Couldn't we feed in a `BuildMode` (like `watch` or `dist`) directly into the context or do the actual look in an `init()` step, exposing only this mode. So that the generator itself doesn't need todo this. Its also more the task of the surrounding execution environment to provide this information then for letting the generator doing the lookup.  Also the generator is then not coupled to specific goal names.

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -52,6 +60,40 @@ public boolean isApplicable(List<ImageConfiguration> configs) {
     }
 
     @Override
+    public List<ImageConfiguration> customize(List<ImageConfiguration> configs) throws MojoExecutionException {
+        // if enabled lets generate a spring devtools token
+        String flag = getProjectProperty("fabric8.spring-devtools.generateToken", null);
+        boolean generateToken = !Strings.isNullOrEmpty(flag) && Configs.asBoolean(flag);
+        if (generateToken || getContext().runningWithGoal("fabric8:watch-spring-boot", "fabric8:watch")) {
```
rhuss: as said above, wouldn't do the lookup here but from the outside, injecting the build mode so to use a `getContext().getBuildMode()` here.

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
jstrachan: thats what the code above does ;)

jstrachan: we assume that if the property exists in the users properties file then we just use it - there's no need to change or regenerate it if the user takes ownership of configuring it

plugin/src/main/java/io/fabric8/maven/plugin/PushMojo.java
jstrachan: we'd need a new docker-maven-plugin release to make that change

jstrachan:
lgtm [merge]

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
rhuss: ah ok :)

plugin/src/main/java/io/fabric8/maven/plugin/PushMojo.java
rhuss: ok, let me look into this.

rhuss:
Thanks ;-) ! I will have a  look how to optimise the remaining bits if you don't mind.

PR #514, opened 2016-10-07
Fix for OSFUSE-357: Do better PATH/executable detection on windows.

chirino:


core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
```diff
@@ -123,6 +127,14 @@ protected static void processOutput(Logger log, InputStream inputStream, Functio
...
         List<File> pathDirectories = getPathDirectories(log);
         for (File directory : pathDirectories) {
+            if( isWindows() ) {
+                for (String extension : new String[]{".exe", ".bat", ".cmd"}) {
+                    File file = new File(directory, name+extension);
+                    if (file.exists() && file.isFile()) {
+                        return file;
+                    }
+                }
+            }
```
rhuss: Shouldn't there be an `else` ?

core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
```diff
@@ -156,7 +168,14 @@ private static String canonicalPath(File file) {
 
     private static List<File> getPathDirectories(Logger log) {
         List<File> pathDirectories = new ArrayList<>();
-        String pathText = System.getenv("PATH");
+        HashMap<String, String> envs = new HashMap<>(System.getenv());
+        if( isWindows() ) {
+            // On windows, the PATH env var is case insensitive, force to upper case.
+            for (Map.Entry<String, String> entry : System.getenv().entrySet()) {
```
rhuss: why convert the whole environment when only the path is needed ?

```java
String getPath() {
  ...
  if (entry.getKey().toUpperCase().equals("PATH") {
      return entry.getValue()
  }
}
```

core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
```diff
@@ -123,6 +127,14 @@ protected static void processOutput(Logger log, InputStream inputStream, Functio
     public static File findExecutable(Logger log, String name) {
         List<File> pathDirectories = getPathDirectories(log);
         for (File directory : pathDirectories) {
+            if( isWindows() ) {
```
apupier: I would prefer to search first on the exact name and then with the extra extensions.

core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
```diff
@@ -123,6 +127,14 @@ protected static void processOutput(Logger log, InputStream inputStream, Functio
...
+                for (String extension : new String[]{".exe", ".bat", ".cmd"}) {
+                    File file = new File(directory, name+extension);
+                    if (file.exists() && file.isFile()) {
```
apupier: the "executability" is not checked for Windows

core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
apupier: you're right

seems that the best would be to be portable and use a java library

core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
apupier: image
there is a "read & execute" permission so I suppose that it is the one used

core/src/main/java/io/fabric8/maven/core/util/ProcessUtil.java
rhuss: yes, makes sense.

rhuss:
@chirino could you please rebase so that we can merge it ?

chirino:
rebased and squashed.

rhuss:
lgtm [merge]

PR #527, opened 2016-10-10
lets remove the actual apps from the app catalog so the json/yaml just includes the catalog only

jstrachan:


jstrachan:
lgtm [merge]

PR #533, opened 2016-10-11
lets by default add a PVC chmod init container annotation for kubernetes clusters to try chmod the PV so that it can be used correctly; otherwise ignore for openshift

jstrachan:


jstrachan:
lgtm [merge]

rhuss:
@jstrachan @rawlingsj Sorry, I complete missed this PR. 

What is the exact reason why a chmod 777 is required on the mounted directory ? Doesn't the init container run under the same user as the pods later, so if he can do a chmod on the dir then the main pod itself can write on it anyway, too ? What is the default mode of the mounted directories from Kubernetes ?

For me this feels strongly like a hackish workaround which we apply to every volume mount, but shouldn't that be fixed within Kubernetes ?

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -281,7 +302,121 @@ private boolean isOpenShiftMode() {
...
+            PodSpec podSpec = template.getSpec();
+            if (podSpec != null) {
+                Map<String, String> nameToMount = new TreeMap<>();
+                List<Volume> volumes = podSpec.getVolumes();
+                if (volumes != null) {
+                    for (Volume volume : volumes) {
+                        PersistentVolumeClaimVolumeSource persistentVolumeClaim = volume.getPersistentVolumeClaim();
+                        if (persistentVolumeClaim != null) {
+                            String name = volume.getName();
+                            nameToMount.put(name, "");
```
rhuss: Why do you populate the Map with an empty path ? afais it is only used to detect whether there are PVC which should be handled. If such a volume isn't refered to in the volume mount sections below (line 364), this empty string will end up in the chmod command (since it doesnt get overwritten), causing very likely an error.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -281,7 +302,121 @@ private boolean isOpenShiftMode() {
...
+                    StringBuffer mountPaths = new StringBuffer();
+                    StringBuffer mounts = new StringBuffer();
+                    int idx = 0;
+                    for (Map.Entry<String, String> entry : nameToMount.entrySet()) {
+                        String name = entry.getKey();
+                        String mountPath = entry.getValue();
+                        mountPaths.append(", \"" + mountPath + "\"");
+
+                        String separator = ++idx >= nameToMount.size() ? "" : ",";
+                        mounts.append("                    {\n" +
```
rhuss: Can we avoid the newlines and indentation ? The do not end up very well in the YAML anyway and have no purpose then. Maybe we could even use a String.format on a previously defined constant here.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -281,7 +302,121 @@ private boolean isOpenShiftMode() {
...
+                            "                \"name\": \"init\",\n" +
+                            "                \"image\": \"busybox\",\n" +
+                            "                \"command\": [\"chmod\", \"777\"" + mountPaths + "],\n" +
+                            "                \"volumeMounts\": [\n" + mounts +
+                            "                ]\n" +
+                            "            }\n" +
+                            "        ]";
+
+                    String initContainerAnnotation = "pod.alpha.kubernetes.io/init-containers";
+                    log.verbose("Adding annotation: " + initContainerAnnotation + " = " + pvAnnotation);
```
rhuss: Debug would be probably better here.

plugin/src/main/java/io/fabric8/maven/plugin/ResourceMojo.java
```diff
@@ -281,7 +302,121 @@ private boolean isOpenShiftMode() {
...
+                                "                        \"name\": \"" + name + "\",\n" +
+                                "                        \"mountPath\": \"" + mountPath + "\"\n" +
+                                "                    }" + separator + "\n");
+                    }
+
+                    String pvAnnotation = "[\n" +
+                            "            {\n" +
+                            "                \"name\": \"init\",\n" +
+                            "                \"image\": \"busybox\",\n" +
+                            "                \"command\": [\"chmod\", \"777\"" + mountPaths + "],\n" +
```
rhuss: As mentioned previously, don't really understand why this needed. Any reference to an issue or so ?

rhuss:
Other question: Is this required for a specific Kuberenetes setup only (e.g. only for minikube) ? Or is this solution required for every Kubernetes installation ?

If this differs from installation to installation its even more questionable whether to put this workaround into resource descriptors which are released and meant to be generally applicable independent on where they are applied.

rhuss:
Shouldn't it be possible to align the permission by using the proper GID on volumes ? Not really deep in the K8s security means, though. Just started to read http://kubernetes.io/docs/user-guide/security-context/ ...

jimmidyson:
Yeah this is very dependent on volume type. Some provisioners will properly handle ownership, etc but others don't. I think the original reason for this was for host path PVs which are only used on development environments, important but not everything of course.

rawlingsj:
A bit of background; this was raised by the Stackpoint folks when deploying fabric8 with persistence on AWS and GKE.  They suggested this approach which has worked well.  If there's another better approach then that's great but as long as it's tested and works on AWS + GKE via Stackpoint and manual installs.

rhuss:
So the init container runs always as root but e.g. gogs runs as [UID 10000](https://github.com/fabric8io/docker-gogs/blob/master/Dockerfile#L22) (is this for OpenShift) ?

For me it stills feels like a dirty workaround for these three use cases. With 'manual' install you mean minikube ? Or also any other kind of cluster setups ? Doest it work with NFS or other volume providers ?

Sorry for being dogged, but it still feels for me like an ugly brute-force workaround for a scenario we don't really understand and which we are going to apply globally. E.g. couldn't we use the GID for the volumes and use a `securityContext:` within the pods templates ? (not really sure what I'm saying here, though :)

If we add this we should make it at least more explicit configurable and use something different as config variable than `kubernetesInitContainerChMod` (e.g. something  like `mountWorldWritable` or so)

rhuss:
Does the `chmod` go through to the underlying volume provider ? (e.g. the host path if `hostPath` is used, so that the permission are changed on the host to `777` for this directory).

rhuss:
To generalize the discussion a bit: Should we include into the resource descriptors that we create during a *build* also informations targeted to specific *runtime aspects* ?. 

K8s does a nice job in separating PV (runtime) and PVC (app) so that they are decoupled. IMO what we are doing here is mangling with the runtime always. The proper solution in my eyes would be to specify permissions and owners within a PVC (not possible yet I guess ?)

Another solution for the current problem e.g. with Gogs would be to specify that the Pods in the Kubernetes case should run as `root` (don't know if this possible within a `Deployment` object).

rawlingsj:
> With 'manual' install you mean minikube ? Or also any other kind of cluster setups ? Doest it work with NFS or other volume providers ?

Yeah I meant a Kubernetes cluster on AWS and GKE that isn't provisioned by Stackpoint.  The two should be the same but I just wanted to be sure that any changes to the approach they suggested doesn't break fabric8 deployments via Stackpoint.

Also just to highlight that this works for all PVCs.  When fabric8 is initially deployed then Nexus, Gogs, Jenkins need this but also any of the one click Runnable apps from fabric8.  At the moment that includes ElasticSearch, Taiga, Prometheus (I think) and many more to come.

If this isn't an acceptable approach then it sounds like there will need to be a better understanding of different options.  I guess what I'm suggesting is when looking at the options be aware that it may be different for each persistent implementation and cloud provider  You'll probably want to test each of them.

rhuss:
If I understand the docs right (well, its understand them, even wrong ;), then IMO this would be the proper solution to let a user other then root to write to a mounted PV:

```yaml
securityContext:
  fsGroup: 10000
```

if the user is set to UID 10000 in the `Deployment`. See e.g. also https://github.com/kubernetes/kubernetes/issues/2630#issuecomment-231848919

Is this something we could consider ?

BTW, another technical point: I would love to add such stuff always to **enrichers**, since they can be easily switched on and off and even profiles can be created which include particular enricher (or not). So we could have a 'stackpoint' profile or a 'devops' profile which could include the very specific workarounds for these use cases.

Ok, let me extract this into an enricher, guess it makes more sense. I will also create a 'fabric8-devops' profile which should be enabled in all devops apps and where we then can be as hackish as we want.

jimmidyson:
So from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L2143-L2153:

```go
// A special supplemental group that applies to all containers in a pod.
// Some volume types allow the Kubelet to change the ownership of that volume
// to be owned by the pod:
//
// 1. The owning GID will be the FSGroup
// 2. The setgid bit is set (new files created in the volume will be owned by FSGroup)
// 3. The permission bits are OR'd with rw-rw----
//
// If unset, the Kubelet will not modify the ownership and permissions of any volume.
```

So this will work for _some_ volume types, but not all.

rhuss:
@jimmidyson so if the kubelet is not allowed to change the ownership for the other volume types, why should it then be possible for an init container changing permissions with `chmod` ?

jimmidyson:
I don't know enough about it to answer tbh.

PR #540, opened 2016-10-12
set destfile executable when copying

sebastienblanc:


rhuss:
Thanks ! We should be good now for the moment, but we consider to directly download into the target directory and also do checksum verification.

lgtm [merge]

PR #546, opened 2016-10-13
fixes #542 so that we can add dependencies in the plugin dependencies to avoid adding duplicate transitive dependencies also lets warn if we find duplicate resources - when not making app-catalog when we expect them added a transient location annotation so that when we get duplicates we can figure out why and where they came from finally an optional merge when we want to merge local fragments with a classpath dependency (versus the generator's resources)

### LINKED ISSUE #542: avoiding duplicate kubernetes resource dependencies
right now we use regular `<dependency>` relationships so that we can compose things together; e.g. `cd-pipeline` depends on `gogs`, `jenkins` etc.

However due to transitive dependencies, when the `DependencyEnricher` kicks in, it sees `cd-pipeline` (which contains gogs) and also `gog` so it ends up duplicating all the resources twice.

I wonder if we can help trim dependency trees; so when we create `cd-pipeline` jar its not really transitively dependent on `gogs` & `jenkins` etc.

Not 100% sure the ideal approach. We could use `provided` or `test` scope for these kinds of dependencies? Not totally sure which; either should hopefully limit the scope and reduce this issue.

jstrachan:


jstrachan:
lgtm [merge]

jstrachan:
looks like hub doesn't like mult-line commit messages for PR names :)

PR #561, opened 2016-10-14
OSFUSE-367: Add a simple fabric8.skip mvn property to control skipping fabric8 goals.

chirino:


rhuss:
An update of the documentation would be also awesome. Needs not to be complete, a simple `_global_configuration.adoc` in doc/src/main/asciidoc/inc would be enough, just that we don't forget collect the global options.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -234,13 +228,8 @@
     public ResourceMojo() {
     }
 
-    @Override
-    public void execute() throws MojoExecutionException, MojoFailureException {
-        clusterAccess = new ClusterAccess(namespace);
```
rhuss: There was a reason to put it that early and not into `executeInternal()`. Just cant remember it but will check it ...

rhuss: I knew there was a reason, but as it is now it indeed doesn't make sense to have this here. I blame amnesia for this ;-)

lgtm [merge]

rhuss:
lgtm

rhuss:
[merge]

PR #581, opened 2016-10-20
fixes #580 to support app catalog and packaging for openshift too

### LINKED ISSUE #580: lets make the packaging of jars work properly for OpenShift builds
right now the DependencyEnricher just repackages all the kubernetes.yml so we lose the OpenShift specific resources and have unnecessary ConfigMaps for app catalog.

This should mean we'll have a single YAML we can use `oc apply -f ` with to then install fabric8 into OpenShift (assuming roles are setup etc)

jstrachan:


jstrachan:
lgtm [merge]

PR #586, opened 2016-10-21
Reading spring-boot configuration from yaml files

nicolaferraro:
This is related to #509.


nicolaferraro:
> Thanks again, and thanks for bearing with my sometimes picky review style ;-)

I like your review style! I'm new, and I'm here to learn :)

PR #589, opened 2016-10-21
Fixing all profiles

nicolaferraro:
The `raw` and `fmp-spring-boot` profiles generated a NullPointerException.
The `internal-microservice` profile was disabling the wrong enricher.

The NPE was caused by the missing generator config in the declaration. Now, when no include/exclude rules for generators or enrichers are provided, uses the default list (ProcessorConfig.FULL).

I added an example of deployment using yaml files only based on the `raw` profile.

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
+++ b/core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
@@ -31,6 +31,8 @@ public class ProcessorConfig {
 
     public static final ProcessorConfig EMPTY = new ProcessorConfig();
 
+    public static final ProcessorConfig FULL = new ProcessorConfig(null, new TreeSet<String>(), null);
+
     /**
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
```diff
@@ -31,6 +31,8 @@
 
     public static final ProcessorConfig EMPTY = new ProcessorConfig();
 
+    public static final ProcessorConfig FULL = new ProcessorConfig(null, new TreeSet<String>(), null);
+
```
rhuss: BTW, as far as I see, the `EMPTY` and `FULL` above are doing the same: Including everything. I just introduced a 'INCLUDES_ALL' and 'EXCLUDES_ALL' to make it more easier to understand (although `EXCLUDES_ALL` is not really used).

rhuss:
Has been already added.

PR #604, opened 2016-10-25
Requiring the GenericWebApplicationContext class before adding health…

nicolaferraro:
… checks

Health checks should be enabled when the actuator is available and the current project is a web application.

A necessary condition for being in a web application is the presence of the `GenericWebApplicationContext` class: 
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/OnWebApplicationCondition.java#L40-L41

This way, if eg. `spring-boot-starter-web` is not added, the health checks are not created.


jstrachan:
lgtm [merge]

Issue #517, opened 2016-10-10
Support for PetSet and DaemonSet deployment options


ucosty:
I'm working on a project that requires the use of both Pet Sets and Daemon Sets for certain parts of the application. Trying to generate these using YAML fragments produces errors like the following

> resource failed: Unknown type 'daemonset' for file
> resource failed: Unknown type 'petset' for file

rhuss:
Thanks, we will pick that up into the list of recognized types.

In the meantime you could name your fragments without type (e.g. `name.yml`) and set a 'kind: PetSet' within the file.

ucosty:
Thanks for the speedy response, I had no idea the error was coming from the filename.

ucosty:
I may have spoken too soon. Daemon sets seem to work properly, but the PetSet type is not liked at all.

> [ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.1.75:resource (default-cli) on project consul: Execution default-cli of goal io.fabric8:fabric8-maven-plugin:3.1.75:resource failed: No resource type found for kind:PetSet
> [ERROR] at [Source: N/A; line: -1, column: -1]
> [ERROR] -> [Help 1]
> org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:fabric8-maven-plugin:3.1.75:resource (default-cli) on project consul: Execution default-cli of goal io.fabric8:fabric8-maven-plugin:3.1.75:resource failed: No resource type found for kind:PetSet
>  at [Source: N/A; line: -1, column: -1]
> 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
> 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
> 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
> 	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
> 	at io.takari.maven.builder.smart.SmartBuilderImpl.buildProject(SmartBuilderImpl.java:334)
> 	at io.takari.maven.builder.smart.SmartBuilderImpl$ProjectBuildTask.run(SmartBuilderImpl.java:104)
> 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[...]

PR #620, opened 2016-10-30
Issue #517: PetSet and DaemonSet support

ucosty:
I've had some time to work on issue #517, adding support for generating PetSet and DaemonSet configurations using the fabric8-maven-plugin. I'll have a go at updating the documentation as well, to reflect some of these changes. 

For now I've been using it by configuring the `fmp-controller` enricher to generate the PetSet and DaemonSet types.

The following snippet of code seems to work for my needs.

```
<enricher>
    <config>
        <fmp-controller>
            <type>petSet</type>
        </fmp-controller>
    </config>
</enricher>
```


rhuss:
Thanks, will have a look ASAP.

core/src/main/java/io/fabric8/maven/core/handler/PetSetHandler.java
```diff
@@ -0,0 +1,50 @@
...
+                .build();
+
+        return petSet;
+    }
+
+    // ===========================================================
+
+    private ObjectMeta createPetSetMetaData(ResourceConfig config) {
+        return new ObjectMetaBuilder()
+                .withName(KubernetesHelper.validateKubernetesId(config.getReplicaSetName(), "replica set name"))
```
rhuss: wdyt, maybe we should rename `replicaSetName` to `controllerName` ?

core/src/main/java/io/fabric8/maven/core/handler/ProbeHandler.java
```diff
@@ -46,6 +47,9 @@ public Probe getProbe(ProbeConfig probeConfig)  {
             probe.setTimeoutSeconds(timeoutSeconds);
         }
         HTTPGetAction getAction = getHTTPGetAction(probeConfig.getGetUrl());
+        if(StringUtils.isEmpty(getAction.getHost())) {
```
rhuss: This check can lead to a NPE since getAction can be null. Could explain briefly why an empty String as host is an issue ?

plugin/src/main/java/io/fabric8/maven/plugin/enricher/MetadataVisitor.java
```diff
@@ -184,4 +185,21 @@ protected ObjectMeta getOrCreateMetadata(DaemonSetBuilder item) {
...
     }
+
+    public static class PetSet extends MetadataVisitor<PetSetBuilder> {
+        public PetSet(EnricherManager enricher) {
+            super(enricher);
+        }
+
+        @Override
+        protected Kind getKind() {
+            return Kind.DAEMON_SET;
```
rhuss: Shouldn't this be `Kind.PET_SET` ?

core/src/main/java/io/fabric8/maven/core/handler/PetSetHandler.java
ucosty: Good idea, I've added that to this PR.

core/src/main/java/io/fabric8/maven/core/handler/ProbeHandler.java
ucosty: It was a bit of a hack to allow an empty HTTP host to be set in the liveness/readiness checks. Without this minor change an empty host value would result in a literal empty host value being added to the output YAML. For now I've just removed this, as it should have been a part of issue #622.

oscerd:
Any news on this one? :-)

plugin/src/main/java/io/fabric8/maven/plugin/enricher/SelectorVisitor.java
```diff
@@ -92,6 +91,44 @@ public void visit(DeploymentSpecBuilder item) {
...
+    static public class PetSet extends SelectorVisitor<PetSetSpecBuilder> {
+
+        public PetSet(EnricherManager enricherManager) {
+            super(enricherManager);
+        }
+
+        @Override
+        public void visit(PetSetSpecBuilder item) {
+            Map<String, String> selectorMatchLabels =
+                    KubernetesResourceUtil.removeVersionSelector(enricherManager.extractSelector(getConfig(), Kind.REPLICATION_CONTROLLER));
```
rhuss: Should be `Kind.PET_SET`

plugin/src/main/java/io/fabric8/maven/plugin/enricher/SelectorVisitor.java
```diff
@@ -92,6 +91,44 @@ public void visit(DeploymentSpecBuilder item) {
...
+    static public class DaemonSet extends SelectorVisitor<DaemonSetSpecBuilder> {
+
+        public DaemonSet(EnricherManager enricherManager) {
+            super(enricherManager);
+        }
+
+        @Override
+        public void visit(DaemonSetSpecBuilder item) {
+            Map<String, String> selectorMatchLabels =
+                    KubernetesResourceUtil.removeVersionSelector(enricherManager.extractSelector(getConfig(), Kind.REPLICATION_CONTROLLER));
```
rhuss: shoud be `Kind.DAEMON_SET` (btw, there is an error for the `Deployment` selector, I will fix that.

rhuss:
If you don't mind I would fix the outstanding issues and integrate this PR next week.

oscerd:
I'd like to see this one merged, to update the cassandra quickstart with a PetSet configuration :-)

rhuss:
@oscerd back right now, will go through the PRs here soon. Happy new year ;-) !

oscerd:
@rhuss happy new year! :-)

Thanks for your work! :-)

ucosty:
Sorry for the delay, this has been updated now.

crashdump:
I would love to see this one merged.

rhuss:
will check that this later today or tomorrow, will merge it this week.

jstrachan:
though `PetSet` is now `StatefulSet` right?

rhuss:
Right, should be changed to `StatefulSet`.

rhuss:
We could support `PetSet` for backwards compatibility, too, but probably its better to start with out baggage. So I opts for supporting only `StatefulSet`.

rawlingsj:
FWIW this PR will bring in the StatefulSet rename https://github.com/fabric8io/kubernetes-client/pull/646, once we're happy with that I was going to upgrade fabric8 and then fabric8-maven-plugin asap.

rhuss:
@rawlingsj thanks a lot for StatefulSet, I'm going to rename it in this PR. Is the kubernetes-client already released with this addition ?

rawlingsj:
> thanks a lot for StatefulSet, I'm going to rename it in this PR. Is the kubernetes-client already released with this addition ?

no problem!  yes if you rebase this PR then you should get the new fabric8 version which includes the updated kubernetes client and model with StatefulSet.

rhuss:
Hmm. I see still fabric8 2.2.193 as the latest referring to kubernetes-client 2.0.0 (but 2.0.2 is the latest which contains the changes).

rhuss:
Let me debug that a bit here, looks a bit weird ... (using a kubernetes client 1.4.33 and so). Probably my fault.

rhuss:
ok, got it now properly. all good :)

rawlingsj:
Looks like there's been a couple more releases but 2.0.0 was the one that I did which includes the statefulset rename.  So hopefully you should be good.

rawlingsj:
> ok, got it now properly. all good :)

ah cool :)

rhuss:
I updated from `PetSet` to `StatefulSet` (thanks to @rawlingsj  for the upstream kubernetes-client update !).

Changes look good to me. Some tests would have been nice but due to the fact that we don't have tests for the other handlers, too, its not fair to call for one here now ;-)

rhuss:
[merge]

PR #623, opened 2016-10-31
OSFUSE-391: f-m-p annotates services with prometheus.io annotations by default

lburgazzoli:


enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/PrometheusEnricher.java
```diff
@@ -0,0 +1,84 @@
...
+                    break;
+                }
+            }
+        }
+
+        return prometheusPort;
+    }
+
+    private List<ImageConfiguration> getImageConfigurations() {
+        List<ImageConfiguration> images = getContext().getImages();
```
rhuss: images should be always non-null (i.e. an empty list if nothing is configured)

lburgazzoli:
Yes, I think it would be nice to have Prometheus related stuff in a single place but this could be generic enough so that if we add new stuffs, they can be handled the same way.

I'm wondering if it could make sense to have a programmatic way to hook enrichers to a phase/event/object so that an enricher is triggered just after a service has been created.

lburgazzoli:
Added some small unit tests

lburgazzoli:
Added minimal dock

rhuss:
Thanks ! will merge it that evening. 

w.r.t to enricher hooks: This is definitely something to consider. Up to now a lot of responsibility is put on the enricher to not mess up things. Maybe a bit more controlled flow would help, e.g. by splitting up into to phases as said, and then, when thing get created maybe some hooks. However, when enricher are modifying or 'enriching' existing objects, they possible want to do this on freshly created or already existing/configured objects.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/PrometheusEnricher.java
```diff
@@ -0,0 +1,78 @@
...
+    public PrometheusEnricher(EnricherContext buildContext) {
+        super(buildContext, ENRICHER_NAME);
+    }
+
+    @Override
+    public Map<String, String> getAnnotations(Kind kind) {
+        if (kind.isService()) {
+            String prometheusPort = findPrometheusPort();
+            if (Strings.isNotBlank(prometheusPort)) {
+                log.info("Add prometheus.io annotations: %s=%s, %s=%S",
```
rhuss: Can you switch this to `log.verbose()` please ? (can be enabled by setting `-Dfabric8.verbose` or `<verbose>true</verbose>`. The reason is, that typically all concrete actions (like the commands in a Dockerfile) are logged with this level. We should probably enhance the interface a bit to provide more context information like the service name. but that's another story ;-)

`info()` should be reserved for important steps like 'Calling enrichers'.

rhuss:
[merge]

rhuss:
thanks !

Issue #622, opened 2016-10-31
Can't add liveness and readiness health checks using inline XML


ucosty:
### Description

When generating the Kubernetes configuration, the liveness and readiness health checks are not being added to deployment.

The following XML snippet, in theory, should add a liveness health check to the deployment.

```
<resources>
    <liveness>
        <initialDelaySeconds>30</initialDelaySeconds>
        <timeoutSeconds>3</timeoutSeconds>
        <getUrl>HTTP://:8500/v1/status/peers</getUrl>
    </liveness>
</resources>
```

Running `mvn fabric8:resource` failes to produce the appropriate `livenessProbe` configuration.

I had a look through the code, while it looks like the configuration is being parsed there doesn't seem to be a way of adding the configuration to the appropriate deployment/petset/pod.

I think the solution is a health check enricher, which takes the configured liveness and readiness resources and attaches the configuration as required.

### Info

* f-m-p version :
* Maven version (`mvn -v`) :
```
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T16:41:47+00:00)
Maven home: /usr/local/Cellar/maven/3.3.9/libexec
Java version: 1.8.0_45, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_45.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.10.5", arch: "x86_64", family: "mac"
```
* Kubernetes / OpenShift setup and version :
* If it's a bug, how to reproduce :
* If it's a feature request, what is your use case :
* Sample project : https://github.com/ucosty/kubernetes-consul-test.git

ucosty:
I feel I should have dug through the existing issues a bit before raising this, as this can likely be covered by a combination of Issue #600 and Issue #601. Perhaps the work done in Pull Request #624 can be extended to support these scenarios?

PR #624, opened 2016-10-31
Issue #622: Can't add liveness and readiness health checks using inline XML

ucosty:
In order to configure Liveness and Readiness health checks using inline XML, I've written an enricher which grabs this configuration and applies it to any configured Deployments, DaemonSets and PetSets. I've also updated the documentation to help in it's usage.

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/HealthCheckEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/HealthCheckEnricher.java
@@ -1,0 +1,109 @@
+package io.fabric8.maven.enricher.standard;
+
+import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
+import io.fabric8.kubernetes.api.model.Probe;
+import io.fabric8.kubernetes.api.model.extensions.DaemonSet;
+import io.fabric8.kubernetes.api.model.extensions.Deployment;
+import io.fabric8.kubernetes.api.model.extensions.PetSet;
+import io.fabric8.maven.core.config.ProbeConfig;
+import io.fabric8.maven.core.handler.ProbeHandler;
+import io.fabric8.maven.enricher.api.AbstractHealthCheckEnricher;
+import io.fabric8.maven.enricher.api.BaseEnricher;
+import io.fabric8.maven.enricher.api.EnricherContext;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by matthew.costa on 26/10/2016.
+ */
+public class HealthCheckEnricher extends BaseEnricher {
+
+    private ProbeConfig liveness;
+    private ProbeConfig readiness;
+    private ProbeHandler probeHandler;
+
+    public HealthCheckEnricher(EnricherContext buildContext) {
+        super(buildContext, "fmp-healthcheck");
+
+        probeHandler = new ProbeHandler();
+        liveness = buildContext.getResourceConfig().getLiveness();
+        readiness = buildContext.getResourceConfig().getReadiness();
+    }
+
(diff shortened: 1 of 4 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/HealthCheckEnricher.java
```diff
@@ -0,0 +1,109 @@
...
+        readiness = buildContext.getResourceConfig().getReadiness();
+    }
+
+    @Override
+    public void addMissingResources(KubernetesListBuilder builder) {
+        List<HasMetadata> items = builder.getItems();
+        for (HasMetadata item : items) {
+            List<Container> containers = null;
+
+            if(item instanceof Deployment) {
```
rhuss: Question is whether we want to allow that for ReplicationController and ReplicaSet, too ? (and how to work with OpenShift DeploymentConfigs, when given here). I guess we should add the health-checks here, too.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/HealthCheckEnricher.java
```diff
@@ -0,0 +1,109 @@
...
+import io.fabric8.maven.enricher.api.BaseEnricher;
+import io.fabric8.maven.enricher.api.EnricherContext;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by matthew.costa on 26/10/2016.
+ */
+public class HealthCheckEnricher extends BaseEnricher {
```
rhuss: I'm  not sure whether we should do this as an enricher, or as core part of the `ResourceMojo` plugin. The reason is, that the order of enricher matters, who will 'win', (see e.g. the more specialized enricher like `SpringBootHealthCheckEnricher`). If this functionality were independent of an enricher, then we can e.g. say: "A health check given in the configuration takes always precedence". Of course we can configure enrichers that way that this enricher has the highest priority, but that can easily break.

Another issue is, if you configure a health-check in the plugin configuration but *disable* this enricher (e.g. because of using another profile or exclude it in the enricher config) then this can lead to surprises, because someone would expect that its liveness  / readiness check configured in XML would be respected for sure.

In general I think that the code evaluating XML configuration from the plugin should go into the ResourceMojo (which tbh needs a refactoring). Said that, the XML configuration is not yet really very matured, but this is one of the next steps.

rhuss:
Maybe we can handle it like the `MetaDataEnricher`. This is also an enricher accessing the XML configuration but is added programmatically to the beginning of the pipeline. We could go even that far to provide a dedicated enricher interface for those 'internal' enricher modules which get the `ResourceConfig` while normal enricher does not.

rhuss:
I'm just about to remove the MetaDataEnricher which were used internally. If you don't mind I'm going to implement this functionality as core part of the ResourceMojo, independent of an enricher.

rohanKanojia:
@rhuss : Do we have any issue filed for this PR? If not, let's file an issue so that someone can pick this up and implement it in a cleaner way like you have requested.

rhuss:
Actually I wonder whether a single liveness/readiness check should be added to all Deployment and DeploymentConfigs as they are very specific to the container included in its templates.

So currently I would suggest to use fragments for the health check configuration.

With respect to the XML configuration we should reconsider whether we really want to fully support this, as people nowadays tend to prefer the native syntax (e.g. Dockerfiles for building or Kubernetes Resources for resource definitions). This was not the case when the plugin started when these native formats were not so well known. 

So my tendency would be limit XML configuration only to the bare minimum, but this is another discussion. So adding  health checks via XML configuration should then discussed in the context of that larger picture.

I'm going to close this PR for now, as we shouldn't use the enricher abstraction to evaluate XML syntax anyway. @ucosty sorry for the super long delay and thans a lot for the contribution. I'm confident we find a good solution in the feature when jumping to 4.0 (and as mentioned you already can use fragments for adding healthchecks)

PR #628, opened 2016-11-02
work around regression in kubernetes-model breaking S2I binary builds https://github.com/fabric8io/kubernetes-model/issues/173 also add a retry in case we hit timing issues

jstrachan:


jstrachan:
lgtm [merge]

PR #639, opened 2016-11-07
Fix for OSFUSE-425: First arg of log methods should be String format constant string.

chirino:
Fixed up lots calls but needs more review.


rhuss:
Thanks a lot ! lgtm

[merge]

PR #641, opened 2016-11-08
Merge in the 3.1.80.redhat branch

chirino:
Allows for easy subsequent merges from the 3.1.80.redhat branch.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -22,12 +22,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>fabric8-maven-parent</artifactId>
-    <version>3.1.80.redhat-SNAPSHOT</version>
+    <version>3.1-SNAPSHOT</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
   <artifactId>fabric8-maven-core</artifactId>
-  <version>3.1.80.redhat-SNAPSHOT</version>
+  <version>3.1-SNAPSHOT</version>
 
   <name>Fabric8 Maven :: Core</name>
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -22,12 +22,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>fabric8-maven-parent</artifactId>
-    <version>3.1-SNAPSHOT</version>
+    <version>3.1.80.redhat-SNAPSHOT</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
   <artifactId>fabric8-maven-core</artifactId>
-  <version>3.1-SNAPSHOT</version>
+  <version>3.1.80.redhat-SNAPSHOT</version>
 
   <name>Fabric8 Maven :: Core</name>
 
--- a/doc/pom.xml
+++ b/doc/pom.xml
@@ -21,12 +21,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>fabric8-maven-parent</artifactId>
-    <version>3.1.80.redhat-SNAPSHOT</version>
+    <version>3.1-SNAPSHOT</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
   <artifactId>fabric8-maven-doc</artifactId>
-  <version>3.1.80.redhat-SNAPSHOT</version>
+  <version>3.1-SNAPSHOT</version>
   <packaging>jar</packaging>
 
   <name>Fabric8 Maven :: Documentation</name>
--- a/doc/pom.xml
+++ b/doc/pom.xml
@@ -21,12 +21,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>fabric8-maven-parent</artifactId>
-    <version>3.1-SNAPSHOT</version>
+    <version>3.1.80.redhat-SNAPSHOT</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
   <artifactId>fabric8-maven-doc</artifactId>
-  <version>3.1-SNAPSHOT</version>
+  <version>3.1.80.redhat-SNAPSHOT</version>
   <packaging>jar</packaging>
 
   <name>Fabric8 Maven :: Documentation</name>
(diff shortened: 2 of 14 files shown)


rhuss:
Had just some discussions with @chirino about the purpose of this PR:

tl;dr: The use case is, to allow easy merging of bug fixes from the redhat branch to master without running into conflicts on the version property. Therefor on master the version is set temporarily to the redhat version, then the branch is forked of and finally on master reset to the original version. So the version change is present in both branches and doesn't lead to a conflict. At the end its a no-op on master, but messes up the history for 3 commits. But its only a single time trade off, which doesnt need to be done a second time.

I'm not sure whether this is the best solution for easily merging fixes on the redhat branch back to master.

Questions coming up are:

* Is there a better solution for the use case at hand ?
* What's about getting fixes from master to the branch ?

wdyt, @fabric8io/core ?

```
17:25 ~chirino  so does a PR of a merge work?
17:27 ~chirino  https://github.com/fabric8io/fabric8-maven-plugin/pull/641
17:29 ~chirino  not sure that's a good idea
17:29 ~chirino  we might need to PR into both branches
17:29 ~chirino  or perhaps we can just PR from 3.1.80.redhat into master
17:33  ro14nd sorry, didn't get it. You want to update master from branch 3.1.80.redhat ?
17:34 ~chirino  so if you look at https://github.com/fabric8io/fabric8-maven-plugin/pull/641/files
17:34 ~chirino  you'll see the PR changes nothing on master :)
17:34 ~chirino  but adds a merge point from 3.1.80.redhat
17:34 ~chirino  so that any fixes that go into 3.1.80.redhat  can be merged into master without bringing in all those pom updates
17:38 ~chirino  did that make sense?
17:41  ro14nd hmm, need to think about it (as always when its about Git stuff :)
17:42 ~chirino  so the test is, once you merge that change into master: co the 3.1.80.redhat  branch, make a change into it, then merge that branch into master
17:42 ~chirino  you'll see that just the change is brought in
17:42 ~chirino  and not all those pom updates that exist on the 3.1.80.redhat  branch
17:43 ~chirino  basically I want to make it easy for our product team to fix bugs on the 3.1.80.redhat  branch but also get them into master too.
17:44  ro14nd but i guess that is messes up the git history on master with noop commits (not that we are using the history quite a lot, but some ppl are more picky here).
17:44 ~chirino  Sure.  But in this case it think it's worth it.
17:45 ~chirino  otherwise we are likely to get into situation where fixes go into the 3.1.80.redhat  branch but never make it into master
17:45 ~chirino  cherry picking between branches is a sure way to miss a commit on one or the other
17:55  ro14nd I wonder whether there is an easier way to avoid version conflicts (as it seems that is the only issue).
17:56 ~chirino  should we discuss on #fabric8?
17:56  ro14nd An alternative solution which wouldn't had that impact on master (as master needs to 'know' also about the redhat branch) would beL
17:56  ro14nd be:
17:57  ro14nd * when a fix should be go back to master, branch the redhat-branch
17:57  ro14nd * change the version to a master-conform version number (3.1-SNAPSHOT) via a scipt.
17:57  ro14nd * check this in
17:57  ro14nd * merge
17:58 ~chirino  that sounds crazy
17:58  ro14nd but you are right, we should discuss this on #fabric8
17:58  ro14nd why ?
17:58  ro14nd the script is a single sed script.
17:58 ~chirino  then the changes set is huge
17:58 ~chirino  but this has to be done for every commit no?
17:59  ro14nd no not really if you do a squash on this branch
17:59  ro14nd before the merge
17:59 ~chirino  which branch?
18:00  ro14nd that temporary branch forked of the redhat branch descibed above.
18:00  ro14nd so: branch of redhat-branch, adapt version via script, rebase -i, merge to master.
18:00  ro14nd finally delete this branch
18:01 ~chirino  yeah sounds painful.
18:01 ~chirino  would need a pipeline to do it
18:01 ~chirino  and hopefully it does not mess up.
18:01  ro14nd doesn't look much more complicated at this 3 commit thing.
18:01 ~chirino  this is a 1 time thing
18:01 ~chirino  which I can prove does not change master
18:03 ~chirino  does not require folks to know about special scripts either
18:03 ~chirino  that allow me to merge 3.1.80.redhat into master at any time
18:03 ~chirino  after the 2nd commit or the 500th commit
18:04  ro14nd how long does such a branch live ? Isn't there one also vo 3.1.81.redhat (so you need to do this again ?)
18:05  ro14nd btw, why isnt there a 3.1-redhat branch ?
18:06  ro14nd to understand the technique: You branch off the redhat branch from the second commit (i.e. 388e0d5109afbb4ff569dc644b06dc0bb4b9ba7b in this example), so that the redhat version is in both branch and if they are not changed in any branch later on there wont be a conflict)
18:10  ro14nd Could we discuss this in a github issue (or maybe even on the given PR) ? Would like to hear some other options as it affects the development workflow of this.
18:10  ro14nd If you don't mind i'd copy this discussion here to the PR and ask for some comments.
18:11  ro14nd (most of the guys are probably already offline, jstrachan e.g. is at Devoxx etc.)
18:15 ~chirino  FIS 2.0 will stay on 3.1.81.redhat  for as long as we support FIS 2.0 :)
18:15 ~chirino  we will not add anymore new features to 3.1.81.redhat
18:16 ~chirino  thats why its not a 3.1.redhat
18:16 ~chirino  ro14nd: feel free to copy
18:16 ~chirino  :)
18:17  ro14nd ok, then your solution seems to be fair (if everything from the redhat branch fixes goes into master without leaving out sth ;-)
```

davsclaus:
Is this only for f-m-p? Or would all the other git repos need to do the same?

rhuss:
afaik this is supposed to be a blueprint for the other productised projects as well.

tdiesler:
The proposed solution seems a bit convoluted imho. Everything that is proposed here can likely be done to some branch other than master too and later cherry-picked (with range) to master. Generally, I'd prefer to only merge to master what has not diverged from it too much and ideally can also be rebased onto it (i.e. is derived from master not too long ago).

chirino:
If we  want to have an easy way to merge fixes from a stable branch without worrying we missed something then this the easy git way.

chirino:
[merge][lgtm]

rhuss:
I suggest we try this approach with f-m-p to see how it works. For now is a no-op anyway (except for some history entries).

rhuss:
[merge]

PR #647, opened 2016-11-12
Add generator documentation, refactored spring-boot-generator

rhuss:
The spring-boot-generator now adds the spring devtool.jar directly to the fat-jar instead
of putting a cuckoo egg into target/classes. This also now works directly
when fabric8:build is called (so no need to rely that fabric8:resource kicked in earlier).

Added documentation to all generators with base images chosen and configuration options.


rhuss:
[merge]

PR #652, opened 2016-11-14
Moved init container which chmod for persistent volumes to a dedicated enricher

rhuss:
Moved init container handling to a dedicated enricher


rhuss:
[merge]

PR #653, opened 2016-11-16
OSFUSE-403: Ensure effective platform mode is set when choosing to ru…

jamesnetherton:
…n individual goals

Relates to: https://issues.jboss.org/browse/OSFUSE-403


rhuss:
The fix looks good to me for the current situation, however I think that the usage of this `FABRIC8_EFFECTIVE_PLATFORM_MODE` is only a workaround for a larger problem, namely that the Enricher or Generator interface are not perfectly suited. Still didn't got fully the use case for this property. Especially the communication between Mojos needs to be reduced.

rhuss:
Please see also the discussion of https://github.com/fabric8io/fabric8-maven-plugin/pull/429 why this all happens currently this way.

rhuss:
[merge]

PR #654, opened 2016-11-16
clarify that adding the deploy goal to the f-m-p executions is only i…

christian-posta:
…f you'd like to deploy to kubernetes each time you run mvn install


rhuss:
Actually you should *never* bind `deploy` to an execution as it forks the lifecycle so that you will end up in running goals twice. Instead use `apply` for this.

christian-posta:
thanks @rhuss updated the PR

davsclaus:
@rhuss lgtm. Should we get this merged?

rhuss:
yeah

[merge]

PR #664, opened 2016-11-18
Enable VertxGenerator and allow reading ports from configuration.

iocanel:


rhuss:
test this please

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/VertxGenerator.java
```diff
@@ -35,13 +52,98 @@
...
+                Map<String, String> configMap = OBJECT_MAPPER.readValue(configFile, Map.class);
+                for (Map.Entry<String, String> entry : configMap.entrySet()) {
+                    String key = entry.getKey();
+                    Matcher m = PORT_PATTERN.matcher(key);
+                    if (m.matches()) {
+                        addPortIfValid(answer, String.valueOf(entry.getValue()));
+                    }
+                }
+            } catch (IOException e) {
+                log.warn("Error reading config: [" + configPath + "], due to: [" + e.getMessage() + "]. Ignoring.");
```
rhuss: see above w.r.t to log format

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/VertxGenerator.java
```diff
@@ -35,13 +52,98 @@
...
+
+    private boolean isVertxPackage() {
+        MavenProject project = getProject();
+        Plugin plugin = project.getPlugin(VERTX_MAVEN_PLUGIN_GA);
+        if (plugin != null) {
+            Map<String, PluginExecution> executionsAsMap = plugin.getExecutionsAsMap();
+            if (executionsAsMap != null) {
+                for (PluginExecution execution : executionsAsMap.values()) {
+                    List<String> goals = execution.getGoals();
+                    if (goals.contains("package")) {
```
rhuss: I wonder whether should be more open as the vertx-maven-plugin could also be used when no bound to an execution phase, e.g. when called directly like in `mvn package vertx:package fabric8:build`. I don't know whether it is possible but it would be awesome if it would be possible to find out which goals are actually in effect when this goal is reached. There is a `GeneratorContext.runningWithGoal()` but I don't know whether it really catches also "resolved" calls.

generator/vertx/src/main/resources/META-INF/fabric8/generator-default
```diff
@@ -1 +1 @@
-io.fabric8.maven.generator.javaexec.VertxGenerator,1000
+io.fabric8.maven.generator.vertx.VertxGenerator,1000
```
rhuss: thanks ;-)

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/VertxGenerator.java
```diff
@@ -35,13 +52,98 @@
...
+        List<String> answer = new ArrayList<>();
+        MavenProject project = getProject();
+        Plugin plugin = project.getPlugin(VERTX_MAVEN_PLUGIN_GA);
+        if (plugin != null) {
+            Xpp3Dom configuration = (Xpp3Dom) plugin.getConfiguration();
+            if (configuration == null) {
+                return null;
+            }
+            Xpp3Dom config = configuration.getChild("config");
+            String configPath = config != null ? config.getValue() : System.getProperty(VERTX_CONFIG, null);
```
iocanel: This isn't configuration for generators its `vertx-maven-plugin` configuration. Do we have something that applies to that case too?

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/VertxGenerator.java
```diff
@@ -35,13 +52,98 @@
...
+            if (!configFile.exists()) {
+                log.warn("Could not find config: [" + configPath + "]. Ignoring.");
+                return null;
+            }
+
+            try {
+                Map<String, String> configMap = OBJECT_MAPPER.readValue(configFile, Map.class);
+                for (Map.Entry<String, String> entry : configMap.entrySet()) {
+                    String key = entry.getKey();
+                    Matcher m = PORT_PATTERN.matcher(key);
```
iocanel: Its not too much, its a great idea.

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/VertxGenerator.java
kameshsampath: just my 2 cent, since maven plugin configuration can be at plugin level (global), goal level - it will be great if we can check both the options to make this a bit robust

iocanel:
@rhuss: applied changes for all your comments _(all but the configuration thing, which is not applicable)_.

Let me know if it looks ok and I'll squash.

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/PortsExtractor.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import io.fabric8.maven.core.util.PrefixedLogger;
+import io.fabric8.utils.Strings;
+
+import static io.fabric8.maven.generator.vertx.Constants.*;
+
+public class PortsExtractor {
```
rhuss: If we can get rid of vertx deps, we could put this into a core util package to allow port extraction from arbitrary configuration files. This might be very helpful for other generators, too. Wonder whether to put this into the `core` or `generator.api`.

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/PortsExtractor.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final String PORT_REGEX = "([a-zA-Z0-9_]+)(([\\.-_]+p{1})|([P]{1}))ort";
+    private static final Pattern PORT_PATTERN = Pattern.compile(PORT_REGEX);
+
+    /**
+     * Extracts ports found in the configuration.
+     * @param project   The maven project to use.
+     * @param log       The logger to use.
+     * @return          The list with the detected ports or an empty list.
+     */
+    public static List<String> extract(MavenProject project, PrefixedLogger log) {
```
rhuss: thats the method which would be quite nice to have general available.

generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/PortsExtractor.java
iocanel: Agree

rhuss:
@iocanel thanks, will have a look later.

davsclaus:
Ah nice we need some port extractor as well for wildfly-swarm. And maybe migrate the spring-boot ones as well.

generator/api/src/main/java/io/fabric8/maven/generator/api/support/AbstractPortsExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private static final String DOT = ".";
+    private static final String JSON_EXTENSION = ".json";
+    private static final String YAML_EXTENSION = ".yaml";
+    private static final String YML_EXTENSION = ".yml";
+    private static final String PROPERTIES_EXTENSION = ".properties";
+
+    private static final String NUMBER_REGEX = "\\d+";
+
+    private static final String PORT_REGEX = "([a-zA-Z0-9_]+)(([\\.-_]+p{1})|([P]{1}))ort";
```
rhuss: just fyi: `{1}` can be omitted (its the default anyway ;-)

generator/api/src/main/java/io/fabric8/maven/generator/api/support/AbstractPortsExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+     * Finds the name of the configuration file from the {@link MavenProject}.
+     * @param project   The {@link MavenProject} to use.
+     * @return          The path to the configuration file.
+     */
+    public abstract String getConfigPathFromProject(MavenProject project);
+
+
+    public File getConfigLocation(MavenProject project) {
+        //The system property has priority over what is specified in the pom.
+        String configPath = System.getProperty(getConfigPathPropertyName(), getConfigPathFromProject(project));
```
rhuss: Doesn't `System.getProperty()` throws an NPE on a null key ? Don't know by heart, but maybe that should checked before ?

generator/api/src/main/java/io/fabric8/maven/generator/api/support/AbstractPortsExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+        if (configPath == null) {
+            return null;
+        }
+        return Paths.get(configPath).toFile();
+    }
+
+    @Override
+    public Map<String, Integer> extract(MavenProject project) {
+        Map<String, Integer> answer = new HashMap<>();
+        File configFile = getConfigLocation(project);
```
rhuss: `configFile` can be null (see above), so would throw an NPE in the next line.

generator/api/src/main/java/io/fabric8/maven/generator/api/support/AbstractPortsExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+        if (f.getName().endsWith(JSON_EXTENSION)) {
+            map = flatten(JSON_MAPPER.readValue(f, Map.class));
+        } else if (f.getName().endsWith(YAML_EXTENSION) || f.getName().endsWith(YML_EXTENSION)) {
+            map = flatten(YAML_MAPPER.readValue(f, Map.class));
+        } else if (f.getName().endsWith(PROPERTIES_EXTENSION)) {
+            Properties properties = new Properties();
+            properties.load(new FileInputStream(f));
+            map = propertiesToMap(properties);
+        } else {
+            map = Collections.emptyMap();
```
rhuss: Maybe better to throw an exception here, that the config format (like 'toml' as Vert.x is starting to look in) is not supported, instead of silently failing to extract a port from a given file name.

generator/api/src/main/java/io/fabric8/maven/generator/api/support/AbstractPortsExtractor.java
```diff
@@ -0,0 +1,178 @@
...
+            Object value = map.get(key);
+            if (value instanceof String) {
+                flat.put(stringKey, (String) value);
+            } else if (value instanceof Map) {
+                for (Map.Entry<String, String> entry : flatten((Map) value).entrySet()) {
+                    flat.put(
+                            new StringBuilder(stringKey).append(DOT).append(entry.getKey()).toString(),
+                            entry.getValue());
+                }
+            } else {
```
rhuss: That can be come difficult if a key is called like 'ports', the maybe a list of ports is expected ? (tbh dont know how realistic this is. However a 'ports' property would match the regexp, too).

generator/vertx/src/main/resources/META-INF/fabric8/generator-default
```diff
@@ -1 +1 @@
-io.fabric8.maven.generator.vertx.VertxGenerator,1000
+io.fabric8.maven.generator.vertx.VertxGenerator
```
rhuss: btw, fixed the issue that genertors having the same order overwrite each other, so that some get lost. The problem was that they returned 0 in `compareTo()` which is used in `TreeSet` also for equals(). See #722 for details.

generator/api/src/main/java/io/fabric8/maven/generator/api/support/AbstractPortsExtractor.java
iocanel: a `ports` property doesn't match the regex, neither a property like `ports.http` (with or without flattening).

rhuss:
Thanks ;-)

rhuss:
[merge]

Issue #665, opened 2016-11-18
Multiport services are ignored


iocanel:
In the case that a generator detects multiple ports, the `EXPOSE` is correctly added to the image, but the generated Service resources, only contain the first service.

It seems that its done on purpose, according to:

https://github.com/fabric8io/fabric8-maven-plugin/blob/e9bc60ab870e8137a6697883b22e9cc536b1d8ae/core/src/main/java/io/fabric8/maven/core/handler/ServiceHandler.java#L77-L77

I don't see a reason for ignoring ports, that we've automatically detected and it pretty much defeats the purpose of auto-detection.

iocanel:
@jstrachan: what's the purpose of the `break`?

jstrachan:
Kubernetes services with more than one port break Routes on OpenShift meaning you can no longer access any ports from the service reliably from outside the cluster when using OpenShift
https://github.com/openshift/origin/issues/2092

its not clear if/when this will ever be fixed TBH so its better to not use multi-port services if you want to use OpenShift

jimmidyson:
Not quite true `route.spec.port` can be used to specify the service port to use for multiport services.

iocanel:
Awesome !!!

iocanel:
@jstrachan: now that we got routes coverred can we revert this?

jstrachan:
we need to check that using Routes with fabric8-maven-plugin with multi-port services actually works though ;) (e.g. check it works with exposecontroller + fabric8-maven-plugin + jenkins pipeline)

jstrachan:
@iocanel why not try create a multi-port service then check it works with exposecontroller + openshift routes with CI / CD pipelines? if all thats good we can revert this

rhuss:
I wonder whether instead of a single multiport service, why don't we create multiples services with a single port ? It seems that e.g. our elasticsearch image does this for port 9300 and 9200 (i.e. creating two services)

Could this be an alternative ?

iocanel:
@rhuss: the main issue with this approach is that we need a different name for each. Isn't that a problem (given that internally we just handle them as a list of numbers)?

rhuss:
We could try to detect a 'main service' which gets the default name, alternative services then this name + the service type (thanks to @jimmidyson we have https://github.com/fabric8io/ianaservicehelper which can lookup the name of an IANA service based on its port), e.g. 'myapp-mysql' (for port 3306 or so). 

[...]

PR #666, opened 2016-11-18
#665: Remove break from service port iteration

iocanel:


rhuss:
retest please

rhuss:
test this please

jimmidyson:
Think this will also add the Prometheus & jolokia ports to the service for Java exec generators won't it? We don't want that. I think single port is good enough for now with yaml files providing extra ports as required.

davsclaus:
Good observations @jimmidyson - I think a single service is the norm over http. 

So if people want additional I am fine with them having to explicit configure this somewhere.

rhuss:
But which one should be taken when there are multiple ports exposed ? Arbitrarly the first one exposed in the Docker image ? If so, that must be well documented.

jimmidyson:
Choices:

* Require port to always be explicitly specified (not a great UX IMO)
* Only add the service port if there is a single port exposed, otherwise require explicit configuration (all of our Javaimages expose multiple ports IIRC so won't apply to much of our stuff & we are the largest consumer of course).
* Add first port discovered & document that behaviour. I think this works for about ~90% of use cases but could of course be wrong about that.

rhuss:
For me its fine to take the first one. The service enricher could even let this behaviour be configurable, e.g. adding a 'mode' with values 'first' or 'all' (just as example. could be also a 'ports' which takes a numeric list, or these two literal string values)

jimmidyson:
Either way, I think this PR should be closed & more discussion in #665 to come up with the right way to handle it.

rhuss:
Added an alternate suggestion to #665 , so going to close this PR for now.

PR #680, opened 2016-11-25
fabric8:debug - Enable JAVA_DEBUG_ENABLED on all containers of a pod (since the application needs to be the first one)

rhuss:


rhuss:
[merge]

PR #681, opened 2016-11-26
Added new profile "explicit" which does not add new objects

rhuss:


rhuss:
[merge]

plugin/src/main/resources/META-INF/fabric8/profiles.yml
```diff
@@ -4,6 +4,13 @@
 - name: raw
   enricher:
     includes: []
+# No defaults added, useful for explicit configuratiomn
```
rhuss: thx :)

PR #692, opened 2016-12-02
Remove default service type

### LINKED ISSUE #693: Change default service type from LoadBalancer to ClusterIP
### Description

Using `LoadBalancer` as standard type is not suitable in situation when there is no load balancer available. I guess the only save default is to use `ClusterIP` or nothing (which actually maps to `ClusterIP`, too). This lead also to issue within OpenShift (https://issues.jboss.org/browse/OSFUSE-469)

If an fabric8 environment we still can use the `exposecontroller` to change this type on the fly.

If there are not objection, I would remove the default (so that no type is added).

jimmidyson:
This is of course a breaking change - I think it's the right thing to do though.

Fixes #693

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
@@ -68,8 +68,8 @@ public class DefaultServiceEnricher extends BaseEnricher {
         // running
         expose {{ d = "false"; }},
 
-        // Type of the service (LoadBalancer, NodePort, ...)
-        type {{ d = "LoadBalancer"; }};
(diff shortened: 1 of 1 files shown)


davsclaus:
lgtm

I just deployed swarm-camel and it was using LoadBalancer and in my local minikube there is no such a thing and it was in pending state.

davsclaus:
Maybe we can release the fmp as 3.3.0 then with this change in mind?

jimmidyson:
Agreed. Can we merge this now?

davsclaus:
@rhuss are you okay with this change and maybe release the next fmp as 3.3.0?

And @rawlingsj @jstrachan any thoughts?

rhuss:
I'm fine with this

jimmidyson:
Let's merge this -can always revert if needed.

PR #694, opened 2016-12-02
Ports should be ordered in default service generator

jimmidyson:
As only the first port is added to default generated service, order of ports is import so have to use a list rather than a set.


rhuss:
This is related to #666 which is discussed in #665 . We should use all ports exported for services if they are not marked as internal.

I dare a bit to applying #666 right now since it potentially breaks some stuff. In fact, I'd like to add the breaking things at the end of next week, when I have time to care about thinks that break. Not that people get angry about things breaking up.

But this PR looks perfeclty safe.

[merge]

Issue #376, opened 2016-09-19
Add WildFly-Swarm readiness check


davsclaus:
WF Swarm has healthcheck if you add the monitor dependency.

However it has a little flaw that when calling /health it returns 503 if there is no custom health endpoints, as it does not provide a default OOTB. So you can use /node to check if the node is running.

Hopefully they make /health better. I logged a ticket:
https://issues.jboss.org/browse/SWARM-693

But we could make f-m-p detect the monitor dependency in the pom.xml and then add /node as readiness/liveness probe - like we do for Spring Boot.

davsclaus:
@christian-posta did a webinar recently and got readiness check working on wf-swarm. So we could likely add the fractions to our swarm quickstarts. And then make the fmp plugin detect this fraction and then automatic add a readiness/liveness check on that url

davsclaus:
The new 2016.12 release now has health check support on the same endpoint `health` as SB does. We can then use that if `swarm-monitoring` is on the classpath.

rhuss:
Implemented already for Wildfly and Thorntail V2.

Issue #678, opened 2016-11-25
Refactor Enricher Workflow to allow direct generation of OpenShift objects


rhuss:
### Description

The current way to support both, Kubernetes and OpenShift is, that we first create all K8s objects for `kubernetes.yml` and the convert them all in one step to corresponding OpenShift object for `openshift.yml`.

Although it worked quite well until now, this approach has now reached its limitations so that a pure 1:1 mapping is not easily possible and lead to ugly constructs and code coupling. (e.g. the creation of OpenShift specific objects for the app-catalog by the dependency enricher, which are then directly extracted by the ResourceMojo. Its a good example how leaky our abstraction became).

The suggested solution is to use a two-pass model, to create the OpenShift and Kubernetes individually by running the Enricher chain twice, with an argument `PlatformMode` for specifying which objects are to be created. That adds gives more flexibility and avoids the brute-force all in once conversion at the end. 

Also, the enricher API can be simplified to two methods:

```java
// Create default resources
void create(KubernetesListBuilder builder, PlatformMode mode);

// Enrich with extra information
void enrich(KubernetesListBuilder builder, PlatformMode mode);
```

The enrichment process then would work like:

```
for mode in "kubernetes", "openshift"
  builder = new ...
  for enricher in enrichers
    enricher.create(builder, mode);
  for enricher in enrichers
    enricher.enrich(builder, mode);
  writeToFile(builder, mode);
```

Also, for special use case like creating the app-catalog it might make sense to call the chain separately instead of relying that fabric8:resource runs before fabric8:app-catalog and doing checks in the ResourceMojo whether it runs together with other Mojos to create different descriptors (with the same file name !).

nicolaferraro:
Just a note. This can create backward compatibility issues, because people usually put some env settings (eg. cpu and mem limits) in the `deployment.yml` file, both if they are deploying on Openshift or Kubernetes. They will need to use a different descriptor for Openshift (`dc.yml`).

rhuss:
@nicolaferraro agree that we should still some 'internal' conversions for resource fragments which we are used for both variations and which we already convert on the fly into each other

* `Deployment` and `DeploymentConfig`
* `Namespace` and `Project`
* `ReplicatSet` and `ReplicationController` (for OpenShift)

nicolaferraro:
Yeah, this may be a (active-by-default) option.

rohanKanojia:
Closed via #1521

PR #696, opened 2016-12-03
#376: Add WildFly-Swarm health check enricher. Not yet working becaus…

davsclaus:
…e fmp doesnt call it.

### DIFF
--- a/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
+++ b/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
@@ -20,13 +20,13 @@ public class WildFlySwarmHealthCheckEnricher extends AbstractHealthCheckEnricher
     // Available configuration keys
     private enum Config implements Configs.Key {
 
-        defaultScheme {{
+        scheme {{
             d = "HTTP";
         }},
-        defaultPort {{
+        port {{
             d = "8080";
         }},
-        defaultPath {{
+        path {{
             d = "/health";
         }};
 
@@ -65,15 +65,15 @@ public class WildFlySwarmHealthCheckEnricher extends AbstractHealthCheckEnricher
     }
 
     protected String getScheme() {
-        return Configs.asString(getConfig(Config.defaultScheme));
+        return Configs.asString(getConfig(Config.scheme));
     }
 
     protected int getPort() {
-        return Configs.asInt(getConfig(Config.defaultPort));
+        return Configs.asInt(getConfig(Config.port));
     }
 
     protected String getPath() {
-        return Configs.asString(getConfig(Config.defaultPath));
+        return Configs.asString(getConfig(Config.path));
     }
 
 
--- a/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
+++ b/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
@@ -1,0 +1,48 @@
+package io.fabric8.maven.enricher.fabric8;
+
+import io.fabric8.kubernetes.api.model.Probe;
+import io.fabric8.kubernetes.api.model.ProbeBuilder;
+import io.fabric8.maven.enricher.api.AbstractHealthCheckEnricher;
+import io.fabric8.maven.enricher.api.EnricherContext;
+
+import static io.fabric8.maven.core.util.MavenUtil.hasDependency;
+
+/**
+ * Enriches wildfly-swarm containers with health checks if the monitoring fraction is present.
+ */
+public class WildFlySwarmHealthCheckEnricher extends AbstractHealthCheckEnricher {
+
+    private static final int DEFAULT_HEALTH_PORT = 8080;
+    private static final String SCHEME_HTTP = "HTTP";
+
+    public WildFlySwarmHealthCheckEnricher(EnricherContext buildContext) {
+        super(buildContext, "wildfly-swarm-health-check");
+    }
+
+    @Override
+    protected Probe getReadinessProbe() {
+        Probe probe = discoverWildFlySwarmHealthCheck(10);
+        return probe;
+    }
+
+    @Override
+    protected Probe getLivenessProbe() {
+        Probe probe = discoverWildFlySwarmHealthCheck(180);
+        return probe;
+    }
+
+    private Probe discoverWildFlySwarmHealthCheck(int initialDelay) {
+        if (hasDependency(this.getProject(), "org.wildfly.swarm", "monitor")) {
+            // TODO: wildfly-swarm does not yet have a standard configuration file
+            Integer port = DEFAULT_HEALTH_PORT;
+            String scheme = SCHEME_HTTP;
+
+            // lets default to adding a spring boot actuator health check
+            return new ProbeBuilder().
+                    withNewHttpGet().withNewPort(port).withPath("/health").withScheme(scheme).endHttpGet().
+                    withInitialDelaySeconds(initialDelay).build();
+        }
+        return null;
+    }
+
+}
--- a/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
+++ b/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
@@ -2,6 +2,7 @@ package io.fabric8.maven.enricher.fabric8;
 
 import io.fabric8.kubernetes.api.model.Probe;
 import io.fabric8.kubernetes.api.model.ProbeBuilder;
+import io.fabric8.maven.core.util.Configs;
 import io.fabric8.maven.enricher.api.AbstractHealthCheckEnricher;
 import io.fabric8.maven.enricher.api.EnricherContext;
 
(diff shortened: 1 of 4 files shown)


davsclaus:
Hey @rhuss I started on a health check enricher for wildfly-swarm. 

This comes OOTB in swarm if you have a dependency on that monitor dependency.

But our quickstart
https://github.com/fabric8-quickstarts/swarm-camel

Does not make fmp generate the livness and readiness probes in the generate k8s resource files.

I debugged the code a bit and in 
https://github.com/fabric8io/fabric8-maven-plugin/blob/26126525719c363c5e608d8d15dd4dad58795e49/enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java#L22-L22

Then item is never an instance of Deployment and the code below is not triggered that adds those probes. 

I wonder if you could take a look?

rhuss:
Intersting, because afaik there _should_ be a `Deployment` object egernated automatically by the `DefaultControllerEnricher`. Of course the Deployment is only available if the DefaultControllerEnricher is running before the HealtCheckEnricher. 

I will check that.

rhuss:
In fact, the enricher for the health-checks kicks in to early (as it parts of the `f8` enricher in contrast to the `fmp` standard enrichers).

The solution is to add a high order value to the health check enrichers.

However, the _real_ solution is to separate the generation of default resources (like the `Deployment` in this case) with the enrichment process (used by the health-checl). See #678 for more details on this.

I will submit a PR for the order issue.

rhuss:
@davsclaus could you try to add a high order number to the healtcheck enrichers in https://github.com/rhuss/fabric8-maven-plugin/blob/e9bc60ab870e8137a6697883b22e9cc536b1d8ae/enricher/fabric8/src/main/resources/META-INF/fabric8/enricher-default#L9-L8 ? (like for the ExposeController) and check whether this works for you ?

I'm currently working on a more general solution to explicit define the order instead on relying of an order when merging enrichers from different files.

davsclaus:
Thanks @rhuss that worked with ,1000, eg I see now
```
[INFO] --- fabric8-maven-plugin:3.2-SNAPSHOT:resource (default) @ swarm-camel ---
[INFO] F8: Running in Kubernetes mode
[INFO] F8: Running generator wildfly-swarm
2016-12-05 11:23:47 INFO  Version:30 - HV000001: Hibernate Validator 5.2.4.Final
[INFO] F8: fmp-controller: Adding a default Deployment
[INFO] F8: fmp-service: Adding a default Service with ports [8080]
[INFO] F8: wildfly-swarm-health-check: Adding readiness probe on port 8080, path='/health', scheme='HTTP', with initial delay 10 seconds
[INFO] F8: wildfly-swarm-health-check: Adding liveness probe on port 8080, path='/health', scheme='HTTP', with initial delay 180 seconds
```

And yay the readiness only signal its ready to k8s when swarm is really started after 31 seconds
```
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   0/1       Running   0          5s
davsclaus:/Users/davsclaus/$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   0/1       Running   0          8s
davsclaus:/Users/davsclaus/$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   0/1       Running   0          8s
davsclaus:/Users/davsclaus/$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   0/1       Running   0          13s
davsclaus:/Users/davsclaus/$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   0/1       Running   0          14s
davsclaus:/Users/davsclaus/$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   0/1       Running   0          21s
davsclaus:/Users/davsclaus/$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
swarm-camel-2856146500-ng5yy   1/1       Running   0          31s
```

Before the pod was started asap and it would of course fail to service the traffic.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
```diff
@@ -0,0 +1,48 @@
...
+    protected Probe getLivenessProbe() {
+        Probe probe = discoverWildFlySwarmHealthCheck(180);
+        return probe;
+    }
+
+    private Probe discoverWildFlySwarmHealthCheck(int initialDelay) {
+        if (hasDependency(this.getProject(), "org.wildfly.swarm", "monitor")) {
+            // TODO: wildfly-swarm does not yet have a standard configuration file
+            Integer port = DEFAULT_HEALTH_PORT;
+            String scheme = SCHEME_HTTP;
```
jimmidyson: Can we make the scheme configurable via property please?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
```diff
@@ -0,0 +1,48 @@
...
+import io.fabric8.maven.enricher.api.EnricherContext;
+
+import static io.fabric8.maven.core.util.MavenUtil.hasDependency;
+
+/**
+ * Enriches wildfly-swarm containers with health checks if the monitoring fraction is present.
+ */
+public class WildFlySwarmHealthCheckEnricher extends AbstractHealthCheckEnricher {
+
+    private static final int DEFAULT_HEALTH_PORT = 8080;
```
jimmidyson: Is there no way we can discover this port? If not, then a config property would be great.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    protected Probe getReadinessProbe() {
+        Probe probe = discoverWildFlySwarmHealthCheck(10);
+        return probe;
+    }
+
+    @Override
+    protected Probe getLivenessProbe() {
+        Probe probe = discoverWildFlySwarmHealthCheck(180);
```
jimmidyson: 180 seconds is quite a long time - do we expect that to be a reasonable startup expectation for swarm containers?

davsclaus: Its using the same values as SB

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
davsclaus: property from where? In the f-m-p? And if so, how would you do that @rhuss ?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
jimmidyson: Yeah OK let's discuss that in a separate issue so we have common settings.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
jimmidyson: You can define configurable properties as in https://github.com/fabric8io/fabric8-maven-plugin/blob/master/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DependencyEnricher.java#L62-L76 & then retrieve them as in https://github.com/fabric8io/fabric8-maven-plugin/blob/master/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DependencyEnricher.java#L172

davsclaus: Thanks @jimmidyson I have just pushed more commits with configurable properties.

davsclaus:
Any thoughts of this?

With this PR then our swarm quickstarts at fabric8 now works and deploy correctly and wont be ready until swarm is ready.

So I would like to see it merge.

Then we can open another ticket about trying to discover wildfly-swarm configuration files and grab port from there, which essentially also the service enricher would need to, if it needs to detect a different port than 8080.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WildFlySwarmHealthCheckEnricher.java
```diff
@@ -0,0 +1,80 @@
...
+public class WildFlySwarmHealthCheckEnricher extends AbstractHealthCheckEnricher {
+
+    public WildFlySwarmHealthCheckEnricher(EnricherContext buildContext) {
+        super(buildContext, "wildfly-swarm-health-check");
+    }
+
+    // Available configuration keys
+    private enum Config implements Configs.Key {
+
+        defaultScheme {{
```
jimmidyson: As these are actually the names of the properties used, they shouldn't have the `default` prefix AFAIK. @rhuss - halp?

davsclaus: Yeah it makes sense. I pushed a commit without the default prefix.

davsclaus:
[merge]

davsclaus:
Logged a ticket about discovering port number from configuration files
https://github.com/fabric8io/fabric8-maven-plugin/issues/718

PR #698, opened 2016-12-03
Add auto TLS enricher

### LINKED ISSUE #697: Add enricher for auto TLS on openshift with conversion of keystore
### Description

Openshift provides a mechanism to auto issue certs as secrets for services with appropriate annotations. This can be mounted in pods & used to secure access to endpoints. The certs & keys are provided in PEM format which Java cannot use directly, so we can automatically enrich with an init container to do the conversion so that a keystore is available on startup for the Java process to use.

So the enricher will have to add annotations to the service, add secret volume, add memory-backed empty dir volume to store keystore, & add init container. This all works great manually, but that's quite a bit of stuff to do & this could be really useful for a lot of applications, hence the enricher.

The enricher is only applicable to openshift origin  >= 1.3 & OSE >= 3.3. This should be an opt-in enricher rather than a default (not every service wants to be secured in this way).

I've already made a start on this - enjoying working with enricher api btw!

jimmidyson:
Fixes #697 

Advice needed on how to best apply this so it only applies to OpenShift manifests.

Also, how to add this as an optional enricher that can be added to `<enricher><includes><include>...` in a project? Easy to add as a default enricher, but this should be opt-in, rather than a default.


jimmidyson:
And yes I need to add docs.

jimmidyson:
So this has grown a bit because I had to refactor the init container stuff, had to update `VolumePermissionEnricher` so added a test for that too.

This needs to wait to merge until @rhuss has sorted out the default/none profiles as discussed on IRC.

rhuss:
@jimmidyson hope that I have something running this evening, but we need to test this thoroughly since the algorithm changed quite a bit. Let you know when I'm done.

rhuss:
I have something running , but unfortunately not yet for prime time. Still missing tests and docs (and the definition of the 'default' profile). Work is happening on #709

cmoulliard:
What this TLE enricher is supposed to do (generate a self signed certificate + keystore) ? Is there a document describing the use case implemented ?

jimmidyson:
See the referenced issue #697 plus the OpenShift docs at https://docs.openshift.org/latest/dev_guide/secrets.html#service-serving-certificate-secrets

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -114,4 +127,32 @@ protected boolean isOpenShiftMode() {
...
+
+    protected void ensureMetadata(PodTemplateSpecBuilder obj) {
+        ObjectMeta metadata = obj.getMetadata();
+        if (metadata == null) {
+            obj.withNewMetadata().endMetadata();
+        }
+    }
+
+    protected void addInitContainer(ObjectMetaBuilder metadata, JSONObject initContainer) {
+        String initContainerAnnotation = metadata.getAnnotations().get(INIT_CONTAINER_ANNOTATION);
```
rhuss: annotations are always non-null, right ? 

I always get confused with when using builder and when using object directly. I.e. I would have expected here a 'editAnnotations()' or so. But I remember darkly that for lists one always use the 'direct' objects.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -114,4 +127,32 @@ protected boolean isOpenShiftMode() {
...
+
+        JSONArray initContainers = Strings.isNullOrBlank(initContainerAnnotation) ? new JSONArray()
+                : new JSONArray(initContainerAnnotation);
+        initContainers.put(initContainer);
+
+        metadata.addToAnnotations(INIT_CONTAINER_ANNOTATION, initContainers.toString());
+    }
+
+    protected void addInitContainer(PodTemplateSpecBuilder builder, JSONObject initContainer) {
```
rhuss: What the difference between both 'addInitContainer()' methods ?

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -114,4 +127,32 @@ protected boolean isOpenShiftMode() {
...
+        ensureMetadata(builder);
+        String initContainerAnnotation = builder.getMetadata().getAnnotations().get(INIT_CONTAINER_ANNOTATION);
+
+        JSONArray initContainers = Strings.isNullOrBlank(initContainerAnnotation) ? new JSONArray()
+                : new JSONArray(initContainerAnnotation);
+        initContainers.put(initContainer);
+
+        builder.editMetadata().addToAnnotations(INIT_CONTAINER_ANNOTATION, initContainers.toString()).endMetadata();
```
rhuss: Ah, here it is :) That's what I expected how to deal with annotations on builders.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
```diff
@@ -0,0 +1,190 @@
...
+
+        keystoreCertAlias          {{ d = "server"; }};
+
+        public String def() { return d; } protected String d;
+    }
+
+    public AutoTLSEnricher(EnricherContext buildContext) {
+        super(buildContext, ENRICHER_NAME);
+
+        this.secretName = getConfig(Config.tlsSecretName);
```
rhuss: you could use directly

```java
this.secretName = getConfig(Config.tlsSecretName, getProject().getArtifactId() + "-tls");
```

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
```diff
@@ -0,0 +1,190 @@
...
+        if (Strings.isNullOrBlank(secretName)) {
+            this.secretName = getProject().getArtifactId() + "-tls";
+        }
+    }
+
+    @Override
+    public Map<String, String> getAnnotations(Kind kind) {
+        if (!isOpenShiftMode()) {
+            return super.getAnnotations(kind);
```
rhuss: You should probably return `null` directly here since your know that your enricher doesn't want to add any annotations in this case.

Actually I would collapse this method in sth like

```java
if (!isOpenShiftMode() || !kind.isService()) {
   return null;
}
return Collections.singletonMap(AUTOTLS_ANNOTATION_KEY, this.secretName);
```

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
```diff
@@ -0,0 +1,190 @@
...
+            return Collections.singletonMap(AUTOTLS_ANNOTATION_KEY, this.secretName);
+        }
+
+        return super.getAnnotations(kind);
+    }
+
+    @Override
+    public void addMissingResources(KubernetesListBuilder builder) {
+        if (!isOpenShiftMode()) {
+            super.addMissingResources(builder);
```
rhuss: wouldn't call super here as it is not necessary. 

Enrichers are not chained, they are called individually and are hence independent. The `BaseEnricher` was introduced merely to allow an implementer to ommit implementing certain interface methods if she is not interested in them. Actually its more like the former Swing Adapters used for event handling.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
```diff
@@ -0,0 +1,190 @@
...
+                }
+                return false;
+            }
+        });
+    }
+
+    @Override
+    public void adapt(KubernetesListBuilder builder) {
+        if (!isOpenShiftMode()) {
+            super.adapt(builder);
```
rhuss: dito

enricher/standard/src/main/test/java/io/fabric8/maven/enricher/standard/VolumePermissionEnricherTest.java
```diff
@@ -0,0 +1,126 @@
...
+            final ProcessorConfig config = new ProcessorConfig(null, null,
+                    Collections.singletonMap(VolumePermissionEnricher.ENRICHER_NAME, new TreeMap(Collections
+                            .singletonMap(VolumePermissionEnricher.Config.permission.name(), tc.permission))));
+
+            // Setup mock behaviour
+            new Expectations() {{ context.getConfig(); result = config; }};
+
+            VolumePermissionEnricher enricher = new VolumePermissionEnricher(context);
+
+            PodTemplateBuilder ptb = new PodTemplateBuilder().withNewMetadata().and()
```
rhuss: just out of curiosity: what does 'and()' do ?

jimmidyson:
This is now working thanks to @rhuss's brilliant profile work.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -114,4 +127,32 @@ protected boolean isOpenShiftMode() {
...
     }
+
+    protected void ensureMetadata(PodTemplateSpecBuilder obj) {
+        ObjectMeta metadata = obj.getMetadata();
+        if (metadata == null) {
+            obj.withNewMetadata().endMetadata();
+        }
+    }
+
+    protected void addInitContainer(ObjectMetaBuilder metadata, JSONObject initContainer) {
```
rhuss: Seems like that this method is never used ?

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
jimmidyson: This is retrieving current annotation, hence `getAnnotations`, so that the two can be merged. Should always be non-null.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
jimmidyson: Removed the unused one :)

enricher/standard/src/main/test/java/io/fabric8/maven/enricher/standard/VolumePermissionEnricherTest.java
jimmidyson: Just like `end...` but more fluent (reads better & means you don't have to remember what kind you're editing).

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
jimmidyson: Removed.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
jimmidyson: :)

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
jimmidyson: Done.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
jimmidyson: Done.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
jimmidyson: Done.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
jimmidyson: Done.

jimmidyson:
Do you want me to squash?

BTW the changes to VolumePermissionEnricher were to reuse the init container editing between two enrichers that contribute init containers. I think as a side effect it may actually have fixed #772  btw - checking now.

rhuss:
Yeah, would make sense to squash, as I often have to juggle with the v3.1.80-redhat branch, too. hence would prefer one commit per feature.

jimmidyson:
Hmm can't recreate #772 anyway... odd.

jimmidyson:
Squashed commits.

rhuss:
thanks ;-) Looks good. Now only the documentation is left (a simple table with options is enough) and the entry in `enricher-default`, then we should merge.

jimmidyson:
Sorry - I've added some docs for the options now.

jimmidyson:
@rawlingsj I refactore the VolumePermissionEnricher as part of this PR to share handling of init containers in BaseEnricher. This also included a change to the visitor approach for updates to resources. I've just rebased including your recent PR to fix VolumePermissionEnricher - would you mind casting your eye over it to see I haven't broken anything?

I've also added a test for it so I can add the case you've just fixed to that test if you'd like - can you just explain what the problem was so I can update the test case appropriately?

jimmidyson:
@rawlingsj Don't worry - figured out the test stuff. Works a treat :)

rawlingsj:
@jimmidyson ah cool - yeah LGTM, the main bit for me was only adding the initcontainer annotation for PVC volumes which you have, so all good!

rhuss:
Thanks ! LGTM, too

[merge]

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
```diff
@@ -0,0 +1,180 @@
...
+    public Map<String, String> getAnnotations(Kind kind) {
+        if (!isOpenShiftMode() || !kind.isService()) {
+            return null;
+        }
+        return Collections.singletonMap(AUTOTLS_ANNOTATION_KEY, this.secretName);
+    }
+
+    @Override
+    public void addMissingResources(KubernetesListBuilder builder) {
+        if (!isOpenShiftMode()) {
```
rhuss: While trying to fix other stuff, stumbled over this base `isOpenShiftMode()` method which has been introduced lately.

@jimmidyson @jstrachan @rawlingsj : Just a heads up (and also it was used for adding docker env vars to pod envs when using s2i binary build): It won't work as you might expect, since we always create kubernetes.yml and openshift.yml, regardless of the platform mode (auto detected or with -Dfabric8.mode). When the platform mode is used in an enricher it will create different kubernetes.yml and openshift.yml depending on the mode, but it still creates both files. And you can perfectly apply an openshift.yml to OpenShift which has been created with a `mvn -Dfabric8.mode=kubernetes fabric8:resource` (or when its running in auto detect mode and not being connected to an active OpenShift context). 

This is also dangerous when you refer to these files included in dependencies as it now matters with which mode these dependencies has been build.

The solution is of course again described in #678 so that actually kubernetes and openshift objects are not converted into each other in an after thought but by calling each enricher twice, once for kubernetes and once for openshift, independent on the fabric8.mode in use (which however still is needed for deciding how to do the build).

Until then this might work or not and for sake of simplicity I even will take this call over to fix the VolumePermissionEnricher to not add init containers for OpenShift.

PR #699, opened 2016-12-05
Allow multi Java Options + fix all license headers

rhuss:


generator/api/src/test/java/io/fabric8/maven/generator/api/BaseGeneratorTest.java
```diff
@@ -0,0 +1,24 @@
+package io.fabric8.maven.generator.api;
+/*
+ *
+ * Copyright 2016 Roland Huss
```
davsclaus: We should use the RH license header file. Also its more common to have the lic header before package

rhuss: ah, excellent point :)

Will update my IDE prefs accordingly.

rhuss:
[merge]

rhuss:
[merge]

PR #703, opened 2016-12-05
Update samples to 3.2-SNAPSHOT

apupier:
fabric8-maven-sample-eng left behind as there are other issues with
spring-boot-generator-jersey


rhuss:
thanks !

[merge]

Issue #700, opened 2016-12-05
Provide default configuration for m2e


apupier:
_Please fill out the following information to help us in analyzing the issue, but feel free to skip it if you don't have the information at hand or if it does not apply. Please remove everything which does not apply to the issue. Including this info paragraph ;-)_

### Description

When using fabric8-maven-plugin in Eclipse, currently there is an error because there is no m2e (maven 2 Eclipse) connector available neither a default mapping.
To avoid that a default mapping can be provided in the plugin itself (see https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html )
As the phase/goal mapping is not clear yet, I propose to start by just ignoring this plugin execution in Eclipse build.

### Info

* f-m-p version : 2.2
* Maven version (`mvn -v`) :
```

```
* Kubernetes / OpenShift setup and version :
* If it's a bug, how to reproduce :
* If it's a feature request, what is your use case :

Import a Quickstart in Eclipse IDE

* Sample project : *[GitHub Clone URL]*

PR #704, opened 2016-12-05
#700: Default m2e configuration to avoid errors in Eclipse

apupier:
Configure to ignore fabric8-maven-plugin goals in Eclipse to avoid
errors when importing a project using fabric8-maven-plugin

### DIFF
--- a/plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml
+++ b/plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE xml>
+<lifecycleMappingMetadata>
+  <pluginExecutions>
+    <pluginExecution>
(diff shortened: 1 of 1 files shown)


rhuss:
Don't know m2e (and tbh I'm not an Eclipse user either) and how it works and why its get confused by the current setup but if this makes the life of Eclipse developers easier I'm all good with it (as I suppose it does not have any side effects in another environment than eclipse).

apupier:
there is no effect to other environments.
m2e requires lifecycle information for every maven plugins in order to know how to bind them in the Eclipse build lifecycle. Errors are reported if there is an unknown mapping.
The implementation of this PR is just saying: "this maven plugin is not mapped to any lifecycle, just ignore it".
It means that the user will have to launch a Maven build explicitly, no automatic build will done by Eclipse.
I hope that I will be able to provide a better mapping in coming weeks/months to have some better integration with the Eclipse OpenShift Tooling provided by Red Hat.

rhuss:
Make sense, thanks !

[merge]

PR #709, opened 2016-12-06
Introduction of a 'default' profile and allow profile merging

rhuss:
Previously all enrichers and generators defined in the their service definiton files (generators-default and enrichers-default) where taken. If multiple such files were found, they were merged which lead to unpredictable results (when no explicit order is given). Also is was not possible to define _optional_ enrichers since all defined enrichers were active by default.

This commit introduces a "default" profile which defines the set of enrichers and generatos fine granularily. The definition order of these objects are not relevant anymore. The default profile is active if not an explicit profile is given. The default profile can be tuned by providing an own profile with name "default".

The way how this can be used is documented in [_profiles.adoc](https://github.com/fabric8io/fabric8-maven-plugin/blob/d8d1d28918f82d685af05e3c62789f6d9f846ed9/doc/src/main/asciidoc/inc/_profiles.adoc)

Unit tests has been added to secure this rather big change.

**Note for reviewers**: Its a big change and I tried my best to secure these with unit tests. I'm especially keen on a review of the [documentation](https://github.com/fabric8io/fabric8-maven-plugin/blob/d8d1d28918f82d685af05e3c62789f6d9f846ed9/doc/src/main/asciidoc/inc/_profiles.adoc) whether its understandable and whether it makes sense at all.


core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
```diff
@@ -121,24 +83,117 @@ public boolean use(String name) {
...
+                List<String> includes = config.includes;
+                if (includes != null) {
+                    ret.addAll(includes);
+                }
+            }
+        }
+        return removeDups(ret);
+    }
+
+    // Remove duplicates such that the later element remains
```
jimmidyson: Why only `later element remains`? Wouldn't insertion order be better so could just use a `LinkedHashSet`instead of `ArrayList`?

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
```diff
@@ -34,105 +35,148 @@
...
     /**
-     * Find an enricher or generator config, possibly via a profile
+     * Find an enricher or generator config, possibly via a profile and merge it with a given configuration.
      *
      * @param extractor how to extract the config from a profile when found
      * @param profile the profile name (can be null, then no profile is used)
      * @param resourceDir resource directory where to lookup the profile (in addition to a classpath lookup)
-     * @return the configuration found or {@link ProcessorConfig#INCLUDE_ALL} to include all enricher / generators
-     * if no profile was found
+     * @return the merged configuration which can be emoty if no profile is given
```
jimmidyson: typo: `emoty` -> `empty`

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
```diff
@@ -34,105 +35,148 @@
...
+     * Read all profiles found in the classpath.
+     *
+     * @param name name of the profile to lookup
+     * @param ext to use (e.g. 'default' for checking 'profile-default.yml'. Can also be null or empty.
+     * @return all profiles with this name stored in files with this extension
+     *
+     * @throws IOException if reading of a profile fails
+     */
+    public static Profile[] readAllFromClasspath(String name, String ext) throws IOException {
+        List<Profile > ret = new ArrayList<>();
```
jimmidyson: How about using a `TreeSet` with the `Profile.OrderComparator()` specified?

core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
```diff
@@ -121,24 +83,117 @@ public boolean use(String name) {
...
+        for (ProcessorConfig config : configs) {
+            if (config != null) {
+                Set<String> excludes = config.excludes;
+                if (excludes != null) {
+                    ret.addAll(excludes);
+                }
             }
-            ret.add(named);
         }
         return ret;
```
jimmidyson: Do we need to `removeDups` here too?

core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
rhuss: AFAIK does LinkedHashSet not change the insertion order if an element is added a second time:

>  Note that insertion order is <i>not</i> affected if an element is <i>re-inserted</i> into the set.  (An element <tt>e</tt> is reinserted into a set <tt>s</tt> if <tt>s.add(e)</tt> is invoked when <tt>s.contains(e)</tt> would return <tt>true</tt> immediately prior to the invocation.)

Otherwise it would be better to use such a collection instead of redoing it. I could do the inertion in reversed order but end it the end its not really simpler (reverse, add, reverse)

rhuss: That quote is from LinkedHashSet's javadoc.

core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
rhuss: No, because its a plain Set (so simply overwriting an existing element). The set is only used to determine whether an enricher should be excluded, no order is needed here.

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
rhuss: fixed

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
rhuss: Yeah, makes more sense. Save on sort step. I'll fix that.

rhuss:
@jimmidyson thanks for the fast review. The coverage could be higher but I left out those parts which I want to refactor / remove anyway.

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
rhuss: Ah, just saw that I made my old error again: If a comparator returns 0, then both elements must be also equal. If not, they override each other in the TreeSet. In our case the order can be the same for both processors (0 by default), so I need an extra check whether its the same.

core/src/main/java/io/fabric8/maven/core/config/Profile.java
```diff
@@ -94,13 +102,30 @@ public ProcessorConfig getGeneratorConfig() {
...
-    public static class OrderComparator implements Comparator<Profile> {
+    public int getOrder() {
+        return order;
+    }
 
-        @Override
-        public int compare(Profile o1, Profile o2) {
-            return o1.order - o2.order;
+    @Override
+    public int compareTo(Profile o) {
```
rhuss: @jimmidyson could you give this method a quick look ? It tries hard to determine a reproducible order, if the profiles are not equal (equals in the sense of ==). Don't know whether the creationTime thing is needed, since its arbitrary then anyway (the "name" is always the same when comparing since only same named profiles are compared).

Maybe going with the hashcode() is simpler.

core/src/main/java/io/fabric8/maven/core/config/Profile.java
```diff
@@ -56,4 +102,30 @@ public ProcessorConfig getGeneratorConfig() {
...
+            return orderDiff;
+        }
+        // Equals is only the very same object. This is ok, since at the end
+        // profiles are merge to a single profile
+        if (this.equals(o)) {
+            return 0;
+        } else {
+            // Try to determine the order based on creation time. The rule is, a newer profile
+            // is supposed to have a higher order.
+            long timeDiff = creationTime - o.creationTime;
```
jimmidyson: `creationTime` could quite often be equal here I assume so not sure it adds any value.

Can we work out an order from the various sources of profile?

rhuss: Not really since these profiles which are 'sorted' come from a `ClassLoader.getResources()` which is also non deterministic. I don't mind to have the order of Profiles unspecified when the order number is the same. However it should be stable. 

Actually the sorting seems to be still a bit messy since it happens separately in different list while collecting profiles and not at the end in an overall list. So, I need to work on this anyway still.

tbh, don't know whether this is really so important. however I think a consistent behaviour is important.

So I suggest that we use an static ordinal counter (AtomicInteger) and count that up and use this for the discriminator. Guess this makes most sense.

rhuss:
@jimmidyson just changed the sorting again to use `List` instead of `SortedSet` . The sort costs a negligible and hopefully the sort of a list doesn't loose any elements even when compareTo() is implemented wrong. However, I'm pretty confident that it should work now.

If there are no objections I would merge that PR now.

jimmidyson:
:+1: :tada:

rhuss:
[merge]

PR #711, opened 2016-12-06
lets improve the validation error message if there is some validation error on some resource so the user has a clue whats wrong

jstrachan:


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -417,30 +419,36 @@ public boolean accept(File pathname) {
...
-        // Enrich descriptors
-        enricherManager.enrich(builder);
+            // Enrich descriptors
+            enricherManager.enrich(builder);
 
-        return builder;
+            return builder;
+        } catch (ConstraintViolationException e) {
+            String message = ValidationUtil.createValidationMessage(e.getConstraintViolations());
+            getLog().error("ConstraintViolationException: " + message, e);
```
rhuss: Could we please use our own logger (member `log`) and using the %s format here, too ? That would be more consistent.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -417,30 +419,36 @@ public boolean accept(File pathname) {
...
 
-        // Enrich descriptors
-        enricherManager.enrich(builder);
+            // Enrich descriptors
+            enricherManager.enrich(builder);
 
-        return builder;
+            return builder;
+        } catch (ConstraintViolationException e) {
+            String message = ValidationUtil.createValidationMessage(e.getConstraintViolations());
```
rhuss: I'm ok to move out the validation of the message to the outside, but is there any other place where such a validation could happen ? If this is not likely, we could also leave it in this MoJo.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
jstrachan: unfortunately none of the validation exceptions thrown by kubernetes-model have any message at all; so this is a workaround to avoid the message being just "ConstraintViolationException" - which if you have lots of YAMLs coming from lots of dependencies isn't particularly helpful figuring out whats wrong.

davsclaus:
lgtm

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
jimmidyson: The validation exceptions are just raw exceptions throws by Hibernate Validator. Handling constraint violations is up to the consumer, in this case f-m-p.

jstrachan: when you throw a ConstraintViolationException you have a choice to pass in a meaningful text message that will appear in any output in, say, maven or JUnit. 

However kubernetes-model chooses not to pass in any message at all (and the base class of ConstraintViolationException helpfully chooses not to pass any default message either even if you have 20 violations :)

it'd be nice for all statements like `throw new ConstraintViolationException(violations)` in kubernetes model to use a helper method to generate a human useful error message. 

I've raised a separate issue to address this generically in kubernetes-model: https://github.com/fabric8io/kubernetes-model/issues/178

jimmidyson: In a library like kube model, there are various consumers that require the error message in different formats. It's up to the consumer to format those validation errors.

Regardless, the "problem" actually lies in sundrio generation rather than in the kube model - will comment on the issue you raised.

jstrachan: whether or not you supply a message to `new ConstraintViolationException()`  the violation objects are available via `getConstraintViolations()` so that any downstream consumer can format the violations however they want. 
So thats not an argument for turning the ConstraintViolationException exception into a meaningless exception with no message at all ;)
A library could even catch the exception, change the message format and rethrow if it really wished to do so.

jimmidyson: Not sure a lack of message makes it a "meaningless" exception but I get your point. Over to @iocanel to fix.

jstrachan: @jimmidyson well there's no other meaning other than the class name `ConstraintViolationException` which isn't terribly useful when kubernetes resources are so numerous and large

jstrachan: @rhuss already migrated to `log.error` in a separate commit

jstrachan: @rhuss see https://github.com/fabric8io/fabric8-maven-plugin/pull/711/commits/2c72ff2290a3ff3e4b226df581ea462e30ccb771

jimmidyson: @jstrachan As you've written in your code, meaning comes from the violations in `e.getConstraintViolations()` & the context it's used in (e.g. validating a file, validating your generated Service resources, etc), which only the caller knows. Seems like the message in the thrown exception will just be a formatted set of the contained ConstraintViolations & the caller will need to add context to that anyway to tell the user what's actually happened. Anyway... enough of this joy.

jstrachan: the thing is though without an actual `message` parameter to the exception, nothing at all about any of the violations ever appears in any stack trace - leaving a very frustrated programmer with no clue. Usually stack traces are pretty meaningful - as its common practice to put messages in exceptions as to what went wrong. Passing a message is a nice, polite thing to do - folks can always reformat a better message. But any message is better than none. e.g. just a hello world formatting of the validation exceptions will tell you *exactly* what failed to validate and why in your kubernetes resources; which is pretty damn useful

jimmidyson: Ah now I get your point! There was no mention of a stack trace before. Missed that important piece of info.

jstrachan: the origin of this issue was maven just said build failed due to: `ConstraintViolationException`. As you can imagine - thats not terribly useful ;)

rhuss:
[merge]

PR #726, opened 2016-12-09
Initial setup for f-m-p integration tests

rhuss:
Shamlessly stolen from vertx-maven-plugin ;-)

It use Citrus for doing more relaxed validation so that values can be ignored, too.


rhuss:
[merge]

PR #727, opened 2016-12-09
Add support Citrus validation matchers

rhuss:
These are described here: http://www.citrusframework.org/reference/html/validation-matchers.html#validation-matcher


rhuss:
[merge]

PR #731, opened 2016-12-12
improve vertx generator and enricher

cescoffier:
* Add an enricher to add health checks to Vert.x applications
* Add documentation about the Vert.x generator
* Add documentation about the enricher
* Small improvements of the generator

### DIFF
--- a/doc/src/main/asciidoc/inc/_enricher.adoc
+++ b/doc/src/main/asciidoc/inc/_enricher.adoc
@@ -262,49 +262,82 @@ These values can be configured by the enricher in the `fabric8-maven-plugin` con
 ==== f8-vertx-health-check
 
 This enricher adds kubernetes readiness and liveness probes with Eclipse Vert.x. The probes are added if the projects
-uses the Vert.x Maven Plugin or depends on the `io.vertx:vertx-core` artifact.
+uses the Vert.x Maven Plugin or depends on the `io.vertx:vertx-core` artifact **and** the path is explicitly configured.
 
 The enricher will use the following settings by default:
 
 - port = `8080`
 - scheme = `HTTP`
-- path = `/`
+- path = _none_ (disabled)
+
+To enable the health checks, configure the probed path using:
+
+* the `vertx.health.path` project properties (`<vertx.health.path>/ping</vertx.health.path>`)
+* the `path` in the `fabric8-maven-plugin` configuration:
+
+[source, xml]
+<plugin>
+    <groupId>io.fabric8</groupId>
+    <artifactId>fabric8-maven-plugin</artifactId>
+    <version>3.3.0</version>
+    <executions>
+    <execution>
+    <id>fmp</id>
+    <goals>
+      <goal>resource</goal>
+      <goal>helm</goal>
+      <goal>build</goal>
+    </goals>
+    </execution>
+    </executions>
+    <configuration>
+    <enricher>
+    <config>
+      <vertx-health-check>
+        <path>/health</path>
+      </vertx-health-check>
+    </config>
+    </enricher>
+    </configuration>
+</plugin>
 
 
-These values can be configured by the enricher in the `fabric8-maven-plugin` configuration as shown below:
+
+These path, port and scheme can be configured by the enricher in the `fabric8-maven-plugin` configuration as shown below:
+
 [source,xml]
-      <plugin>
-        <groupId>io.fabric8</groupId>
-        <artifactId>fabric8-maven-plugin</artifactId>
-        <version>3.3.0</version>
-        <executions>
-          <execution>
-            <id>fmp</id>
-            <goals>
-              <goal>resource</goal>
-              <goal>helm</goal>
-              <goal>build</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
-          <enricher>
-            <config>
-              <vertx-health-check>
-                <port>4444</port>
-                <scheme>HTTPS</scheme>
-                <path>health/myapp</path>
-              </vertx-health-check>
-            </config>
-          </enricher>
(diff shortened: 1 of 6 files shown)


davsclaus:
I wonder if there is a better way of detecting if health checks can be in use for vert.x as I don't see a standard way in vert.x for that. How would you assume that every vert.x application has port 8080 enabled and uses the root path for a health check?

cescoffier:
@davsclaus There is no way to detect it. The enricher comes with sensible default and can be disabled per configuration. Even if we would have a dependency to handle health check (which we would have soon), it does not mean it is used in the code, and which path (URL) is going to serve the response.

davsclaus:
I dont think this is a good idea to have this enabled by default, as you will have k8s call the root context path every X interval which gets mixed up with regular service calls. It must be a special endpoint for only health checks which IMHO should come as a sub module in vert.x that people can easily turn on. Or we write our own fabric8-vertx-healthcheck sub module that is easy to turn on.

cescoffier:
Even with a module, you have no way to be sure it is used and on which "route" it will be used. Unlike spring boot and all these declarative approaches, Vert.x is API-first (like node). So having a specific module does not work. What would this module do by itself? Nothing, because the only way for a Vert.x application to use this module is to use it explicitly in the code. No magic. 

If you are looking for more advanced health check, we are working on such kind of module. But again, just declaring in your dependencies it does nothing, on purpose. This module has also be designed to build advanced health check procedure. Simple application won't need it, and just returning `200 - OK` is generally fine.

If you target "/health" by default, there is a good chance that this path is not served, so would return a 404, which will be considered as not ready by Kubernetes. Is is an application issue? Not sure, as for the application POV, it is probably ready (or alive).

If the application developer is worried that it uses "/", the developer just has to change the path to a served route (but at least he is aware of this).

davsclaus:
Sorry I am -1 on such a thing to add magic OOTB when you build an application. This makes fabric8 look bad.

If people want to have vert.x health check they should configure this manually with the <configuration> in the fmp plugin, and add whatever vert.x code he/she needs.

Until vert.x comes with some easy to use health check endpoints that end users can easily install with a few lines of code, which we could even code parse and detect, or something else, we should not add magic health check in k8s that otherwise makes it fail to deploy.

I suggest you separate this PR and do a PR with the cleanup and health check separated. Then we can get the cleanup merged, and the health-check needs more discussion.

cescoffier:
We can disable the default probe until the user specifically configure the path. If even with this it's a showstopper, I will move the enricher in its own module somewhere else.

davsclaus:
Yes having to explicit configure the path to turn it on can be a compromise.

cescoffier:
@davsclaus I've updated the PR, please have a look.

jimmidyson:
The purpose of the different probes need to be remembered.

Liveness probe is there to ensure that the process is still, um, alive. Using a default path like `/` is fine for that to my mind, you don't care about the response, you just care that it's serving something in a reasonable amount of time. If a liveness probe fails, the pod is restarted by Kubernetes.

Readiness probes on the other hand are there to check that the app is able to serve successfully. That should, preferably, involve more in-depth checks: e.g. that the app can connect to the database, etc. This is used by Kubernetes to take the pod in & out of service without restarting it: if it fails a readiness probe then the pod is removed as a service endpoint. The pod is not added as a service endpoint until it passes it's first readiness probe: it will not receive traffic until then.

If you don't have a readiness probe then Kubernetes will consider it a Ready pod immediately & route traffic to it, perhaps before it's really ready to serve which can look like an outage to the service consumer so I'd recommend having a basic readiness probe in there, even if that is the same as the liveness probe, just checking a default request returns 200.

jimmidyson:
There's some good info on it in [Kubernetes docs](http://kubernetes.io/docs/user-guide/pod-states/#when-should-i-use-liveness-or-readiness-probes).

cescoffier:
@jimmidyson I was wondering if we should have 2 different configurations: liveness and readiness. By default, it would use the same "configuration" for both, but if you explicitly provide a path for readiness, it would use it for the readiness probe.

for Vert.x is could be:

* `vertx.health.path`: default path
* `vertx.health.readiness.path`: specific path for readiness
(same in the XML config)

I didn't see this in the existing enrichers.

jimmidyson:
:+1:  Liveness & readiness probes should be separately configurable, they serve different purposes.

cescoffier:
The PR has been updated to allow configuring a different path for the "Readiness" probe. Checks can be enabled / disabled separately.

rhuss:
@cescoffier @davsclaus the canonical way to switch on enricher on and off is by including / excluding them from the enrichers. This was one reason to have a fine granular component model, which is also backed by profiles for pre-configured sets of enrichers & generators.

Enabling them but switching them off via configuration is of course technical possible, but surprises users who explicitly enabled the enricher. 

Currently I'm [reworking](https://github.com/fabric8io/fabric8-maven-plugin/pull/709) the profile mechanism so that we have a list of enrichers / generators which build up a *default* profile. All other enrichers which are then _optional_ and can be enabled with 

```
<enricher>
   <include>vertx-healthcheck-enricher</include>  
</enricher>
```

and then individually configured, too. Also one could define many profiles, one could be a `vertx` profile which has this enricher enabled. The idea of profiles is described in https://maven.fabric8.io/#profiles but please remember, the semantics will change slightly. 

In light of this moving target, I suggest to merge it in as it is, but would opt for going as an _optional_ enricher as soon as we have this in place, so that it can be switched on like other _optional_ enrichers (e.g. https://github.com/fabric8io/fabric8-maven-plugin/pull/698).

davsclaus:
ok to test

davsclaus:
[merge]

PR #735, opened 2016-12-14
Several changes: IT-Tests, visitor pattern for healthcheck enricher, defined order of enrichers

rhuss:
* Added a spring-boot IT test to verify health checks
* Refactored DefaultServiceEnricher + the HealthCheckEnrichers to use a visitor pattern to examine the model
* Fixed each enricher's order to be independent whether the fabric8 enrichers or the standard enrichers looked up first. That lead to nasty effects when the 'creator enrichers' for adding objects run suddenly later in the chain. This will be fixed more rigorously with introduction of the default profile in #709


rhuss:
[merge]

PR #738, opened 2016-12-15
Handle default value and lowercase answers in maven prompts.

isavin:
When doing fabric8:cluster-start:

> [INFO] F8: Please add the following to your ~/.bashrc:
> [INFO] F8:   export PATH=$PATH:/home/comp_/.fabric8/bin
> Would you like to add the path setting to your ~/.bashrc now? (Y/n):

The capital "Y" suggests that is the default value but just hitting Return will be interpreted as "n". Same for "y" and "yes".


davsclaus:
[merge]

PR #745, opened 2016-12-23
OSFUSE-391: f-m-p annotates services with prometheus.io annotations by default

lburgazzoli:


davsclaus:
ok to test

jimmidyson:
@rhuss would this be better dealt with using a RHT enricher profile as default once profiles are properly supported?

rhuss:
@jimmidyson it makes absolutely sense to extract the prometheus annotations from the `DefaultServiceEnricher` since this is a different task and only then it is possible to switch the annotation on and off. Thanks for that @lburgazzoli !

But you are absolutely right, it makes sense to handle this with a dedicated `redhat-default` profile which can be used as the default when using a `-redhat` version (same as we do for the base images). But for this we need #709 finished (some unit tests and docs are missing, but is already functional otherwise) so that we can have _optional_ enrichers which are not activated by default (currently all enrichers mentioned in `enrichers-default` are active by default).

However, I'm afraid that's too late for FIS 2 since its (a) already quite a bit away from this branch and (b) some bigger change which needs to be retested rigorously. If time permits I try to tackle #678 in the next days, as this has also an higher impact (but is imo absolutely necessary for the long run to cover all the use cases we discovered in the last time).

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/PrometheusEnricher.java
```diff
@@ -0,0 +1,78 @@
...
+                return annotations;
+            }
+        }
+
+        return super.getAnnotations(kind);
+    }
+
+    private String findPrometheusPort() {
+        String prometheusPort = getConfig(Config.prometheusPort);
+        if (!Strings.isNotBlank(prometheusPort)) {
```
rhuss: Wouldn't a `if (Strings.isBlank(prometheusPort))` a bit less confusing ?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/PrometheusEnricher.java
```diff
@@ -0,0 +1,78 @@
...
+            for (ImageConfiguration configuration : getContext().getImages()) {
+                List<String> ports = configuration.getBuildConfiguration().getPorts();
```
rhuss: Not sure whether this can happen here, but since `ImageConfiguration` is inherited from d-m-p, where an `null` BuildImageConfiguration is allowed (e.g. when using only a RunImageConfiguration) it is not _guaranteed_ that this is non-null (although it doesn't make sense). I wonder whether we shouldn't be passive here and ignore null values (as it wont have a prometheus port for sure.

lburgazzoli:
btw, this is a back-port of an old change done on master in rh branch

enricher/fabric8/src/main/resources/META-INF/fabric8/enricher-default
```diff
@@ -5,3 +5,4 @@ io.fabric8.maven.enricher.fabric8.IconEnricher
 io.fabric8.maven.enricher.fabric8.ExposeEnricher,1000
 io.fabric8.maven.enricher.fabric8.SpringBootHealthCheckEnricher
 io.fabric8.maven.enricher.fabric8.KarafHealthCheckEnricher
+io.fabric8.maven.enricher.fabric8.PrometheusEnricher
```
rhuss: While we are at it, I don't know whether an important bug fix made it already to branch `-redhat-80` which caused enrichers with the same order to overwrite each other. (I will check that later and send an appropriate PR in case).

However, I suggest to use orders on _all_ enrichers since the enrichers from the modules _fabric8_ and _standard_ are merged into one list and if no order are given explicitely the order is not really deterministic (it depends which jar is found first in the classpath* lookup). Again an argument for #709 which lets define an overall order for all enricher directly without resorting to order number and conventions across module (which again introduces implicit and hard to detect dependencies).

rhuss: See the current `default-enrichers` from master how it is solved for the moment.

rhuss:
@lburgazzoli ah, ok ;-) haven't checked that it is in master already yet (just in recover mode from a nasty flu)

enricher/fabric8/src/main/resources/META-INF/fabric8/enricher-default
lburgazzoli: btw, this enricher is removed by the pipeline in redhat branch so I do not thing it matters now, but should be fixed on master

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/PrometheusEnricher.java
lburgazzoli: Going to send a PR on master

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/PrometheusEnricher.java
lburgazzoli: Going to send a PR on master

rhuss:
[merge]

PR #747, opened 2016-12-23
Backport of important fix for lookup of enricher entries

rhuss:
* Backport of important fix for lookup of enricher entries with the same order number
* added explicit order number to every enricher to ensure consistent enricher order
* minor update for PrometheusEnricher

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/PluginServiceFactory.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/PluginServiceFactory.java
@@ -48,8 +48,11 @@ public final class PluginServiceFactory<C> {
     // Parameters for service constructors
     private C context;
 
-    public PluginServiceFactory(C context) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/fabric8/maven/core/util/PluginServiceFactory.java
```diff
@@ -235,7 +238,8 @@ public int hashCode() {
 
         /** {@inheritDoc} */
         public int compareTo(ServiceEntry o) {
-            return order - o.order;
+            int ret = this.order - o.order;
+            return ret != 0 ? ret : this.className.compareTo(o.className);
```
rhuss: Problem was that `equals()` and `compareTo()` were not consistent and a `TreeSet` always used `compareTo() == 0` for equality.

lburgazzoli:
lgtm

jimmidyson:
[merge]

PR #749, opened 2016-12-30
Update pom property version.fabric8 to 2.2.185

rawlingsj:


rawlingsj:
The release train is stuck because of this fabric8.version upgrade PR.  If this isn't an error and is just a case of updating the test then it it's probably a simple change.

The `getMetaData()` retuns null here https://github.com/fabric8io/fabric8-maven-plugin/blob/master/plugin/src/test/java/io/fabric8/maven/plugin/enricher/EnricherManagerTest.java#L93

In case anyone who knows more about this that isn't on PTO at the moment then for the release pipeline to continue (which is needed for fabric8 to run on kubernetes 1.5.x) these are the steps:

```
git clone git@github.com:fabric8io/fabric8-maven-plugin.git
cd fabric8-maven-plugin
git fetch origin pull/749/head:fixPR749
git checkout fixPR749

[resolve issue]

git commit -a -m 'resolved merge issues caused by release dependency updates'
git push origin fixPR749:versionUpdatecd8ff6fe-e071-4133-9f41-f7990d760023
```

jstrachan:
@ro14nd any ideas why `getMetadata()` now returns null? I guess something's changed in the underlying kubernetes model?

jstrachan:
@iocanel @ro14nd it looks like the cause is that this line of code is no longer invoked: https://github.com/jstrachan/fabric8-maven-plugin/blob/240246e55e2b7d8cc6a8d4d68e54fecabfde8eed/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ProjectEnricher.java#L69-L69 

when being processed with a List containing a single `ReplicaSetBuilder`. Any ideas @iocanel? This looks like a regression in the kubernetes-model?

PR #767, opened 2017-01-09
OSFUSE-479: Add route for fabric8.deploy.createExternalUrls for OpenShift regardless of Service Type

### LINKED ISSUE #756: OpenShift route generation quirks
### Description

Currently the generation of OpenShift `Routes` is broken and doesn't work in the quickstarts. See https://issues.jboss.org/browse/OSFUSE-479 for the downstream issue.

The main reason is, that although `-Dfabric8.deploy.createExternalUrls` is applied to `fabric8:apply` no Route is created when the `Services` found dont have as type `LoadBalancer`. IMO this is a to harsh restriction, since the user explicitly request the generation of a route, so it should be created regardless of the type (and the type should be changed to `LoadBalancer` if this is required). I know, that route / ingress creation is nowadays delegated to exposecontroller, but this is not of much help for the downstream products which do not have an exposecontroller available (not productised). So we need a direct solution here.

At least as it stands now, `fabric8.deploy.createExternalUrls` is useless for auto-generated Services with service type `ClusterIP`.

@fabric8io/maintainers  any opinion on this ? My suggestion is to move the route creation as post-process step which would (a) change the service type to `LoadBalancer` and (b) create the route.

rhuss:
When using this flag, the only precondition is now that only a single service is exposed.
Relates to https://issues.jboss.org/browse/OSFUSE-479

Fixes #756.

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ApplyMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ApplyMojo.java
@@ -232,10 +232,10 @@ public class ApplyMojo extends AbstractFabric8Mojo {
 
     private ClusterAccess clusterAccess;
 
-    public static Route createRouteForService(String routeDomainPostfix, String namespace, Service service, Log log) {
+    private Route createRouteForService(String routeDomainPostfix, String namespace, Service service) {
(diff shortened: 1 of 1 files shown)


rhuss:
[merge]

PR #769, opened 2017-01-10
Support for ImageStreamTags in generators

rhuss:
If a configuration `fromMode == "istag"` is given for a generator, then an ImageStreamTag is used for the base image. This can be either given in the `from` config in the format "namespace/stream-name:tag" or a default image is used based on the generator (similar like for the plain docker case)

The deault imagestream tags are "openshift/fis-java-openshift:2.0" and "openshift/fis-karaf-openshift:2.0" for the repective generator.

The fromMode can also be set from the command line with '-Dfabric8.generator.fromMode=istag".

Open question is whether to make the "istag" mode the default when running under OpenShift. However since the ImageStreamTag must already exists, this can lead to errors (in the plain docker case it would be simply pulled from a registry). A more sophisticated default algo could be: Check first whether such a default istag already exists. If so, use the "istag" mode. If not, user the plain docker mode.

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/config/OpenShiftBuildStrategy.java
+++ b/core/src/main/java/io/fabric8/maven/core/config/OpenShiftBuildStrategy.java
@@ -24,6 +24,7 @@ package io.fabric8.maven.core.config;
  */
 public enum OpenShiftBuildStrategy {
 
+    // Constants used to extract extra information from a `fromExt` build configuration
     /**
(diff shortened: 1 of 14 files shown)


rhuss:
Going to merge this today since its needed for FIS 2.0. Anyone fancy for a review ? Otherwise I would self-merge later on if everyone is too busy.

rhuss:
[merge]

PR #771, opened 2017-01-11
Better error message when parsing of a fragment fails

rhuss:
Related to https://issues.jboss.org/browse/OSFUSE-416


rhuss:
[merge]

Issue #768, opened 2017-01-10
No such generated manifest when running `mvn fabric8:build -Popenshift`


iocanel:
In a new project I am running:

    mvn fabric8:build -Popenshift 

which fails with:

    [ERROR] F8: No such generated manifest file: /tmp/quick_rest_springboot-tomcat/target/classes/META-INF/fabric8/openshift.yml

### Description

Noticed this glitch while I was looking at: https://issues.jboss.org/browse/OSFUSE-483

This goes away if I add `fabric8:resource` into the mix. 

### Info

* f-m-p version :  fabric8-maven-plugin:3.1.80.redhat-000004
* Maven version (`mvn -v`) : 3.3.9

iocanel:
@rhuss: Do we need to explictly specify fabric8:resource?

Can we automatically generate the resources, if we need them and they are not available or explicitly generated?

rhuss:
@iocanel could you please paste the stacktrace coming from `mvn -e` ? 

Normally a `fabric8:build` should not require a `fabric8:resource`, however it still makes much sense to call this before `fabric8:build` as the generated descriptors are also included in the artifact (jar) which are then part of the image (so it as much required a 'resources' is required for a 'compile' goal). Therefore when we bind both `fabric8:resource` and `fabric8:build`  both to executions so that they typically always called in the right order/combination when using standard maven phases/goals.

iocanel:
[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.1.80.redhat-000004:build (default-cli) on project springboot-rest: No such generated manifest file: /tmp/quick_rest_springboot-tomcat/target/classes/META-INF/fabric8/openshift.yml -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:fabric8-maven-plugin:3.1.80.redhat-000004:build (default-cli) on project springboot-rest: No such generated manifest file: /tmp/quick_rest_springboot-tomcat/target/classes/META-INF/fabric8/openshift.yml
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
[...]

PR #778, opened 2017-01-13
OSFUSE-473: Allows using build goal without using resource goal

dhirajsb:


dhirajsb:
@jimmidyson do we need goal tests now to hit changed code and pass codecov?

jimmidyson:
@dhirajsb Probably a question for @rhuss as the core maintainer of this plugin rather than me. I'm sure it would be good to have _some_ coverage of changes though.

dhirajsb:
@rhuss since the PR is for running build goal, it's kinda hard to integration test it, since it will need a k8s environment, and the generated imagestream manifest with sha256 tag is not fixed.

rhuss:
Actually this PR is associated with #768 , and the actual question is, whether the generation of these imagestreams tags are required at all. The arguments to remove this are given in the issue above, so I really would like to find another solution to create these sort of tags and avoid a coupling of 'fabric8:resource' with 'fabric8:build' indirectly via the resource descriptors so that they can be used independently. Its not good that the OpenShift state is different whether `mvn clean fabric8:resource fabric8:build` is different when then for `mvn clean fabric8:build`. So simply ignoring when no resources are there is not a solution.

w.r.t to the code coverage: @dhirajsb I see you point that Mojos are hard to test and a lot would have been to be mocked. Since this Mojos tend to grow to monsters, the solution is to refactor stuff into Maven independent objects (like the 'ServiceHub'), so that the Mojos are only a thin layer over the business logic, not containing the logic.

said that, for such a small change, I wouldn't mind to merge nevertheless. Its not the missing test, why I hesitate to include it, but because of the arguments given above.

rhuss:
This PR is superseded by #782

PR #781, opened 2017-01-15
Fix NPE if no vertx configuration for a port is given

rhuss:


rhuss:
This is an initial fix for #781

cescoffier:
👍

rhuss:
[merge]

PR #782, opened 2017-01-16
Make writing of the imagestream tags created to resource file optional (#768)

### LINKED ISSUE #768: No such generated manifest when running `mvn fabric8:build -Popenshift`

rhuss:
In order to avoid dependencies between ResourceMojo and BuildMojo, the resulting imagestreams from a S2I build are written to an extract file, independent from openshift.yml.

This file needs to be specified as config 'imageStreamFile', which is empty by default (so no imagestreams are written by default).

Also extracted the logic into an extra ImageStreamService to make it easier to test (though the test are still missing an will be added soon).

Fixes #768


rhuss:
[merge]

PR #783, opened 2017-01-16
Make writing of the imagestream tags created to resource file optional

rhuss:
This is a backport of #782 to v3.1.80-redhat

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -58,6 +58,11 @@
       <artifactId>javassist</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.jmockit</groupId>
+      <artifactId>jmockit</artifactId>
+    </dependency>
+
   </dependencies>
 
   <build>
--- a/core/src/main/java/io/fabric8/maven/core/service/ImageStreamService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/ImageStreamService.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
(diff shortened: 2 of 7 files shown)


rhuss:
I asked several times (@jstrachan ?) but I can't see any usage currently for FIS:

* The modified openshift.yml is not contained in the generated artefact in `META-INF/classes/fabric8` since the update is already too late (the jar file has already been build in `mvn package`. So if this jar would be used in an aggregation scenario (by specifying this jar as a dependency), the imagestream is not included anyway.
* The ImageStream inserted contains the SHA sum of the image generated, which is very specific to the OpenShift cluster where the image has been build. So its not portable (and hence shouldnt be in a released and distribute file like openshift.yml)
* Since the images stream has already been created in the cluster, `fabric8:apply` wont recreated this.

So thats my current knowledge. If it would be used somewhere (I remember darkly something with refering to ImageStream across namespaces within the Fabric8 CD), your fix won't help much, too (since the stream would always be missing when 'fabric8:resource' has been forgotten :)

At least now `openshift.yml` is consistent regardless of the Maven goals calle, as it is only created/modified by `fabric8:resource`.

dhirajsb:
Ok, sounds like due diligence has been done, and I do agree that it might have been a misnomer to put the tags in the yaml file. I am ok with this change.

rhuss:
I just upate the PR according to the discussion in https://issues.jboss.org/browse/OSFUSE-473 so that we now generated always a `target/artifact-is.yml`  where "artifact" is the image stream name.

rhuss:
[merge]

PR #784, opened 2017-01-18
Osfuse 283 istag is default from mode for redhat

rhuss:
Set `fromMode=="istag"` for `-redhat` plugins, if 

* build mode is "OpenShift"
* No `fromMode` is given directly
* Build strategy is "s2i"

The default imagestream tag is calle "fis-java-openshift" for `-redhat` and "fabric8-java" for upstream.


rhuss:
I'm going to merge this on my own (without review), so that it can be tested by QA (and I'm a bit under time pressure), which hopefully comes back to me in case of an error. mea culpa.

rhuss:
[merge]

PR #785, opened 2017-01-18
Started to move out OpenShift build things to an external service.

rhuss:
This also is related to https://issues.jboss.org/projects/OSFUSE/issues/OSFUSE-483


rhuss:
self [merge] , sorry. But its related to #784, too (and is as urgent)

PR #787, opened 2017-01-19
OSFUSE-455: restored the error message in case of binary build error

nicolaferraro:
Probably while refactoring, the Openshift Builder service has been duplicated and the plugin was using the old one.


core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
```diff
@@ -63,7 +63,7 @@ public Build startBuild(OpenShiftClient client, File dockerTar, String buildName
             }
             if (exp.getCause() instanceof IOException && exp.getCause().getMessage().contains("Stream Closed")) {
                 log.error("Build for %s failed: %s", buildName, exp.getCause().getMessage());
-                log.error("If you are refering to an ImageStream as S2I builder image, please ensure that this ImageStream exists (with 'oc get is')");
+                log.error("If you are refering to an ImageStream as S2I builder image, please ensure that this ImageStream exists (with 'oc get is -n openshift')");
```
nicolaferraro: It's better than pointing the user to the current namespace, because the FIS imagestreams are installed in the "openshift" namespace, and it is the default in f-m-p (rh branch). I'll check if I can improve the message.

jimmidyson: From https://github.com/nicolaferraro/fabric8-maven-plugin/blob/2bd0d1ec0c39651d3653bd0269e87ffa724a6cab/core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java#L56, we should be able to validate the IS for the user perhaps? Not sure, but we should at least be able to get the reference to the IS & hence the correct namespace for it.

nicolaferraro: @jimmidyson printing the users to the right namespace in case they changed the default, or pointing them to check the docker image or the image stream, require some refactoring. I think we can leave as it is and postpone the refactoring (we are going to release the first version shortly).

nicolaferraro: Missed your comment, you're probably right...

core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
```diff
@@ -138,4 +137,25 @@ public void onClose(KubernetesClientException cause) {
...
+        try {
+            BuildConfig build = client.buildConfigs().withName(buildName).get();
+            ObjectReference ref = build.getSpec().getStrategy().getSourceStrategy().getFrom();
+            String kind = ref.getKind();
+            String name = ref.getName();
+
+            if ("DockerImage".equals(kind)) {
+                log.error("Please, ensure that the Docker image '%s' exists and is accessible by OpenShift", name);
+            } else if ("ImageStreamTag".equals(kind)) {
+                String namespace = ref.getNamespace() != null ? ref.getNamespace() : "";
```
jimmidyson: nit: I think this block would be more readable like:

```java
String namespace = ref.getNamespace();
String namespaceInfo = "current";
String namespaceParams = "";
if (!namespace.isEmpty()) {
  String namespaceInfo = "'" + namespace + "'";
  String namespaceParams = " -n " + namespace;
}
```

but not really bothered if you don't agree.

jimmidyson:
add to whitelist

jimmidyson:
retest this please

core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
nicolaferraro: Sometimes I've too much fear of nested ifs :)

Think your version is more readable (I'll add just a null check for the namespace).

jimmidyson:
LGTM thanks! [merge]

PR #791, opened 2017-01-28
Partial fix for #790

### LINKED ISSUE #790: VolumePermissionEnricher should only generate initcontainer annotations for Kubernetes resources
The f-m-p VolumePermissionEnricher enriches openshift resources which it didn't used to do.  This has a knock on affect and can cause problems when deploying apps on openshift. See https://github.com/fabric8io/elasticsearch-cloud-kubernetes/pull/69#issuecomment-275744474

We should update f-m-p to not add the chmod init container for openshift resources and leave it to just kubernetes resources.  

I can have a crack at this unless @rhuss you have a good idea of where the change needs to happen?

rhuss:
This will avoid running the VolumePermissionEnricher for OpenShift. However, the proper solution is to use an appropriate profile where this enricher is not included if you dont want to have this sort of enrichment (which is maybe still a hack ?)

Fixes #790

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/Constants.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/Constants.java
@@ -17,6 +17,8 @@ package io.fabric8.maven.core.util;
 
 /**
  */
+// TODO-F8SPEC : Should be move the to AppCatalog mojo and must not be in the general available util package
+// Also consider whether the Contstants class pattern makes (should probably change to real enums ???)
 public class Constants {
     public static final String RESOURCE_SOURCE_URL_ANNOTATION = "maven.fabric8.io/source-url";
     public static final String RESOURCE_APP_CATALOG_ANNOTATION = "maven.fabric8.io/app-catalog";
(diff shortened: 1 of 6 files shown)


jimmidyson:
IMO this enricher should be opt-in, not opt-out. It is not required for all PV provisioners & as we're learned can actually cause issues.

rhuss:
Would be no problem for me, but then the Fabric8 CD pipeline builds should be started with something like `-Dfabric8.profile=cd` (or whatever name) where in this profile this enricher is enabled. This would also have the advantage, that one really can fine tune the build for CD and do what is required withou affecting all the other use case.

Would this be an issue to introduce a dedicated profile, @rawlingsj ?

rhuss:
Since there is no feedback yet, and I guess the fix is needed for the CD to work, I'm going to merge this now. 

I will also setup a 'fabric8-cd' profile with the things I believe belong to CD and an experimental 'default-ng' with the enricher / generators for a setup for a neutral (i.e. plain Kubernetes / OpenShift) environment. This will happen in an extra PR and can be discussed separately.

rhuss:
[merge]

PR #794, opened 2017-01-31
Removed default of 3 h for a DeploymentConfig on OpenShift

rhuss:
It seems that this long timeout could cause issues to detect errors early in deployments e.g. when the pod fails to start. AFAIK it was introduced to compensate long pull times of images over slow internet connections.

However, since I consider this a rare use case compared to the normal usage (where an image can easily be downloaded in the 10 minute or so default), I wouls like to remove that as default. The timeout can always be set explicitly when coming into a very-slow network situation (where an upfront pull of images might make even more sense).

I open the discussion, whether a default of 3 hours should be used or not. Its **not** about the feature itself, which is quite useful (and which btw, if set, automatically uses a `RollingUpdate` deployment strategy, otherwise no deployment strategy is set).

// @jstrachan @jimmidyson @rawlingsj and everybody interested.

### DIFF
--- a/it/src/it/simple/expected/openshift.yml
+++ b/it/src/it/simple/expected/openshift.yml
@@ -46,10 +46,6 @@ items:
       project: fabric8-maven-sample-zero-config
       provider: fabric8
       group: io.fabric8
-    strategy:
-      rollingParams:
-        timeoutSeconds: 10800
-      type: Rolling
     template:
       metadata:
         annotations:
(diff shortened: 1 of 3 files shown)


jimmidyson:
:+1: from me.

rhuss:
ok, seems that this is ok so far (== no one complaining ;-). We can still revert it back if there are objections.

[merge]

PR #807, opened 2017-02-04
fixes #806

### LINKED ISSUE #806: when combining manifests together to make packages we seem to repeatedly create init containers with the same name which breaks the Deployments
e.g. currently the fabric8-platform includes `gogs` which has 3 init containers called `init` which are all identical

jstrachan:


jstrachan:
[merge]

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -133,8 +134,30 @@ protected void addInitContainer(PodTemplateSpecBuilder builder, JSONObject initC
...
+            initContainers.put(initContainer);
+        }
         builder.editMetadata().addToAnnotations(INIT_CONTAINER_ANNOTATION, initContainers.toString()).endMetadata();
     }
+
+    private boolean duplicateInitcontainer(JSONObject a, JSONObject b) {
+        Object name1 = a.get("name");
+        Object name2 = b.get("name");
+        if (Objects.equal(name1, name2)) {
+            log.warn("Found duplicate init containers with names " + name1 + " so ignoring " + b);
```
jstrachan: so gogs doesn't include > 1 init container. The issue happens when composing packages together (e.g. cd-pipeline in fabric8-platform). It seems the volume enricher fires again on already enriched resources.

Though any time an enricher goes bad, adding multiple init containers with the same name totally breaks the Deployment. So I'd much rather have f-m-p not totally break the fabric8-platform every time some refactoring happens and an error creeps in. So I'd rather a warning over generating something silently that has no purpose.

I'm more than happy for you to fix the root cause though & feel free to polish the code ;)

BTW I explicitly log the init container JSON so its easier to grok which enricher caused the duplicate; I can see things like this cropping up over time as different enrichers generate containers whose names clash

rhuss: No enricher should fired again on already enriched resources introduced by the `DependenciesEnricher`. That must be fixed, that's the root cause.

And if this is fixed, I don't get why the first init container should be used and not the second one (if they differ) ? And why not stop with an error because at least one init container won't run ? Why not renaming the second init container to something unique ?

The fix of course will prevent the build from breaking, but does it not also allow for inconsistencies and bugs to creep in silently (or buried in mega bytes of logs) ?

I really would love to have a more rigour discussion of this kind before merging (guess, that's what reviews are for). 

I'm not against quick fixing a pressing issue with a patch, but maybe we can agree to open a new issue for looking to fix the real cause ?

jstrachan: I prefer to have usable software first and then to move asynchronously in parallel to more rigour. fabric8-maven-plugin had completely broken the fabric8-platform project rendering it utterly useless - so I've merged a trivial patch to make sure fmp doesn't completely a break a release again after some factoring accidentally breaks downstream projects. 

We can over time figure out the more rigorous fix. 

BTW its really trivial to revert this change - thats what SCM is for - if you figure out a better way to fix this and avoid fmp breaking stuff

rhuss: No problem. All I said it would be nice to have an extra issue created for fixing the real cause if you prefer the 'merge-first' model and revert things later in case (which works only smoothly if no much work has been done in between).

Personally I prefer a 'review-first' style since I have the feeling that it's safer to not forget things and helps in communication (== is more direct instead of passively relying on that PR notification mails get tracked).

But thats probably a matter of taste.

jimmidyson: BTW I'm not sure how code changes in f-m-p `breaks downstream projects`? Upgrading the plugin and not testing that the upgrade works properly is what `breaks downstream projects` IMO. Bugs creep in to dependencies regularly and need testing before upgrading. Plugins should be treated the same as dependencies.

jimmidyson: And I agree with @rhuss that any trivial fix still should have code review, but as it's trivial it should be really quick to review and merge. Maybe not instant like self-merging but quick enough for almost all uses.

jstrachan: well if you're not sure how that can happen - just review the fabric8-platform git commit logs ;) f-m-p released a new version; then fabric8-platform's YAML became unusable.

Though I'm hoping we can setup CI tests to prevent this kinda thing ever happening again really - it happens very frequently with changes in kubernetes model, client & fmp. Now we actually have a working CD cluster finally; we can hopefully get that in place soon...

jstrachan: @jimmidyson if we can stop fmp and kubernetes-client from breaking downstream projects I'm happy to spend a year reviewing each line of code in detail before merging anything. When downstream stuff is broken by an upstream refactor I'm also totally happy merging a temporary fix - afterall we're still reviewing the commit that happened 2 days ago. We don't have to break the world and wait until every line of code is polished to within an inch of its life before we fix things again. I repeat - its super trivial to revert old workarounds if a newer shinier approach is figured out.

But as I just said; having better CI will prevent fmp/kube client refactors from breaking everything downstream so you can both take your time reviewing things in detail before fixes get applied

rhuss: @jstrachan would be really happy to help in avoiding such situations. Is there anything we should test regularly before doing a f-m-p release w.r.t to fabric8-platform ?

There was an issue in the fabric8-platform lately which lead to method mismatch errors because a dependency added to the plugin used a different kubernetes-model than the plugin itself (I documented that in the corresponding PR in fabric8-platform). This what *not* something f-m-p broke, but the dependency added to the plugin deps. I introduced a fix (moving the deps to regular project deps which also works for the dependency enricher) but could test this only superficially (whether the yaml files were created). However it seems, that this was not good enough. Is there any other rigorous test I could perform ? If so, I would run it (even manually) before each f-m-p release.

jimmidyson: Yes code reviews slow things down, superficially at least. The extent they slow things down is dependent on the scope of the change. @rhuss especially is extremely quick at giving feedback so shouldn't slow down the f-m-p significantly but should prevent unknown technical debt being introduced without tracking for when it needs to be cleaned up later. We are also trying to get more maintainers that would help speed up code reviews by sharing the load.

The reason we're spending time discussing this change is not the change itself, but rather the hack a fix, self-merge, review and fix properly later (maybe). While this has a perception of being faster for you, what's actually happening is that other people (@rhuss in this case) are having to spend much longer undoing the changes to make the code base predictable, maintainable and supportable.

In lieu of automated testing for dependency/plugin upgrades, perhaps manual testing before merging upgrades would be prudent?

jstrachan: @rhuss given the complexity and number of independently moving parts its kinda hard to just test fmp or kube-client in isolation and to have real confidence its not gonna break something downstream. e.g. the classloader issue you tried fixing is almost impossible to guess ahead of time; its CI that catches it really.

So having downstream CI which tries to recreate & use the fabric8-platform is the best I can think of. e.g. we really need to deploy the platform using new quickstarts using the new fmp versions and client versions to check there's not an issue there, in the pipeline or in the deployment. ClassLoader or version mismatch issues can crop in anywhere really. 

The easiest way to do that is actually to do lots of releases then use fabric8-platform as a CI test before it can release a tested set of dependencies. I'd be totally happy if we can get CI to catch these issues so if fmp/client do a bad release, we can raise an issue and get a downstream release out soon-ish (where we revert the new fmp/client version as a temporary workaround until the real upstream fix is figured out - hopefully automatically too with nice upstream + downstream feedback).

Though maybe another bit of CI we could try do is using the latest fmp to build a new fabric8-platform distribution and try test it out and see if it works? That would have caught this particular issue at least (though not lots of the other ones).

Once we've automated the downstream fabric8-platform CI tests we should be able to reuse that same CI logic for a snapshot test of fmp upstream hopefully. But given the failures that have happened over the last couple of years I'd expect the downstream CI to catch way more issues - but yeah, the earlier we catch things the better

jstrachan: @jimmidyson so a trivial bit of code in a single commit to avoid duplicates (which we probably will always need as duplicates totally fail the Deployment) is gonna take @rhuss longer time to deal with than manually testing fabric8-platform works with all fmp commits going forward? That sounds totally sound logic there :).  

BTW the reviewed refactorings of fmp which introduced this issue also wasted lots of my time too, which was nothing to do with me, I was just trying to get fabric8-platform released. Yes code reviews can help. But bugs slow us all down too. 

Anyway am hoping we can get automated CI soon though which should save us all time going forward

rhuss: @jstrachan still don't get it why we should silently ignore the second init-container. This init-container is there for a reason and its IMO not the task of the plugin to decide which one of the init containers is the correct one, and ignore the other one. 

Only if both init-containers are the very same, then its ok to ignore one and continue. But assuming they are the same because they have the same name is not enough. 

Otherwise its simply a configuration error which should lead to an error. Agreed that the error at deployment time is too late, so its better to die already when doing the enrichment.

Fixed that with #809.

jstrachan: @rhuss I never said we should silently ignore the second init container. My point is that when these errors crop up now and again due to refactorings of fmp and because we don't yet have sufficient CI across lots of independently released projects - I prefer to focus on generating valid kubernetes resources first and generating warnings instead of breaking things.

Maybe we could have a 'strict' mode that always fails the build instead of generating warnings so folks can turn on strict mode in CI builds to ensure that CI jobs fail if duplicates are generated - but avoid breaking release pipelines?

rhuss: What exactly is a duplicated init-container ? An init-container with the same name or with the same content ? Completely agree that we should ignore a second init-container if it has the same content (with a warning), but would always croak if they have the same name but different content. If not, that will **break** things eventually even with valid Kubernetes resource (because of a missing init-container then).

Let me add that check for equality of the content, so that we are on the safe side. This would cover your use case, too (but [as said](https://github.com/fabric8io/fabric8-platform/pull/142#issuecomment-277751283) there is a simple solution without code change.

jstrachan: Each init-container must have unique names as the Deployment becomes unusable if they have 2 init-containers with the same name - the whole Deployment is rejected by kubernetes in the case of duplicate names. 

Having actual duplicates of the same container metadata with different names may or may not break things - doing a chmod twice is fine but other init stuff may barf if done twice. And yes - ignoring a different init-container that just happens to have the same name will lose init-container logic that will probably break things as well. (Though at least the Deployment won't be rejected by kubernetes - but something bad will most likely happen down the line).

Given the plugin & configurable nature of fmp it'd be nice to try nail the logic going forwards as I can see lots of different developers writing lots of generators+enrichers+custom metadata of various shapes and sizes for different use cases & frameworks and with ordering and whatnot I'm sure clashes will happen. Probably all possible clashes will happen at some point ;) (The same kinds of clashes can happen all over the place TBH with services/ports/containers of the same name too).

So maybe we should:

* if the containers are real duplicates log a warning; we could maybe even get clever and tell the developer that its probably an ordering thingy fix on the enricher (e.g. your patch to the volume enricher to avoid dupes) to help them really fix the problem properly. This warning in strict mode could fail the build as well (for CI jobs)
* if the containers are different but the names are the same we could either log a warning and change the name of the 2nd one automatically to something unique (which is a kind option for downstream projects; trying to work around mistakes/regressions in fmp or fmp's configuration and keeping the lights on - and the init container name is fairly meaningless in the grand scheme of things anyways; adding a postfix to enforce uniqueness sounds OK to me as a temporary work around) or fail the build until the a new release of fmp comes out with updated enrichers or the user figures out a manual workaround

rhuss: Actually I implemented the logic as you described here https://github.com/rhuss/fabric8-maven-plugin/blob/61ff566633f7c54b3b666cfe0514296a7474389b/enricher/api/src/main/java/io/fabric8/maven/enricher/api/util/InitContainerHandler.java#L80-L90

(with bailing out in a same-name, different-content situation).

btw, I don't think clashes are very likely as init container names should reflect their purpose. It was simply unfortunate it was decided to use `init` as init-container name for the volume enrichment. Its now called like the enricher: `fmp-volume-permission`, which I consider to be a good practice as enricher names have to be unique, too.

PR #814, opened 2017-02-09
Fix image stream issue when using no tag.

rhuss:
Now "latest" as tag is assumed


rhuss:
This PR also removes a duplicate class which slipped in (along with the duplicated test, so the coverage is decreasing a bit)

rhuss:
[merge]

PR #815, opened 2017-02-10
Fixes #799: New "Watcher" abstraction

### LINKED ISSUE #799: Introduce new "Watcher" abstraction to replace `fabric8:watch-spring-boot`
Currently there is a top leve goal `fabric8:watch-spring-boot` which directly connects spring-boot as a technology to the the abstract goal. We already have `fabric8:watch` so the user would need to _know_ in advance what tech is used. 

This is (a) in contrast to the Maven philosophy and (b) is too much of coupling (and its too much to type). E.g. should we introduce goals like 'watch-vertx', 'watch-wildfly-swarm' then later ? 

This issue does **not** question the awesome functionality that `watch-spring-boot`  brings on the table. The goal should be to hide this functionality behind `fabric8:watch` and automatically selected by the plugin when it detects that its about s spring-boot app.

Therefore I suggest to introduce the new concept of a `Watcher` (on the same footprint as `Enricher` or `Generator`) which abstracts a way the concrete implementation _how_ to watch. This is flexible, easier to use from an UX perspective and open for additional watching features).

nicolaferraro:
I've moved all (2) watchers out of their mojos and unified their goals. Now watchers are treated like generators and enrichers (although only one watcher is selected and executed, the first one that is applicable). The spring-boot watcher will be used for spring-boot projects, while the other one (called it `docker-image`...) will be the fallback one. The fallback one is not applicable on Openshift.

The two watchers were different in their usage. The spring-boot one forked a maven lifecycle by executing the deploy goal. The docker one required that the user typed all the goals, so I've decided to execute the deploy goal by default for all watchers.

I've removed all the code from the `AbstractTailLogMojo` and introduced a specific service (that is also used by one watcher). Introduced a new `KubernetesClientUtil` and moved there some code that were in the abstract mojo. 

I need to change the docs. I'll do it as soon as this gets approved.
There are a lot of improvements that can be done on both watchers, but I'll open other issues for that. Now they just work as before (just fixed some small bugs on the spring-boot one).

**Note**: the parent pom depends on version `0.19-SNAPSHOT` of the docker maven plugin. [This PR](https://github.com/fabric8io/docker-maven-plugin/pull/701) should be merged in d-m-p and the plugin released before merging this one (I guess TravisCI won't appreciate it :D).


rhuss:
Awesome ! Thanks a lot @nicolaferraro, I will have a look over the weekend.

rhuss:
@nicolaferraro just released d-m-p v0.20.0 and updated this PR. Now it's time for review .... :)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -67,16 +87,10 @@
...
-import java.io.FileFilter;
-import java.io.IOException;
-import java.util.*;
-
-import static io.fabric8.maven.core.util.Constants.*;
+import static io.fabric8.maven.core.util.Constants.RESOURCE_APP_CATALOG_ANNOTATION;
 import static io.fabric8.maven.core.util.KubernetesResourceUtil.isAppCatalogResource;
+import static io.fabric8.maven.core.util.KubernetesResourceUtil.loadResources;
 import static io.fabric8.maven.plugin.mojo.build.ApplyMojo.DEFAULT_OPENSHIFT_MANIFEST;
-import static io.fabric8.maven.plugin.mojo.build.ApplyMojo.loadResources;
```
rhuss: awesome that get rid of this coupling. Let's think also where to move the DEFAULT_OPENSHIFT_MANIFEST constant.

Have I already mentioned that I'm not a big fan of [classes only for constants](http://stackoverflow.com/a/66228/207604) ? (always so negative, those Germans ;-) It might be harder to think a bit for a concrete class where to put the constant on (ideally its used only within a single class anyway), but it reduces coupling and increases [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science))

But, well, probably belongs to the category 'matter of taste'. I only hate when unrelated constants are collected in a `utils.Constants` class because its _technically_ easy since this class is reachable from every (but will hurt for maintenance or when moving things around). Don't know whether this the case here, though.

(as a general note, and because I'm already at it ;-), the same applies for "util" classes which are also accessible from everywhere. It's so tempting to stuff all things into utility classes because so you don't have to think about dependencies.)

rant over ;-)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -490,7 +504,7 @@ private KubernetesList convertToOpenShiftResources(KubernetesList resources) thr
             Controller controller = new Controller(client);
             Set<HasMetadata> oldEntities;
             try {
-                oldEntities = loadResources(client, controller, namespace, openshiftManifest, project, log);
+                oldEntities = loadResources(openshiftManifest);
```
rhuss: you can remove the local variables "namespace" and "controller" above as they were only used for this call.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/AbstractTailLogMojo.java
```diff
@@ -15,285 +15,31 @@
...
 /**
  */
 public class AbstractTailLogMojo extends ApplyMojo {
-    public static final String OPERATION_UNDEPLOY = "undeploy";
-    public static final String OPERATION_STOP = "stop";
-    public static final String FABRIC8_LOG_CONTAINER = "fabric8.log.container";
-    private static final String FABRIC8_LOG_POD = "fabric8.log.pod";
 
-    @Parameter(property = FABRIC8_LOG_CONTAINER)
+    @Parameter(property = "fabric8.log.container")
```
rhuss: thanks for inlining ;-)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/AbstractTailLogMojo.java
```diff
@@ -15,285 +15,31 @@
...
+    protected PodLogService.PodLogServiceContext getLogServiceContext() {
+        return new PodLogService.PodLogServiceContext.Builder()
+                .log(log)
+                .logContainerName(logContainerName)
+                .podName(podName)
+                .newPodLog(createLogger("[[C]][NEW][[C]] "))
+                .oldPodLog(createLogger("[[R]][OLD][[R]] "))
+                .s2iBuildNameSuffix(s2iBuildNameSuffix)
+                .build();
     }
```
rhuss: Looks much nicer indeed.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/LogMojo.java
```diff
@@ -40,7 +42,7 @@
 
     @Override
     protected void applyEntities(Controller controller, final KubernetesClient kubernetes, final String namespace, String fileName, final Set<HasMetadata> entities) throws Exception {
-        tailAppPodsLogs(kubernetes, namespace, entities, false, null, followLog, null, true);
+        new PodLogService(getLogServiceContext()).tailAppPodsLogs(kubernetes, namespace, entities, false, null, followLog, null, true);
```
rhuss: I wonder whether the PodLogService could be already put into the base class, so you don't have to expose the PodLogServiceContext but just a `AbstractTailLogMojos/getPodLogService()` ?

plugin/src/main/java/io/fabric8/maven/plugin/watcher/WatcherManager.java
```diff
@@ -0,0 +1,77 @@
...
+        Logger log = watcherCtx.getLogger();
+        List<Watcher> usableWatchers  = config.prepareProcessors(watchers, "watcher");
+        log.verbose("Watchers:");
+        Watcher chosen = null;
+        for (Watcher watcher : usableWatchers) {
+            log.verbose(" - %s",watcher.getName());
+            if (watcher.isApplicable(ret, resources, mode)) {
+                chosen = watcher;
+                break;
+            }
```
rhuss: Maybe one should still finish the loop to print out all available watchers (and also whether they would apply if not the first was taken).

Somtehing like:

```
Watchers:
- docker-image [ignored]
- spring-boot [selected]
- another-one [ignored]
- yetanother-one 
```

Just to help debugging (which is always hard when things are collected dynamically and using profiles etc.)
```

File: plugin/pom.xml [hunk 7]
```diff
@@ -109,6 +109,16 @@
       <artifactId>fabric8-maven-generator-webapp</artifactId>
     </dependency>
 
+    <dependency>
```
rhuss: It would be super awesome if we could get rid of the [spring-boot dependency](https://github.com/nicolaferraro/fabric8-maven-plugin/blob/3a9ff06301becc26d4f3510a67896f219ab73a04/plugin/pom.xml#L32-L42) in this pom.xml and only have it in the watcher module for now. This plugin pom.xml ideally should be completely agnostic to the specific technology the builds are using (vert.x, spring-boot, wildlfy-swarm, ....)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
nicolaferraro: Yeah, I'm not even a fan of classes for constants and I've seen too many projects with 100+ constants in a `utils.Constants` file... The value of that constant will go eventually into a parameter of the `AbstractFabric8Mojo`, but we need to break some inheritance from the `d-m-p` project mojos before ;)

I think it's natural to define a set of utility classes and then start putting stuff into them, eventually filling them with unrelated things. But you know, when you develop a new thing, you never create a utility class for holding a single method. When an utility class grows (`KubernetesResourceUtil` is one of them..) you start splitting it in multiple utility classes or you extract a service from it. The good thing is that refactoring utility classes is always much easier than refactoring a hierarchy of dependent mojos :)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/LogMojo.java
nicolaferraro: Good idea :+1:

nicolaferraro:
I've made the changes. Improving logging is really useful.. you can discover that the order of watchers is wrong :)

plugin/src/main/java/io/fabric8/maven/plugin/enricher/EnricherManager.java
```diff
@@ -135,9 +135,9 @@ public void enrich(ProcessorConfig config, KubernetesListBuilder builder) {
 
 
     private void logEnrichers(List<Enricher> enrichers) {
-        log.verbose("Enrichers:");
+        log.debug("Enrichers:");
```
rhuss: why do you choose `debug()` over `verbose()` ? Debug is very noisy and probably mostly switched off. The list of enrichers/generators though are more on the level of the build step of a Docker or S2I build which are also shown with verbose.

I probably was a bit misleading in my commetn with _is useful for debug_ ;-)

nicolaferraro: Ok, step backward.. didn't know how to see verbose logging :)

rhuss:
thanks !

rhuss:
[merge]

PR #817, opened 2017-02-10
Add automated deployment and system tests after a fabric8-maven-plugi…

rawlingsj:
…n release

After a fabric8-maven-plugin is released and updated its downstream dependencies we will:
 - build a snapshot version of downstream projects including fabric8-devops and quickstarts
 - update fabric8-forge so that the system test quickstarts include the new fabirc8-maven-plugin
 - creates a new kubernetes cluster
 - deploys the snapshot fabric8-platform
 - runs the end to end system tests that create new projects, runs pipelines and rolling updates
 - notifictions on failure including fabric8 console link to use and investigate
 - notification on success and automatically tear down cluster

This needs to run at the end of the release rather on each PR because we need the updates on the fabric8-quickstarts that are added to forge to include the new version.  Also lets see how many tests we run as we dont want large GKE costs.  This extension to the pipeline takes ~1 hour to complete.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,6 +18,7 @@
 def dummy
 mavenNode {
   checkout scm
+  readTrusted 'release.groovy'
   sh "git remote set-url origin git@github.com:fabric8io/fabric8-maven-plugin.git"
 
   def pipeline = load 'release.groovy'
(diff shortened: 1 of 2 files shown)


rawlingsj:
@jstrachan @ro14nd this is the initial look at automating all those testing steps needed following a fabric8-maven-plugin release.  I've been testing and it works really nicely!  Still things to improve but this hopefully removes a HUGE amount of effort and worry, ensuring we get fast feedback on possible impacts to the fabric8 platform!

jstrachan:
@rawlingsj awesome!!!!

rhuss:
@rawlingsj thanks a lot ! Its was really one of the last missing pieces ;-) think now releasing gets smoother.

BTW, is there a way how I could let this run upfront before I do a release ? Or is the setup to involved ? Maybe that doesnt work because as you said the quickstarts already need the new version. (personally I prefer to not break things instead of breaking and then fixing, but if its to complicated I can live with the latter, too).

rhuss:
lgtm

rhuss:
[merge]

rawlingsj:
@rhuss yeah I originally wanted this to run as a CI job on each PR but I couldn't figure out a nice way to update the fmp in the quickstarts which are then included in forge.  I think thats where we need to aim for, I've just not figured out a way to do that yet.  

We might hit a problem or two early on so I figured as its at to the end of a release if it screws up for some reason, at least it wont block a release or PR.

rawlingsj:
BTW its really quite cool even if I do say so myself :) you get a link to the console and you can follow the system tests that @jstrachan added.  We can sit back and watch the two projects pipelines running and perform rolling updates in realtime.

PR #819, opened 2017-02-10
Awareness of Spring Boot context path to readiness and liveness checks

srjturner:
Spring Boot apps with a non-empty context root (i.e. a value for "server.context-root" in application.properties) currently get a readiness and liveness check that points to the wrong path ("/health" rather than "/context-root/health"). This obviously means that Kubernetes will continually redeploy the pods because it thinks they are unhealthy. This PR fixes that. Not sure if there is a place for this in the documentation - it's really something that should "just work".


rhuss:
Sorry for the delay. Looks good to me (its not a big change anway ;-)

Thanks !

[merge]

PR #827, opened 2017-02-14
Fixes #723: Add env variables to container

### LINKED ISSUE #723: Add <env> specified env variables to be added to generators
### Description

Our provided generators should respect the `<env>` variables section in the config and add environment variables set there.

nicolaferraro:
Added env variables to the container through the image enricher.

The order of the `fmp-image` enricher was wrong. It must run after the `fmp-controller` to be effective. The enricher simply did nothing because the Deployment/DC/RS/DS were not present in the builder.

I added an integration test to prevent this happening again.

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ImageEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ImageEnricher.java
@@ -17,15 +17,40 @@
 package io.fabric8.maven.enricher.standard;
 
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
 import io.fabric8.kubernetes.api.builder.TypedVisitor;
-import io.fabric8.kubernetes.api.model.*;
-import io.fabric8.kubernetes.api.model.extensions.*;
+import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.EnvVar;
+import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
+import io.fabric8.kubernetes.api.model.PodTemplateSpecBuilder;
+import io.fabric8.kubernetes.api.model.PodTemplateSpecFluent;
+import io.fabric8.kubernetes.api.model.ReplicationControllerBuilder;
+import io.fabric8.kubernetes.api.model.ReplicationControllerFluent;
+import io.fabric8.kubernetes.api.model.ReplicationControllerSpecFluent;
+import io.fabric8.kubernetes.api.model.extensions.DaemonSetBuilder;
+import io.fabric8.kubernetes.api.model.extensions.DaemonSetFluent;
+import io.fabric8.kubernetes.api.model.extensions.DaemonSetSpecFluent;
+import io.fabric8.kubernetes.api.model.extensions.DeploymentBuilder;
+import io.fabric8.kubernetes.api.model.extensions.DeploymentFluent;
+import io.fabric8.kubernetes.api.model.extensions.DeploymentSpecFluent;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSetBuilder;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSetFluent;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSetSpecFluent;
+import io.fabric8.kubernetes.api.model.extensions.StatefulSetBuilder;
+import io.fabric8.kubernetes.api.model.extensions.StatefulSetFluent;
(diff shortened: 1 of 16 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ImageEnricher.java
```diff
@@ -81,6 +106,9 @@ private void ensureTemplateSpecs(KubernetesListBuilder builder) {
         ensureTemplateSpecsInReplicationControllers(builder);
         ensureTemplateSpecsInRelicaSet(builder);
         ensureTemplateSpecsInDeployments(builder);
+        ensureTemplateSpecsInDaemonSet(builder);
+        ensureTemplateSpecsInStatefulSet(builder);
+        ensureTemplateSpecsInDeploymentConfig(builder);
```
rhuss: Thanks for adding these ;-)

enricher/standard/src/main/resources/META-INF/fabric8/enricher-default
```diff
@@ -4,14 +4,13 @@
...
-io.fabric8.maven.enricher.standard.ImageEnricher,101
-
 # Add a default Deployment, ReplicaSet or ReplicationController if none is given
-io.fabric8.maven.enricher.standard.DefaultControllerEnricher,102
+io.fabric8.maven.enricher.standard.DefaultControllerEnricher,101
+
+# Add image information such as name, image pull policy, environment variables
+# to a container. Controllers (like Deployment, DeploymentConfig, etc.)
+# must be already present
+io.fabric8.maven.enricher.standard.ImageEnricher,102
```
rhuss: btw, the order in this file is not important anymore (and mostly ignored, maybe we should get rid of it altogether).

The order in the enricher chain is completely defined now in the "default" profile (so that also optional enrichers can be defined in this file).

enricher/standard/src/test/java/io/fabric8/maven/enricher/standard/ImageEnricherTest.java
```diff
@@ -0,0 +1,195 @@
...
+                .addNewReplicaSetItem()
+                .endReplicaSetItem();
+
+        imageEnricher.addMissingResources(builder);
+
+        // Validate that the generated resource contains
+        KubernetesList list = builder.build();
+        assertEquals(list.getItems().size(),1);
+
+        String json = KubernetesResourceUtil.toJson(list.getItems().get(0));
```
rhuss: can we extract these assertion into an extra method and use it in the other tests, too ? I'm always a fan of being DRY in tests as well, so that when things changes only a single sport needs to be adapted.

plugin/src/main/resources/META-INF/fabric8/profiles-default.yml
```diff
@@ -61,8 +61,8 @@
   enricher:
     includes:
     - fmp-name
-    - fmp-image
     - fmp-controller
+    - fmp-image
```
rhuss: good point :) guess we should move also fmp-service up, since fmp-controller and fmp-service are both enrichers which 'create' objects. That will be not that critical anymore when #678 is in place.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ImageEnricher.java
```diff
@@ -198,4 +266,22 @@ private void mergeImagePullPolicy(ImageConfiguration imageConfiguration, Contain
...
+        if (env == null) {
+            env = new LinkedList<>();
+            container.setEnv(env);
+        }
+
+        ResourceConfig resource = getContext().getResources();
+        Map<String, String> userEnv = resource != null ? resource.getEnv() : null;
+        if (userEnv != null) {
+            for(Map.Entry<String, String> entry : userEnv.entrySet()) {
+                boolean set = KubernetesResourceUtil.setEnvVar(env, entry.getKey(), entry.getValue(), false);
```
rhuss: Just a minor thing (not required, though): If we could change setEnvVar to return the old value here (and then check getValue() vs. oldValue), we could add both values in the warning messages so that it becomes easier to detect the origin of the values in order to fix the warning.

enricher/standard/src/main/resources/META-INF/fabric8/enricher-default
nicolaferraro: Yeah, it is confusing.. at least, it confused me :)

nicolaferraro:
squashed!

rhuss:
@nicolaferraro awesome, thanks !

rhuss:
retest this please

rhuss:
[merge]

PR #830, opened 2017-02-15
Fixes #829: do not enrich raw resources

### LINKED ISSUE #829: Resources under raw should not be enriched
The same applies to any resource in `src/main/fabric8/<profile-id>/*`. Only enrichers contained in the corresponding profile should be used.

nicolaferraro:


plugin/src/main/java/io/fabric8/maven/plugin/enricher/EnricherManager.java
```diff
@@ -204,7 +204,7 @@ public Void apply(Enricher enricher) {
 
 
     private List<Enricher> filterEnrichers(ProcessorConfig config, List<Enricher> enrichers) {
-        return defaultEnricherConfig.prepareProcessors(enrichers, "enricher");
+        return config.prepareProcessors(enrichers, "enricher");
```
rhuss: gotcha ;-) thanks for spotting this :)

rhuss:
lgtm, thanks !

[merge]

Issue #835, opened 2017-02-19
storage-class annotation is missing from kubernetes PVCs after a recent fmp release


rawlingsj:
### Description

Comparing two recent fabric8-devops yaml files that contain a PVC, it seems the annotation `volume.beta.kubernetes.io/storage-class: standard` has stopped being added.  This means dynamic persistent provisioning will not work on Google Container Engine or AWS so any pod with a PVC is stuck pending.

Examples are 

present in http://central.maven.org/maven2/io/fabric8/devops/apps/fabric8-docker-registry/2.2.311/fabric8-docker-registry-2.2.311-kubernetes.yml which used f-m-p version 3.2.14

missing in http://central.maven.org/maven2/io/fabric8/devops/apps/fabric8-docker-registry/2.2.313/fabric8-docker-registry-2.2.313-kubernetes.yml which used f-m-p version 3.2.18

### Info

* f-m-p version : 3.2.18

To recreate run `mvn clean install` on https://github.com/fabric8io/fabric8-devops/tree/master/fabric8-docker-registry and check the generated `PersistentVolumeClaim` yaml.

rhuss:
I will take care of this.

rawlingsj:
Great - thanks @rhuss

PR #837, opened 2017-02-20
Readd missing storage class (#835)

rhuss:
+ added integration test. The integration test is also prepared for checking that the init-container is missing openshift.yml. This needs to be enabled when the proper fix has been merged in.


rhuss:
Bug fix (missing `endMetadata()` style of error), not much to review.

[merge]

PR #838, opened 2017-02-20
Fixes #799: align documentation

### LINKED ISSUE #799: Introduce new "Watcher" abstraction to replace `fabric8:watch-spring-boot`

nicolaferraro:
Removing the `watch-spring-boot` goal and aligning the documentation. I replaced the inclusion of the `d-m-p` docs with a brief summary linking to it because the usage mode, features and config options don't fit well in the case of `f-m-p`.


doc/src/main/asciidoc/inc/goals/develop/_fabric8-watch.adoc
```diff
@@ -2,8 +2,66 @@
...
 
-include::classpath:/META-INF/doc/dmp/inc/watch/_configuration.adoc[]
+[source, sh]
+----
+mvn fabric8:watch
+----
+
+This plugin supports different watcher providers, enabled automatically if the project satisfies certain conditions.
+
+Watcher providers can also be configured manually. The <<generator-example,Generator example>> is a good bluebprint, simply replace `<generator>` with `<watcher>`. The configuration is structurally identical.
```
rhuss: typo "bluebprint"

rhuss:
[merge]

PR #846, opened 2017-02-21
fixes #845 so we can use an annotation to filter out openshift only or kubernetes only resources for different platforms

### LINKED ISSUE #845: allow a target platform annotation so we can filter resources designed only for certain platforms (e.g. kubernetes only or openshift only)
OpenShift doesn't support `CronJob` yet; but then on OpenShift an alternative to `CronJob` is to use a `BuildConfig`. So it'd be nice to say that `CronJob` was for Kubernetes only and then provide a `BuildConfig` for use on OpenShift.

So it'd be nice to have a little annotation we can use to restrict a resource for use on a specific platform only so we can filter things out for the openshift.yml versus the kubernetes.yml.

A trivial PR is on its way shortly...

jstrachan:


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -317,11 +321,24 @@ private KubernetesList removeOpenShiftObjects(KubernetesList list) {
...
+            return false;
+        }
         return item.getClass().getPackage().getName().contains("openshift");
     }
 
+    private boolean isTargetPlatformOpenShift(HasMetadata item) {
+        // Simple package name based heuristic
+        String targetPlatform = KubernetesHelper.getOrCreateAnnotations(item).get(TARGET_PLATFORM_ANNOTATION);
+        if (targetPlatform != null) {
+            if (!targetPlatform.toLowerCase().equals("openshift")) {
```
pepperbot: MAJOR Merge this if statement with the enclosing one. [rule](https://sonarqube.com/coding_rules#rule_key=squid%3AS1066)
MINOR Move the "openshift" string literal on the left side of this string comparison. [rule](https://sonarqube.com/coding_rules#rule_key=squid%3AS1132)
MINOR Replace these toUpperCase()/toLowerCase() and equals() calls with a single equalsIgnoreCase() call. [rule](https://sonarqube.com/coding_rules#rule_key=squid%3AS1157)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -444,6 +461,9 @@ private KubernetesList convertToOpenShiftResources(KubernetesList resources) thr
         List<HasMetadata> objects = new ArrayList<>();
         if (items != null) {
             for (HasMetadata item : items) {
```
pepperbot: MAJOR Reduce the total number of break and continue statements in this loop to use at most one. [rule](https://sonarqube.com/coding_rules#rule_key=squid%3AS135)

jstrachan:
Agreed; though I'm on the fence of the idea of removing it; as I can imagine we may need it there again when we start combining resources together again into packages. Though maybe we just have to be a bit smarter in how we compose packages better

PR #847, opened 2017-02-21
Respect 'skipBuild' (docker.skip.build)

rhuss:


rhuss:
self-[merge] as it is an obviously missing feature.

Issue #832, opened 2017-02-17
Improve the spring-boot watcher


nicolaferraro:
The spring-boot watcher is awesome, but it should be improved to be used in contexts that deviates from the ideal case...

Here is a list of improvements:
- It should include a version of the `devtools` lib that is compatible with the spring-boot version used by the user (not the one bundled with the plugin)
- It should work even if the user has not deployed an expose-controller in kubernetes/openshift
- It should work even if the user doesn't put a remote secret in application.properties (the code is there, but there's probably a bug)
- It should detect the correct port to be used to communicate with the application and print a specific error message when the port is not present or cannot be used

PR #848, opened 2017-02-21
Get rid of (apparently) unused dependencies

rhuss:
especially the dependency on hibernate validator is annoying as hibernate validator tends to be a beast w.r.t to logging

### DIFF
--- a/watcher/standard/pom.xml
+++ b/watcher/standard/pom.xml
@@ -60,22 +60,6 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-devtools</artifactId>
     </dependency>
-    <dependency>
-
-      <groupId>org.springframework</groupId>
-      <artifactId>spring-web</artifactId>
-    </dependency>
-
-    <dependency>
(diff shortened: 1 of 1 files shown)


rhuss:
@nicolaferraro do you know whether we really need this (now removed) dependencies ? Its not actively used in the code, but maybe because of lookup of devtools.jar ? sorry not that deep in the code.

The problem is the hibernate-validator dep which is super noisy sometimes.

nicolaferraro:
@rhuss let me do a check. Spring-boot sometimes refuses to start if you don't provide an implementation of JSR-303, but I don't see why it should stay in the plugin.

nicolaferraro:
lgtm

The spring-boot remote watcher client definitely does not need `hibernate-validator` and libraries declared in the plugin are not included in the deployed artifact (exception made for `devtools.jar`). So they can be removed. We should also get rid of the `devtools` lib completely (#832).

rhuss:
thanks for the cross check ! btw, agreed with everything you mention in #832

rhuss:
[merge]

PR #852, opened 2017-02-23
Removed order ids from enricher/generator definitions

rhuss:
Removed order ids from enricher/generator definitions since the order are defined in the profile used (which is defined in "profiles-default.yml".

Also added some comments to clarify stuff a bit.


rhuss:
[merge]

Issue #206, opened 2016-08-02
when using ${project.artifactId} for the service/DC/RC metadata.name we should trim to 24 characters


jstrachan:
if an archetypeId is too long we should automatically trim it to avoid creating an invalid kubernetes/openshift set of resources

jstrachan:
we might want to also convert to lowercase and remove any invalid characters too

rhuss:
We already have [the truncation](https://github.com/fabric8io/fabric8-maven-plugin/blob/bdcdb32329117f592ae8c63e4fa3855c9c246eed/core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java#L99-L101) when creating the default resource names. 

We can add even more name mangling here. Are there any other spots where this should be done ?

jstrachan:
so in the new [arquillian system tests for the quickstarts](https://github.com/fabric8io/ipaas-quickstarts/tree/master/archetype-itests) I was getting failures - do to Service metadata.name being too big for a few of the quickstarts. I worked around it for now by manually truncating the artifactId:
https://github.com/fabric8io/ipaas-quickstarts/blob/master/archetype-itests/src/test/java/io/fabric8/tooling/archetype/generator/ArchetypeTest.java#L229-L230

I suspect the failure is probably that we don't truncate the yaml files such as this: 
https://github.com/fabric8-quickstarts/cdi-cxf/blob/master/src/main/fabric8/service.yml

i.e. we maybe wanna check we're doing the truncation when we post process partial yaml files

astefanutti:
Note that service names up to 63 characters will be allowed in Kubernetes v1.4 (from RFC-952 to RFC-1035) as per kubernetes/kubernetes#29523.

So we should ideally adjust to that new limit at some point and this will come handy with longer names like quickstart names.

PR #857, opened 2017-02-27
increasing the service name length to 63 to reflect the updated limits in kubernetes v1.4.0

pldimitrov:
#206 

https://github.com/kubernetes/kubernetes/issues/12463
https://github.com/kubernetes/kubernetes/pull/25041
https://github.com/kubernetes/kubernetes/pull/28005


pldimitrov:
Hi there, we were having issues with the plugin truncating service names and I noticed this was fixed in Kubernetes v1.4.0-alpha.3 (https://github.com/kubernetes/kubernetes/pull/29523) so it seems like there is no reason to keep that limited to 24 characters anymore.

rhuss:
Thanks ! looks good to me

[merge]

Issue #526, opened 2016-10-10
WebappGenerator should ignore WildFly-Swarm apps with WAR packaging 


jamesnetherton:
I hacked together a WildFly-Swarm app via the [generator](http://wildfly-swarm.io/generator/) site and added the f-m-p to my POM. Seems the default packaging for these generated projects is WAR - which causes the WebappGenerator to kick in. 

Given that we're going to produce a Swarm fat-jar, the WAR is mostly redundant. So we should avoid doing anything with the WebappGenerator and allow the WF Swarm generator to take over.

PR #858, opened 2017-03-02
fixes #720 so that spring boot / wildfly swarm apps work when using war packaging

### LINKED ISSUE #720: WildFlySwarm - The generator should detect .war package
A wildfly swarm app may be packed as a .war also. Although I guess fat jar is recommended/better.

But this example was using .war
https://github.com/davsclaus/fabric8-hello/blob/master/helloswarm/pom.xml#L8

And then it was the default .war generator that was kicked in and wanted to run it as a tomcat.

We should detect that wildfly-swarm plugin is in the pom
https://github.com/davsclaus/fabric8-hello/blob/master/helloswarm/pom.xml#L35

And let the wildfly-swarm generator kick in instead.

jstrachan:


generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -68,7 +73,30 @@ public WebAppGenerator(GeneratorContext context) {
...
         return shouldAddImageConfiguration(configs) &&
-               MavenUtil.hasPlugin(getProject(), "org.apache.maven.plugins:maven-war-plugin");
+               MavenUtil.hasPlugin(getProject(), "org.apache.maven.plugins:maven-war-plugin") && !isWarUsedWithFatJar(configs);
+    }
+
+    /**
+     * If we are using spring boot or wildfly swarm with a fat jar then we should not use the web app generator
+     * and should stick with the fat jar generators instead
+     */
+    protected boolean isWarUsedWithFatJar(List<ImageConfiguration> configs) {
```
pepperbot: MAJOR Remove this unused method parameter "configs". [rule](https://sonarqube.com/coding_rules#rule_key=squid%3AS1172)

core/src/main/java/io/fabric8/maven/core/util/Constants.java
```diff
@@ -22,4 +22,6 @@
 public class Constants {
     public static final String RESOURCE_SOURCE_URL_ANNOTATION = "maven.fabric8.io/source-url";
     public static final String RESOURCE_APP_CATALOG_ANNOTATION = "maven.fabric8.io/app-catalog";
+    public static final String SPRING_BOOT_MAVEN_PLUGIN_GA = "org.springframework.boot:spring-boot-maven-plugin";
```
rhuss: Don't think its a good idea to add tech specific constants in the core package. said that several times, but I guess I'm wrong somehow. would be happy to be enlighted (btw, we are currently on the way to exactly do the reverse, to decouple stuff. See the e.g. `fabric8:watch` refactorings, where we move out spring specific stuff from the core.

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -68,7 +73,30 @@ public WebAppGenerator(GeneratorContext context) {
...
+        MavenProject project = getContext().getProject();
+        String finalName = project.getBuild().getFinalName();
+        String artifactId = project.getArtifactId();
+        if (!Objects.equal(artifactId, finalName)) {
+            String plugin = null;
+            if (MavenUtil.hasPlugin(project, SPRING_BOOT_MAVEN_PLUGIN_GA)) {
```
rhuss: Don't think its a good solution to let this plugin know about other orthogonal techs. Better to use ordering of the plugins, e.g. to put the war plugin behind spring-boot and wildfly-swarm ? Actually I don't understand why this should be necessary as the check to `shouldAddImageConfiguration` only kicks in when there has not yet already been added a image configuration by string-boot or wildfly-swarm (so the first checks fails if this is the case)

jstrachan: Here's one example of the issue: https://github.com/fabric8io/fabric8-maven-plugin/issues/720 basically any WAR projects using WildFly Swarm barf today on OpenShift.

More than happy to experiment with other solutions though.

If spring boot + wildfly swarm plugins were head in the priority could they effectively disable the webapp generator?

core/src/main/java/io/fabric8/maven/core/util/Constants.java
jstrachan: was just trying to avoid copying and pasting strings all over the place. Happy to revert if your re-ordering idea turns out to work

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
jstrachan: Incidentally it looks like previous attempts to make webapp not break spring boot / swarm apps failed too: https://github.com/fabric8io/fabric8-maven-plugin/issues/526

if they are higher priority and can effectively disable the webapp plugin that sounds fine too

rhuss: That's what I thought, too (except it the option [add](https://github.com/rhuss/fabric8-maven-plugin/blob/20cb6e08917a644b5c128d1d5489624cd71e5cbf/generator/api/src/main/java/io/fabric8/maven/generator/api/support/BaseGenerator.java#L61-L61) is set to `true` in which case generator generator configs as accumulated. otherwise only the first wins). If you give me a bit of time until, I'm happy to check this as the first thing in the morning tomorrow.

jstrachan: Awesome. I'm more than happy to move the 'disable webapp generator' logic into spring boot / wildfly swarm plugins if that works; its a little cleaner then. Folks could then disable, say, the spring-boot generator and the webapp generator would kick back in etc. 

We should be sure to document why spring boot / wildfly swarm plugins have higher priority to avoid folks accidentally breaking this later.

PR #861, opened 2017-03-06
Revert "fixes #720 so that WAR projects using spring boot or wildfly swarm plugins stick with the fat jar generators rather than barf on openshift with the war generator"

### LINKED ISSUE #720: WildFlySwarm - The generator should detect .war package

rhuss:
This reverts commit 9023a297a0a79897e3a09ae3d652f70e6a4524ad.

This is related to #720 which needs a different fix.


rhuss:
The ordering is fixed with #873 so that the original fix is not needed anymore.

rhuss:
[merge]

PR #863, opened 2017-03-06
Do not generate a default selector for jobs

pldimitrov:
`.spec.selector` is optional for jobs. There is no need to generate it by default.


rhuss:
Thanks ! Looks good to me.

@pldimitrov btw, the email associated with the commit is different than your github account email. np, for now, just wanna let you know (as GitHub complain a bit about this with a question mark after your name)

[merge]

PR #873, opened 2017-03-13
Fixed order of generators when merging configs and profiles

### LINKED ISSUE #862: generator: <include> in configuration changes order of generators
If you include the generator in a `<configuration><generator><include>` then this generator is moved to the end of the list of generators coming from a profile. 

Actually, this configuration should only include the referenced generators and not the one from the profile. At least it should not change the order.

This is related to #720 and the referenced JIRA issue within it.

rhuss:
Revised the merging part of profiles and configuration.

Now, when a generator is specified in <configuration> it is moved to the *beginning* of the list of generators, which makes sense since normally the first genertor kicking wins.

For enricher it's not so clear, but now enrichers are also moved to the beginning to the enricher list for consistencies sake.

For fined tuned enricher/generator ordering handling it is recommend to use an explicit (possibly handcrafted) profile where the order can be fine tuned.

Fixes #862

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
+++ b/core/src/main/java/io/fabric8/maven/core/config/ProcessorConfig.java
@@ -112,11 +112,26 @@ public class ProcessorConfig {
     }
 
     /**
-     * Merge in another processor configuration, with an higher priority. This operation mutates
-     * the current config.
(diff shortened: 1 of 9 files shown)


rhuss:
[merge]ing as its a crucial fix for the proper generator ordering in #720

Issue #878, opened 2017-03-14
lets migrate to centos rather than alpine base images OOTB


jstrachan:
as we've seen issues with DNS and alpine issues; seems centos is more stable

rhuss:
That has already be done.

PR #877, opened 2017-03-14
Reworked DefaultServiceEnricher

rhuss:
- Refactored to have simpler methods which are easier to understand
- Added documentation
- legacyPortMapping: config flag to switch on / off the legacy mapping from 8080 / 9090 --> 80. Switched on by default for backwards compatibility. This addresses issue #690.
- port: Config option for explicitly setting a service port to map to. Works only when a single port is exposed. Maybe this should be make more flexible to allow multiple ports ?

PR #879, opened 2017-03-14
switch from alpine to centos images #878

jstrachan:


rhuss:
Sounds good to me. Maybe we should combine this in a new 3.3.0 release together with #877 (as I consider this to be a major change) ?

jstrachan:
@rhuss sounds good to me

rhuss:
@jstrachan jus tto be sure that the size implications are know: alpine == 99 MB, centos == 483 MB. Of course its cache but need to be reloaded for each new installation (minishift, etc.) We all know the discussion about size etc., don't want to waste any energy here now.

just wondering whether there are any other opinions on a default base image switch for non-s2i builds // @fabric8io/fabric8-admin

jstrachan:
Yeah. The s2i images are based on jboss images which are centos based too. When folks start using other things like jenkins or whatever those are increasingly centos based as well; so am not sure folks are gonna be able to do much on minikube/minishift without downloading a base centos image am afraid.

Its a shame we can't make things tiny OOTB; but having things work (DNS is kinda handy!) seems more important than caching a centos base layer unfortunately. (Would love it if minikube/minishift could reuse the cached docker layers between VMs to speed things up ;)

rhuss:
If we want to align with the s2i bits, maybe its better to switch to `fabric8/java-jboss-openjdk8-jdk` which uses the same base image `jboss/base-jdk:8` as the s2i image ?

jstrachan:
@rhuss sounds good to me

rhuss:
ok, just pushed an update.

rhuss:
looks good to me

[merge]

PR #895, opened 2017-03-22
overloading clusterAccess.createDefaultClient to take platform mode into account

pldimitrov:
Removing unnecessary `isOpenShift` check from `Fabric8ServiceHub` constructor by using the new method - this results in a much faster build time in situations when platform mode is not set to `PlatformMode.auto`. 

In this particular case the open shift check would have been done by the call to `clusterAccess.resolvePlatformMode` anyway, if necessary.

Other spots where this can be potentially done:
- BuildMojo (the mode is already exposed here via ` @Parameter(property = "fabric8.mode")`)
- ImportMojo
- WatchMojo


rhuss:
Thanks and sorry for the delay ! Will have a review asap (at the latest when I'm back home next week)

rhuss:
Looks good and makes sense. As it looks like, the code you introduced is not really used. While digging a bit deeper, the code around the client selection is really a big ball of mud, with a lot of unnecessary API calls. Le me track that issue and let's come back later to clean this up in a larger context. 

I opened #1342 to not forget about this.

PR #917, opened 2017-04-19
Allow configurable timeoutSeconds for SpringBootHealthCheckEnricher

Cameronjmayfield:
In attempting to configure the health check timeout for our Spring Boot app, we did not find a configuration option that allows us to do it. This modification allows a user to configure the timeout.

### DIFF
--- a/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/SpringBootHealthCheckEnricher.java
+++ b/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/SpringBootHealthCheckEnricher.java
@@ -29,6 +29,7 @@ import io.fabric8.maven.enricher.api.EnricherContext;
 import io.fabric8.utils.PropertiesHelper;
 import io.fabric8.utils.Strings;
 
+
 /**
  * Enriches spring-boot containers with health checks if the actuator module is present.
  */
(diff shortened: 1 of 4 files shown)


davsclaus:
lgtm, @rhuss can we get this merged?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/SpringBootHealthCheckEnricher.java
```diff
@@ -42,6 +44,12 @@
     private static final String SCHEME_HTTPS = "HTTPS";
     private static final String SCHEME_HTTP = "HTTP";
 
+    private enum Config implements Configs.Key {
+        timeoutSeconds {{ d = "1"; }};
```
rhuss: Could we omit a default value and, if timeoutSeconds is not configured, simply don't add it to the ProbeBuilder ? That way we ensure that we will always take the default of Kubernetes and do not sync the 1s here.

davsclaus: Ah yeah using what is the default in k8s would be better. 

Also I think this functionality about the timeout is needed in the other health checkers for wildfly-swarm, vertx etc. But that can be done in another PR

rafaeltuelho:
what is the current approach recommended to change the default timeout until this PR be merged? Maybe using a fraction on `src/main/fabric8/deployment.yaml`?

PR #920, opened 2017-04-24
fixes #919 so we preserve the DC strategy type on a Deployment

### LINKED ISSUE #919: Deployment.spec.strategy.type seems to get lost from src/main/fabric8/deployment.yml
it looks like if we try to specify `spec.strategy.type` in a custom `src/main/fabric8/deployment.yml` file it gets lost/ignored by the merge of the generated and custom Deployment YAML.

e.g. a yaml file like this does not seem to have an effect

```
spec:
  replicas: 1
  strategy:
    type: Recreate
```

jstrachan:


rhuss:
[merge]

PR #944, opened 2017-05-17
Increase DeploymenConfigTimeout to de-facto infinity to compensate long Docker image download times

### LINKED ISSUE #943: ImageStream triggers break DeploymentConfigs on OpenShift
it seems moving to 3.4.0 of fmp adds mandatory `ImageStreamChange` triggers OOTB to `DeploymentConfig` which breaks them massively on OpenShift:

* using `oc apply`, `mvn fabric8:deploy` or `gofabric8 deploy` results in nothing being deployed at all. You have to then manually go into each DC and trigger it (e.g. via the `Deploy` button in the OpenShift console
* `oc edit dc foo` no longer updates the DC at all ;)

i.e. the DCs become pretty much utterly useless! :) 

So adding `ImageStreamChange` triggers OOTB sounds kinda scary. So lets at least make them optional for now (maybe defaulting to disabled to reduce breaking things?)

I wonder what the logic should be to ever include them since it seems to break things. Maybe we only enable them if we know that we are building the container image itself via an image stream? In my use case (the upstream fabric8 project) the images are public docker hub images and so are not built via ImageStreams

jstrachan:


plugin/src/main/java/io/fabric8/maven/plugin/converter/DeploymentOpenShiftConverter.java
```diff
@@ -99,7 +101,7 @@ public HasMetadata convert(HasMetadata item) {
                 specBuilder.addNewTrigger().withType("ConfigChange").endTrigger();
 
                 // add a new image change trigger for the build stream
-                if (containerToImageMap.size() != 0) {
+                if (containerToImageMap.size() != 0 && useImageStreamTriggers) {
```
rhuss: @jstrachan actually do you know _why_ image stream triggers break the deployment ? What are the guideline to give user when to switch in image stream triggers and when not ? And can this be determined automatically ?

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -174,10 +174,16 @@
      * The OpenShift deploy timeout in seconds:
      * See this issue for background of why for end users on slow wifi on their laptops
      * DeploymentConfigs usually barf: https://github.com/openshift/origin/issues/10531
+     *
+     * Lets default to 2 hours which should be plenty of time on a slow connection to download
+     * all the images for a YAML/JSON manifest (which can contain many separate images)
      */
-    @Parameter(property = "fabric8.openshift.deployTimeoutSeconds")
+    @Parameter(property = "fabric8.openshift.deployTimeoutSeconds", defaultValue = "7200")
```
rhuss: AFAIR we already had this timeout but removed it for some [reason](https://github.com/jstrachan/fabric8-maven-plugin/commit/b526857a184f79fb7d4f660c84cbb20c0b965dd1). Is this related to this PR ?

jstrachan: I did mean to put that in a separate PR. The main issue is, if you try run a YAML/JSON with, say, 5 images then on my wifi the DCs fail to deploy because the images don't download in time. Deployments don't have this issue - there's no timeout for downloading the images - like there is for DCs. Having DCs that don't deploy is kinda a massive UX fail really.

So to make D and DCs work the same I figure we should give folks more than 10 minutes to download all possible images relating to a YAML / JSON and start them. 

I guess if you're not gonna buy my arguments about making DCs behave more like Ds OOTB; we can always just add this property to every pom.xml in every fabric8 repository ;) - but having DCs that typically don't work at all on slow-ish networks sounds kinda wrong to me. I guess a better long term fix is to get openshift not to fail a DC if its still downloading the image

plugin/src/main/java/io/fabric8/maven/plugin/converter/DeploymentOpenShiftConverter.java
jstrachan: I've no clue when an image stream trigger even makes sense - all I know is they break openshift horribly ;)

jstrachan: until we figure out when they have any actual use, I'd prefer we disable them ;)

jstrachan: my guess would be - to only use them when we've created the image via an ImageStream? I wasn't sure how to reliably determine a container image was created via an ImageStream though or I'd have added that to the PR

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
rhuss: If I remember our discussions at that time (along with @jimmidyson and other) we had severe issues with ever hanging deployments which are not due to a slow wifi but because of an real error. 

So this is really a trade off. 

As it happens on deployment time, there is no need to add everything on the pom.xm but if you are on a slow wifi connection you always can add the timeout property on the command line. Tbh, I wouldn't try to be cleverer than OpenShift itself here and rely on the standard OpenShift timeouts. Or asked otherwise: How does `oc apply` reacts in this situations ?

And then, how often happens that you are on a slow wifi (probably mostly only for evangelists travelling around) vs. those who have decent connections (which the would be annoyed because of endlessly hanging deployments).

As I said, that's a trade off and we already had this discussion (and a decision). But we can do it again (wonder though what has changed since that time).

jstrachan: to give a concrete example - `gofabric8 deploy` has never worked for me for fabric8 with DCs with the OOTB 10 minute timeout

jstrachan: I'm still confused by the 'long timeouts cause hanging deployments' comment. Deployments on kubernetes don't have a timeout at all right? Things can fail; sure, but kubernetes never stops deploying

jstrachan: I a short timeout can help developers when they are developing the app; if you know the image is local you may wanna go shorter - to 30 seconds or something. Its very easy to enable this setting in your `~/.m2/settings.xml`

The issue with the default OOTB is when you make an app as an openshift YAML / Template it means you've no idea if the end user can ever even run your app without manually retrying deployments as it totally depends on their network speed - which is a kinda crappy UX if you're trying to make reusable apps for end users on openshift. Especially if your app, like fabric8 itself, has like 7-10 DCs inside which are likely to fail for most folks who don't have crazy fast wifi.

rhuss: Now I remember a bit better: I started a discussion at https://github.com/fabric8io/fabric8-maven-plugin/pull/794 (would have loved to discuss this at this time, but it didn't take off) and the issue we had, that, when a pod fails to start for some reason and the DC retries, it still didn't break the deployment and only after the three hours comes back. Such a large timeout like 3h can only be a patch on a larger issue. Also, this blocking operation does not provide any progress feedback (which it could as it is blocking), so one never knows when to press CTRL-C. Such a large timeout is really a black box.

I agree with you having this as a synchronous operation without progress meters in OpenShift is a PITA. But  a long blocking timeout doesn't solve this really (only for particular use cases and you can configure for this particular use cases instead of making this the default)

BTW, why does 'gofabric8 deploy' calls the fabric8-maven-plugin so that it needs a larger timeout ? Didn't get this part.

How often do you deploy 7-10 DCs with a single Maven build (where you don't have also build the images in the cluster) ? if it is a rare use case (e.g. for fabric8 only), you would need only to set the timeout on this monster deployment then.

@jimmidyson any opinions on a 2 hour deployment timeout ?

jstrachan: It doesn't matter whether you use `mvn fabric8:deploy`, `oc apply`, `gofabric8 deploy` or instantiate a Template via the openshift console - in all cases openshift uses the timeout from the YAML to determine how long all the `DeploymentConfig`s have before they all load the docker images + startup. Using the OOTB 10 minutes timeout means I'm unable to ever deploy anything on minishift with > 3 or 4 images inside. 

I totally understand that this is really openshift's issue; it should differentiate between time taken to download images (which when running a YAML with many images inside can take a while for them all to download) versus time taken to really startup the pods.

But the question is faced with the various openshift releases out there right now (OSCP/OSD/OSO/minishift) what should app developersput in their released YAMLs to make them as useful as they can for end users consuming them via oc/openshift console/gofabric8/mvn.

rhuss: I see your point, my point is that is benefits is converted to the contrary when an error happens in the pod failing it to startup (especially when used in a CI environment where no one will press CTRL-C and one has to wait for the time, except the CI, sets also a shorter timeout). 

Another idea: Could we make the default timeout dynamically based on the number of images inside?

jstrachan: I guess CI builds could easily override this timeout setting via `~/.m2/settings.xml` too - particularly as the CI job knows its already built the image so should have very fast download times? 

Unfortunately we use composition across maven modules to take individual templates and compose them into larger apps. I guess we could post process and keep increasing the timeouts for each app whenever it gets composed; but was hoping we could just pick a less-likely-to-fail default. After all folks could run 5 templates in parallel too to replicate the failure.

rhuss: I find it really unfortunate this timeout value is part of the `DeploymentConfig` so that we need to mangle with this parameter on given resource objects (which can come from dependencies, too). 

Following the upstream issues https://github.com/kubernetes/kubernetes/issues/3074, https://github.com/openshift/origin/issues/10531 and https://github.com/openshift/origin/issues/7522 it seems that there is some (but slow) progress in separating image download times and the other deployment startup. 

I wonder whether the referenced `ProgressDeadlineSconds` API can help here (have to look it up) or whether we should switch to plain `Deployment` as soon as they land in OpenShift (as it is supposed to be planned for a future version).

I'm d'accord with increasing the timeout (maybe 1h is enough for now ?) considering to remove the default at any time when finding a better support from the upstream APIs.

jstrachan: Yeah - am hoping whatever we do will be a short term thing until its properly fixed. 

I guess for cases we're not building images with ImageStreams it might well just be better to go with `Deployment` anyway on OpenShift as I see DC`s as being much worse ;) The only real value of DCs seems to be when using ImageStreams really from what I can tell

rhuss: Agreed. Do you know when `Deployments` will be supported on OpenShift ? (BTW wonder also how we should deal with backwards compatibility with people using older OpenShift installations).

I suggest to continue from here on with:

* increasing the timeout (in a separate PR)
* me checking how to avoid the image trigger for non-s2i scenarios and how to refactor the resource generation as described in #678

rhuss: Another question I have: Why has a single DeploymentConfig more than 3 images as a common case ? I thought the timeout is individual for each DeploymentConfig, not for a whole template with many DeploymentConfigs, so the timeouts add up anyway within a template ?

rhuss: If this is the case (i.e. individual timeout for each DCs), maybe it is sufficient to increase to a more moderate timeout like 20 minutes (doubling the default ?). What was the minimum timeout that worked for you? (still looking for a compromise ;)

jstrachan: Its not that 1 DC has 3 images; its that a composite app (e.g. `gofabric8 deploy` has about 7 DCs which all have 1 image (with maybe an init-container here or there too)

jstrachan: As we've said in this thread; if folks don't like the long timeout in development or CI/CD pipelines its very trivial to override on the CLI or in `~/.m2/settings.xml` so I'd rather have DCs work for more users who may not have fast wifi really - folks in dev / CI / CD will probably wanna use a shorter timeout whatever we pick anyway

jstrachan: 1 hour seems reasonable? It used to take me about that time to install fabric8 for example

rhuss: still don't get it: As the timeout is part of a DC, I suppose it applies only to this very DC, not to the sum of all DCs. So the number of DCs deployed doesn't matter.

The same argument goes for 'you can overwrite if required' can be applied in the other direction as well (when you need a higher timeout, why not simply configure it ?). It's simply about which use case is more likely (with such a high timeout everybody doesn't get a feedback early when a deployment fails for another reason than long image download times, and as said, wonder why we should be cleverer than the OpenSift guys and don't rely on the OpenShift timeout ?)

It's really not about nit-picking, but I really want to understand the issue better (i.e. why the timeout should be applied to all DCs together and not individually). 

Having 1 hour for downloading image feels a bit much for me, even for slow wifi. About what number are we talking ? E.g. for an 2 MBit WiFi and a 500 MB image, this would take 33 mins (if everything needs to be fetched from scratch). 

Within 1 hour you can download a 900 MB image (2 MBit) or 500 MByte with a 1.1 MBit speed. Is this it the usual user then I'm fine with the timeout. But even then I still doubt that the user is willing to wait 1 hour without any progress indicator. More likely he will kill the process before (and since the download continues in the background, the next time it's more likely to succeed faster.

Said that I will add the 1 hour timeout now. Let's wait on user's feedback, as this is the only way to find out the real thing.

jstrachan: if you run `oc apply` or `gofabric8 deploy` or whatever with, say, 7 DCs then they all get created at the same time yet all share the same network bandwidth to download the images; so its pretty much guaranteed that some of those DCs will fail in 10 minutes from the timeouts. i.e. the more DCs you create a the same time, the longer its going to take to download all those images and the more likely a DC will just refuse to deploy.

I see overwriting the timeout value for development / CI / CD a developer thing. We want YAMLs to just work for users whether they use `oc apply`, run via the openshfit console or via `gofabric8` - as its pretty hard for end users to even know about this timeout never mind manually swizzle the templates/yamls to tweak these values. Whereas setting 1 mvn property for development / CI / CD is trivial for developers to do

rhuss:
ok, understood the parallel download issue (thought that DCs would be applied in sequence). 

wrt to developer experience and since the initial deploy is done via gofabric8, wouldn't  it make sense to add an option to gofabric8 which let's tune the timeout within the downloaded YAML files before applying to the cluster (instead of relying on what is defined in the YAMLs) ?

BTW, I consider this also a good example for an environment specific datum distributed as context-independent resource descriptors within jar artefacts (e.g. like imagestreams which also should not be distributed as part of the artefacts). IMO the tool which applies the resource descriptors knows much better about the target cluster and hence should contain an appropriate option to set the timeout. 

In the context of f-m-p, I would that timeout value to `fabric8:apply`, too (which then would overwrite these values in the DCs on the fly).

jstrachan:
there's no 'in sequence' really, kubernetes/openshift applies all resources at the same point in time (and so all the timers start at the same time); there's no way to tell kubernetes/openshift do apply then one at a time or anything. 

So there are many ways to apply a YAML in so many areas; we kinda want to minimise requiring special tools to process them in magic ways. e.g. we want minishift to be able to take standard YAML files and apply them on startup; or have YAML usable in the openshift console, via `oc apply` or via `gofabric8`. So while its not awesome, adding a decent timeout seems like the cleanest, most flexible work around until openshift's DCs are properly fixed to ignore image download time (or have a separate long timeout for that part).

rhuss:
@rohanKanojia re-reading the issue again, I remember darkly the heated discussion we had at that time. My opinion hasn't changed much, so I wonder whether a too short timeout for DeploymentConfigs are still an issue ? 

My POV is, that the timeout should be configurable, but we are relying on the OpenShift default as default. So when someone requires a large timeout she still can configure it, but I suspect the typical use case doesnt require a monster timeout (e.g. in our project where we use intensively DCs for the deploymen of Syndesis, we didn't have any issue).

So I would opt for closing this PR.

rohanKanojia:
@rhuss : I see that intention of this PR was to increase DeploymentConfig timeout due to some issues on Kubernetes itself, but since kubernetes/kubernetes/issues/3074 , openshift/origin/issues/10531 and openshift/origin/issues/7522 are fixed now. I don't think we might have problems now. I think we can close this.

@jstrachan I know you're very busy, but could you please share your thoughts whenever you have time? Do you still have that use case where setting this big timeout could still be helpful?

rhuss:
Cool. Let me close the PR and reopen in case we need to elaborate on it.

PR #881, opened 2017-03-15
Add profile support in spring-boot

kameshsampath:
* Support for adding Spring Boot profiles as JAVA_OPTIONS when the user provides configuration via `activeProfiles`

* ability to handle multiple application yaml docs with/without spring active profiles 

This pull resolves #797 and avoids #880 which can now be customised via profile during build

PR #955, opened 2017-05-30
Spring boot/issue 880

kameshsampath:
moving to new PR for #881


core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
```diff
@@ -40,16 +38,39 @@
...
+
+        //If the profiles are available load the profile resources as well
+        if (activeProfiles != null) {
+            for (String profile : activeProfiles) {
+                ymlResource = getResourceFromClasspath(compileClassLoader, "application-" + profile + ".yml");
+                Properties props2 = getPropertiesFromYamlResource(ymlResource, activeProfiles);
+                propertiesResource = getResourceFromClasspath(compileClassLoader,
+                        "application-" + profile + ".properties");
+                props2.putAll(getPropertiesResource(propertiesResource));
+                props.putAll(props2);
```
rhuss: You could put all of this logic into a method (from line 66 to line 71) and then call it for the case above the if, too (which has the same logic but without a profile name).

I.e. I would add a `addApplicationProperties(props, compileClassLoader, profile)` with `profile == null` means no profile.

core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
```diff
@@ -80,25 +101,82 @@ protected static Properties getPropertiesResource(URL resource) {
...
+                    if (docRoot.containsKey("spring")) {
+
+                        LinkedHashMap value = (LinkedHashMap) docRoot.get("spring");
+
+                        Object profilesValue = value.get("profiles");
+
+                        if (profilesValue instanceof Map) {
+                            Map profileMap = (Map) profilesValue;
+                            String[] arrayOfActiveProfiles =
+                                    StringUtils.split((String) profileMap.get("active"), ",");
```
rhuss: I would use the same split as below with a regexp here, so you can skip the trim iteration (i.e. `activeProfiles.split("\\s*,\\s*")` after checking that the map contains an active profile.

rhuss: Actually why not reuse method `getActiveProfiles()` here ?

core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
```diff
@@ -80,25 +101,82 @@ protected static Properties getPropertiesResource(URL resource) {
...
+                        }
+                    }
+
+                    if (profiles != null) {
+                        String[] profileSplit = profiles.split("\\s*,\\s*");
+                        if (!CollectionUtils.
+                                intersection(Arrays.asList(profileSplit), activeProfiles)
+                                .isEmpty()) {
+                            //if the profiles is in the list of active profiles we add it to our list of docs
+                            profileDocs.put(profiles, docRoot);
```
rhuss: Don't you have to add an entry for **every** profile ? (not only a single entry for the comma separated list, which is not even normalized)

core/src/test/resources/util/test-application-multi.yml
```diff
@@ -0,0 +1,29 @@
+spring:
+  profiles:
+    active: dev,qa
+
+---
+
+spring:
+  profiles: dev
```
rhuss: would be good to have a test with multiple, comma separated profiles, too (as I suspect this is currently broken).

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -63,9 +52,18 @@
     private static final String DEFAULT_SERVER_PORT = "8080";
 
     public enum Config implements Configs.Key {
-        color {{ d = "false"; }};
+        color {{
```
rhuss: could we keep the formatting here, as it looks a bit more concise then.

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -63,9 +52,18 @@
...
+            d = "false";
+        }},
+
+        // comma separated list of spring boot profile(s) that would be passed set as -Dspring.profiles.active
+        activeProfiles;
 
-        public String def() { return d; } protected String d;
+        public String def() {
```
rhuss: same here wrt formatting.

core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
kameshsampath: ok see what you mean

core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
kameshsampath: yeah true makes sense, modified the code

core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
kameshsampath: no thats taken care by previous block, this in case where the yaml has only one doc so ideally we will have only one profile.. i hope my understanding is right

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
kameshsampath: sorry it was my bad when i did formatting IDEA split it as we.. wondering is there way to configure this style ;)

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
kameshsampath: done!

core/src/test/resources/util/test-application-multi.yml
kameshsampath: not sure how this will come ??? you mean something like:

spring:
  profiles: dev, qa 

???

kameshsampath: @rhuss - i dont think spring.profiles can have multiple values https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-profiles.html , so we dont need to worry on it . Only spring.profiles.active can have multiple values via comma seperation

rhuss: yes, something like this as shown in https://github.com/spring-projects/spring-boot/blob/master/spring-boot-samples/spring-boot-sample-profile/src/main/resources/application.yml 

Also, if it could not have multiple values, why did you split it then in the code above ? Also I believe it could be even a YAML list (same like for activeProfiles)

core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
```diff
@@ -80,25 +93,79 @@ protected static Properties getPropertiesResource(URL resource) {
...
+                                    activeProfiles.addAll(getActiveProfiles((String) profileMap.get("active")));
+                                }
+                            }
+                        } else if (profilesValue instanceof String) {
+                            profiles = (String) profilesValue;
+                        }
+                    }
+
+                    if (profiles != null) {
+                        String[] profileSplit = profiles.split("\\s*,\\s*");
```
rhuss: If profiles can take multiple values, why do you split it here then ? And couldn't it be that `profiles` could be also a list ?

rhuss: Afaiu `activeProfiles` is a  cascading mechanism where the profiles later in the list override the former.

rhuss: What's about `spring.profiles.include` ?

kameshsampath:
@ro14nd - made some logical changes in SpringBootUtil and added test case for include profile as well. I have kept the old logic still in tact, just wanted your thoughts before I clean them up. if you are ok with all logic + additional comments if any, will fix it and clean up unused methods then.

rhuss:
Thanks ! I hope I can do a review today, if not, the next slot is probably only Friday next week ;-(

cmoulliard:
@rhuss Should we rebase this PR to make it work ?

nicolaferraro:
@kameshsampath I've not followed all the evolution of this issue but it seems you've added spring-boot dependency in the last solution. We've made many changes to this plugin in order to get rid of the dependency on spring-boot from the core (it brings also some spring stuff as transitive dependencies).
Ideally we should replicate the same logic used by spring-boot, without importing spring-boot (unless they separate the logic into a tiny module).

Is it possible to restore the old logic, and fix the problems occurring with profiles? (probably with multiple profiles)

rhuss:
I rebased this PR and reverted commit https://github.com/fabric8io/fabric8-maven-plugin/pull/955/commits/e27270bd7e2b353e46c3efb7e9f67c0531c0e109 which introduced the spring-boot dependency. 

@kameshsampath is this PR still good to merge (even when not used with a spring-boot dep for parsing the files ?). If so, I will try to make a review this week.

And yes, sorry for the looooooong delay ;-)

kameshsampath:
@rhuss am sorry its been really long time, need to check it why I added the spring dep, am not sure I can do this over next few weeks, but will try to take a peep whenever I get time. sorry about it.

rhuss:
@kameshsampath no problem. It was my fault anyway to not finish the review. Let me make another review, and maybe I can polish the PR myself. Lets see ...

kameshsampath:
none's fault @rhuss .. just our timelines/bandwidth just did not synch

Issue #941, opened 2017-05-16
Add Kompose support


rhuss:
[Kompose](http://kompose.io/) is an awesome conversion tool which creates Kubernetes descriptor objects from [Docker compose](https://docs.docker.com/compose/) files.

The idea is to extend the `fabric8:resource` goal to support Docker compose files via an external CLI call to `kompose`. This should be possible to be switched on either explicitly by a configuration option `<composeFile>` in the XML configuration or maybe even implicitly by monitoring a directory `src/main/fabric8-kompose` for the content of some `docker-compose.yml` files.

The flow would be:

* If enabled and present, `kompose` is called externally which creates Kubernetes resource descriptor yaml files in a build directory below `target/fabric8`. Configuration for `kompose` comes from the plugin configuration.
* These descriptors are read in by f-m-p via the kubernetes client
* The so instantiated `KubernetesList` is then used as starting point for the further processing:
  - Additional fragments can be added
  - Information of the XML configuration can add to this list
  - The enricher chain is fed with this list so that it can be enriched as usual.

An alternative approach would be to re-implement kompose functionality in Java which would provide a better integration (no external tool is required) but is much more work.

@jstrachan @pradeepto wdyt ? Is this a feasible plan ? Or do you see a better integration point ?

jstrachan:
Sounds good to me! A few random thoughts:

* should we default to look in 'src/main/kompose/*' for the files to transform?
* the kompose-generator should have a config option of the file name /wildcard to look for compose/kompose files - then folks could just add fmp to their pom.xml and get kubernetes support even if the developer is already using compose files already - I.e. No need to move them, just point fmp at where they are

rhuss:
Yes, that was the idea to pickup kompose files directly from src/main, like we do for fragments.
Would love to group them all under `src/main/fabric8` to not pollute the `src/main` dir more. 
We can not put it into `src/main/fabric8/kompose/` though, as this would collide with auto-profile detection.

Maybe we should add an `src/main/fabric8/fragments/` where we pick up fragments, too, and where profile directories can be added. I know that sounds a bit confusing (profile directories are probably not well known, but see https://maven.fabric8.io/#profiles-using for "Profiles for resource fragments". Then we are free to have sub-dir in `src/main/fabric8` and we could also e.g. move `src/main/fabric8-includes` to `src/main/fabric8/includes`.

hrishin:
[...]

PR #969, opened 2017-06-27
Issue#968 - Multi Container Support with Resource Fragments

kameshsampath:
* added config to spring boot generator
* ability to parse multiple yaml docs with or without profiles

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
@@ -719,9 +719,20 @@ public class KubernetesResourceUtil {
                 int idx = 0;
                 for (Container defaultContainer : defaultContainers) {
                     Container container;
+                    boolean isNewContainerAdded = false;
                     if (idx < containers.size()) {
                         container = containers.get(idx);
                     } else {
+                        isNewContainerAdded = true;
+                        container = new Container();
+                        containers.add(container);
+                    }
+                    //TODO: SK remove it after review
+                    //Though we have more containers in the builder, but if existing one does not match
+                    // default container name then we assume it to be new one or the one to be added
+                    String containerName = container.getImage();
+                    String defaultContainerName = defaultContainer.getImage();
+                    if (!containerName.equalsIgnoreCase(defaultContainerName) && !isNewContainerAdded) {
                         container = new Container();
                         containers.add(container);
                     }
--- a/enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
+++ b/enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
@@ -17,9 +17,11 @@
 package io.fabric8.maven.enricher.api;
 
 import io.fabric8.kubernetes.api.builder.TypedVisitor;
-import io.fabric8.kubernetes.api.model.ContainerBuilder;
-import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
-import io.fabric8.kubernetes.api.model.Probe;
+import io.fabric8.kubernetes.api.model.*;
+import io.fabric8.kubernetes.api.model.extensions.Deployment;
+import io.fabric8.maven.core.util.Configs;
+
+import java.util.List;
 
 /**
  * Enriches containers with health check probes.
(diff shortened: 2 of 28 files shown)


core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -719,9 +719,20 @@ public static void mergePodSpec(PodSpecBuilder builder, PodSpec defaultPodSpec,
...
+                        isNewContainerAdded = true;
+                        container = new Container();
+                        containers.add(container);
+                    }
+                    //TODO: SK remove it after review
+                    //Though we have more containers in the builder, but if existing one does not match
+                    // default container name then we assume it to be new one or the one to be added
+                    String containerName = container.getImage();
+                    String defaultContainerName = defaultContainer.getImage();
+                    if (!containerName.equalsIgnoreCase(defaultContainerName) && !isNewContainerAdded) {
```
nicolaferraro: Here you're correlating over the image name. This works with sidecars, but it may have problems with pods having multiple containers from the same base image (with different startup options).

I think a better approach would be to correlate over the container image, i.e.
- If a fragment has a name, and the name is different the one that would be used for the current project, then add another container
- If the fragment does not declare a name, or the name matches the one that would be used for the current project, then use that as container for the project

enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
```diff
@@ -30,25 +32,67 @@ public AbstractHealthCheckEnricher(EnricherContext buildContext, String name) {
         super(buildContext, name);
     }
 
+    // Available configuration keys
+    protected enum Config implements Configs.Key {
+
+        probeMode {{ d = "all";}};
```
nicolaferraro: We can decide to enrich all containers here, or we can rely here on the concept of "project's main container", or list the containers that should be enriched using a configuration option (also a global one if this enricher is not the only one that should be scoped to a specific container). I'd not go for "first" / "last" as it relies too much on what the current implementation does but it may change in the future (with plugin updates or with "fragments" added in other ways e.g. #941).

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
kameshsampath: @nicolaferraro  - do you mean this 
```
...
for (Container defaultContainer : defaultContainers) {
                    Container container;
                    boolean isNewContainerAdded = false;
                    if (idx < containers.size()) {
                        container = containers.get(idx);
                    } else {
                        isNewContainerAdded = true;
                        container = new Container();
                        containers.add(container);
                    }
                    String containerName = container.getName();
                    String defaultContainerName = defaultContainer.getName() == null?defaultName:defaultContainer.getName();
                    **if (!defaultContainerName.equalsIgnoreCase(containerName) && !isNewContainerAdded)** {
                        container = new Container();
                        containers.add(container);
                    }
                    mergeSimpleFields(container, defaultContainer);
 ....
```

enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
kameshsampath: @nicolaferraro i agree with you on the first/last/all, how does something like this 
```
<config>
  <probes>
       <container-name-1/>
       <container-name-2/>
 </probes>
</config> 
```
when the probes is empty then we assume "all" and add probes to all containers

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
nicolaferraro: Something like that @kameshsampath, but I see that code adds another container if the fragment does not declare a name (that is the standard case when adding fragments). I'd add another container only if the one declared in the fragments uses a name that is different from the default container name for the project (e.g. it's called `proxy`). Please, also fix the dependencies of the 'it' tests as they refer snapshot artifacts.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/AbstractHealthCheckEnricher.java
nicolaferraro: I'd keep the configuration inside the enricher config, but a list of container names can be used (I don't know if the configuration utils support lists as you have declared them in the example, there should be other cases of list configs). About the default when no configuration is given: both 'all containers' or 'just the one having the default name' are reasonable in my opinion.

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
kameshsampath: @nicolaferraro - am not getting what you mean " I see that code adds another container if the fragment does not declare a name" - can you please point to the line# in the code for it ?

christophetd:
Any plans to continue working on this PR @kameshsampath? Otherwise, would you like some help on that?

kameshsampath:
@christophetd  adding @hrishin to this. he is one of the maintainers right now.

rohanKanojia:
There is a PR open already for multicontainer support based on discussions of Kamesh and Nicola(which is properly rebased with master). I'm closing this one.

PR #973, opened 2017-07-04
feat(upgrade): upgrade to latest kube client + fabric8

jstrachan:
upgrade to the latest client which supports API Groups and CRDs

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/access/ClusterAccess.java
+++ b/core/src/main/java/io/fabric8/maven/core/access/ClusterAccess.java
@@ -23,6 +23,7 @@ import io.fabric8.kubernetes.client.*;
 import io.fabric8.maven.core.config.PlatformMode;
 import io.fabric8.maven.docker.util.Logger;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.OpenShiftAPIGroups;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.utils.Strings;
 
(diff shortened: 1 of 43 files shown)


core/src/main/java/io/fabric8/maven/core/access/ClusterAccess.java
```diff
@@ -72,6 +73,18 @@ public String getNamespace() {
         return namespace;
     }
 
+    /**
+     * Returns true if this cluster is a traditional OpenShift cluster with the <code>/oapi</code> REST API
+     * or supports the new <code>/apis/image.openshift.io</code> API Group
+     */
```
rhuss: @jstrachan Just a remark: The following logic doesn't quite fit the comment above which returns true only if the old openshift detection returns true **and** the API group is supported. 

This means also that by default for older OpenShift clusters a Kubernetes build mode is used which creates images against a Docker daemon (was s2i up to now)

jstrachan: the method `openShiftClient.supportsOpenShiftAPIGroup(OpenShiftAPIGroups.IMAGE)` returns true if the old /oapi REST API is supported or the new API Group

jstrachan: see https://github.com/fabric8io/kubernetes-client/blob/master/openshift-client/src/main/java/io/fabric8/openshift/client/OpenShiftClient.java#L150-L153 and https://github.com/fabric8io/kubernetes-client/blob/master/openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java#L423

rhuss: understand. But when does it happen that isOpenShift() == true but not supporting neither the new nor the old OpenShift imagestream API ? (or in other words: when is isOpenShift() alone not enough)

jstrachan: We should move away from the idea that there's a boolean for whether a cluster is openshift or not really. At some point OpenShift will be a set of microservices that can run on any kubernetes cluster.

Right now we can easily add the `build.openshift.io` API Group to support `BuildConfig` and `Build` resources on kubernetes 1.7 so that we can support the jenkins sync plugin and the fabric8 / openshift.io console on vanilla kubernetes for Jenkins pipelines. There's no S2I (as thats supported via `image.openshift.io` API Group which isn't gonna run on kubernetes any time soon I suspect) but at least we can work with BuildConfig/Build resources from a Jenkins pipeline perspective in a consistent way between kubernetes and today's openshift clusters.

PR #979, opened 2017-07-11
fixes #978

### LINKED ISSUE #978: kubernetes dependencies with Template parameter expressions lose their expressions
we generate `k8s-template.yml` files so that folks can still use `oc process --local` or `gofabric8` to process templates before installing them onto pure kubernetes clusters. We can then reuse these parameter expressions to make helm charts too. 

So the `kubernetes.yml` file has no `${PARAM}` statements but the `k8s-template.yml` does.

However there's a bug in the dependency enricher where the `k8s-template.yml` resources are ignored so that all the template parameters get removed when combining YAML files together. It'd be nice to use the parameterised version instead. The resulting compose manifest then gets generated as the unparameterised `kuberenetes.yml` and the parameterised `k8s-template.yml` files

jstrachan:
lets preserve template parameter expressions in kubernetes dependencies

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DependencyEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DependencyEnricher.java
@@ -36,6 +36,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DependencyEnricher.java
```diff
@@ -53,9 +55,11 @@
  */
 public class DependencyEnricher extends BaseEnricher {
     private static String DEPENDENCY_KUBERNETES_YAML = "META-INF/fabric8/kubernetes.yml";
+    private static String DEPENDENCY_KUBERNETES_TEMPLATE_YAML = "META-INF/fabric8/k8s-template.yml";
```
pepperbot: MINOR Rename this field "DEPENDENCY_KUBERNETES_TEMPLATE_YAML" to match the regular expression '^[a-z][a-zA-Z0-9]*$'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS3008)

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DependencyEnricher.java
```diff
@@ -53,9 +55,11 @@
...
     private static String DEPENDENCY_OPENSHIFT_YAML = "META-INF/fabric8/openshift.yml";
 
     private Set<URL> kubernetesDependencyArtifacts = new HashSet<>();
+    private Set<URL> kubernetesTemplateDependencyArtifacts = new HashSet<>();
```
pepperbot: MAJOR Use the URI class instead. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2112)

PR #984, opened 2017-07-13
OSFUSE-605: build may be already finished in Openshift 3.7

nicolaferraro:
It seems that in Openshift 3.7 binary builds return a HTTP response only when the build is finished. Before 3.7, a build object was returned immediately, allowing us to do tail log.

The late response make chances of blocking indefinitely much higher, because the build may be already finished by the time we start watching on the "finish" event.

This should affect also the "master" branch.

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
+++ b/core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
@@ -81,6 +81,15 @@ public class BuildService {
                                                   "Failed to tail build log", logTerminateLatch, log);
             Watcher<Build> buildWatcher = getBuildWatcher(latch, buildName, buildHolder);
             try (Watch watcher = client.builds().withName(buildName).watch(buildWatcher)) {
+                // Check if the build is already finished to avoid waiting indefinitely
+                Build lastBuild = client.builds().withName(buildName).get();
+                String lastStatus = KubernetesResourceUtil.getBuildStatusPhase(lastBuild);
+                if (Builds.isFinished(lastStatus)) {
+                    log.debug("Build %s is already finished", buildName);
+                    buildHolder.set(lastBuild);
+                    latch.countDown();
+                }
+
                 waitUntilBuildFinished(latch);
                 logTerminateLatch.countDown();
                 build = buildHolder.get();


core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
```diff
@@ -81,6 +81,15 @@ public void waitForOpenShiftBuildToComplete(OpenShiftClient client, Build build)
...
+                String lastStatus = KubernetesResourceUtil.getBuildStatusPhase(lastBuild);
+                if (Builds.isFinished(lastStatus)) {
+                    log.debug("Build %s is already finished", buildName);
+                    buildHolder.set(lastBuild);
+                    latch.countDown();
+                }
+
                 waitUntilBuildFinished(latch);
                 logTerminateLatch.countDown();
                 build = buildHolder.get();
```
pepperbot: MINOR Introduce a new variable instead of reusing the parameter "build". [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1226)

nicolaferraro:
@rhuss can you have a look? This should be fixed soon..

core/src/main/java/io/fabric8/maven/core/openshift/BuildService.java
```diff
@@ -81,6 +81,15 @@ public void waitForOpenShiftBuildToComplete(OpenShiftClient client, Build build)
...
             try (Watch watcher = client.builds().withName(buildName).watch(buildWatcher)) {
+                // Check if the build is already finished to avoid waiting indefinitely
+                Build lastBuild = client.builds().withName(buildName).get();
+                String lastStatus = KubernetesResourceUtil.getBuildStatusPhase(lastBuild);
+                if (Builds.isFinished(lastStatus)) {
+                    log.debug("Build %s is already finished", buildName);
+                    buildHolder.set(lastBuild);
+                    latch.countDown();
+                }
+
```
rhuss: Couldn't this block just be moved up to the beginning of the method and just return if the build already has finished ? would make the logic a bit simpler. 

Or is it too early ? But I guess when this method is called in 3.7 the build is already finished ?

Also I wonder whether this is really true all the case as a build can take quite a long time and I wonder how this is dealt with timeouts on the API. I know that openshift used to uses "Expect: 100-continue" 
 HTTP header and this is supported in the latest OkHttp client. I opened an PR ehttps://github.com/fabric8io/kubernetes-client/pull/620 for this but I didn't really follow this. It was for timeouts happening in an error case. Not sure whether this is related to the issue here.

nicolaferraro: No, it can't be moved. The build returned by Openshift is not "Complete" sometimes, and it is finished right after the response has been returned, so an additional fetch is needed after starting the watch, in order to check if the "Complete" event has been sent in the meantime (probably we should open an issue to Openshift, but version 3.6 is on the way).

Actually this issue is related to Openshift 3.6+ (not 3.7+, I was confused by last Kubernetes release number).

About the timeout, I just did a build that lasted 5+ minutes without issues. I think that this change (in Openshift) will not have impacts on client side timeouts.

At least, I've not experienced timeouts.

nicolaferraro:
Ported to master.

rhuss:
@nicolaferraro looks like it break the CI build on master, e.g. https://travis-ci.org/fabric8io/fabric8-maven-plugin/builds/255127992?utm_source=github_status&utm_medium=notification

See also #990

nicolaferraro:
Looking at it..

Issue #880, opened 2017-03-15
Spring Boot Port Extractor - NumberFormat Exception


kameshsampath:
### Description
The Spring Boot port extractor fails to handle server.port with value like `sever.port: ${PORT:10800}` , throws NumberFormatException

### Info

* f-m-p version : 3.2.30
* Maven version (`mvn -v`) : 3.3.9
* OpenShift setup and version : 3.4.0
* If it's a bug, how to reproduce :
  * Create a simple simple spring boot project 
  * add a resource called `application.yml` with value of `server.port` to be equal to `${PORT:1234}`
 * run`mvn clean package`

### Stack Trace:

```
[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.2.30:resource (default) on project config-server: Execution default of goal io.fabric8:fabric8-maven-plugin:3.2.30:resource failed: Cannot extract generator: java.lang.NumberFormatException: For input string: "${PORT:8888}" -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:fabric8-maven-plugin:3.2.30:resource (default) on project config-server: Execution default of goal io.fabric8:fabric8-maven-plugin:3.2.30:resource failed: Cannot extract generator: java.lang.NumberFormatException: For input string: "${PORT:8888}"
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:185)
	at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:181)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #986, opened 2017-07-13
Allow to substitute key/value using a system property, otherwise use the default value for the server.port

cmoulliard:
This fix allows to resolve the issue reported within #880 

```
/**
     *  Look within all the System Properties if there is a property which corresponds to the key passed
     *  If a match exists, then replace the value otherwise use the default value defined within ${key:default-value}
     *  Example
     *  my.server.port=7777 is a system property that we will search in order to replace its value within the key passed to the method.
     *  If the key is "server.port: ${my.server.port:6666}", then "my.server.port" value will be substituted, otherwise the default value is used
     *  "6666"
     *
     *  @param key The string containing the key to search
     *  @return String The modified key
     */
```

### DIFF
--- a/generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
+++ b/generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
@@ -22,12 +22,7 @@ import java.io.FileOutputStream;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.UUID;
+import java.util.*;
 import java.util.zip.CRC32;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
@@ -45,6 +40,8 @@ import io.fabric8.maven.generator.javaexec.JavaExecGenerator;
 import com.google.common.base.Strings;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.text.StrMatcher;
+import org.apache.commons.lang3.text.StrSubstitutor;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.model.PluginExecution;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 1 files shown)


cmoulliard:
Can somebody has a look to please, .... ?

nicolaferraro:
Hi @cmoulliard, I'll look at it soon.

cmoulliard:
> build time system properties?

This is typically the case when you have a project that you can run locally or deployed on openshift/cloudfoundry. If the application contains many spring boot servers to be started locally, then it will be required to specify different port numbers like also the HOSTNAME of the services to be accessed

```
application1.yaml
server.port: ${PORT:9990}
spring.rabbitmq.host: ${RABBIT_HOST:localhost}

application2.yaml
server.port: ${PORT:9991}
spring.rabbitmq.host: ${RABBIT_HOST:localhost}
...
```
When I will start locally the Application1 or Application2 using the following `mvn spring-boot:run`, it is not required to change the port number or the address of the Rabbit Messaging broker.

Of course when the same project is build/deployed on OpenShift, it could be required to change the PORT and that will impose that we do 2 modifications; 1) First to pass a maven property or system property used by f-m-p to avoid the NumberException and 2) We add it to the JAVA_OPTIONS env key defined for the DeploymentConfig `JAVA_OPTIONS=-DPORT=8080`

For the other key, it will be required that we specify the DNS Name of the Rabbit instance deployed to `rabbitmq` to access it.

generator/spring-boot/src/main/java/io/fabric8/maven/generator/springboot/SpringBootGenerator.java
```diff
@@ -124,12 +124,36 @@ protected boolean isFatJar() throws MojoExecutionException {
...
+     *  If a match exists, then replace the value otherwise use the default value defined within ${key:default-value}
+     *  Example
+     *  my.server.port=7777 is a system property that we will search in order to replace its value within the key passed to the method.
+     *  If the key is "server.port: ${my.server.port:6666}", then "my.server.port" value will be substituted, otherwise the default value is used
+     *  "6666"
+     *
+     *  @param key The string containing the key to search
+     *  @return String The modified key
+     */
+    protected static String replaceKeyWithSystemPropertyValue(String key) {
```
cmoulliard: As Nicola proposed also that we support maven properties and system properties where maven props override system props, I will update the PR accordingly.

Nicola suggested too to move the code of the method `replaceKeyWithSystemPropertyValue` to the class SpringBootUtil packaged within f-m-p core. Do you agree too @nicolaferraro @rhuss ?

kameshsampath:
@nicolaferraro / @rhuss  / @cmoulliard  -  I think we can make this change in common JavaExecGeneator as  then this can be re used by other frameworks as well. Doing in SpringBootGen alone will make it it spring boot specific .. thoughts ?

nicolaferraro:
Well, I asked to move the logic of property resolution to `SpringBootUtils` because I think the syntax/semantics of `application.properties` is Spring-Boot specific, so it can be reused in other parts of the plugin that need to read the configuration.

E.g. the `${prop:default}` syntax is valid within spring-boot, but each property parsing lib (XProperty, Common Config, ...) can use a different syntax afair.

Dynamic configuration of the port using system/maven properties is already available for JavaExecGenerator (and also SpringBootGenerator) using the `webPort` configuration parameter of the generator.

PR #1004, opened 2017-07-25
Fixes #965 by improving spring-boot probes

### LINKED ISSUE #965: Readiness/Liveness probes for srping boot should be aware of server.servlet-path property
Generated probes for spring boot application need to check the `server.servlet-path` property.

There are cases, where mvc path is moved from / to a custom location. 
Example: https://docs.spring.io/spring-boot/docs/current/reference/html/howto-actuator.html#howto-use-actuator-with-jersey

For those cases, it would be really nice if f-m-p could read the configuration and adapt the path to the health endpoint.

nicolaferraro:
I've added fixed most of the configuration cases involving server context path, management context path, servlet path, management key store and server key store.


nicolaferraro:
Thanks @rhuss :)

nicolaferraro:
[merge]

PR #1011, opened 2017-08-02
Fixes #1003: adding fabric8.debug.suspend flag

### LINKED ISSUE #1003: fabric8:debug allow suspend y/n to be optionally specified via a property
it would be nice to enable suspend mode when debugging so that the JVM pauses and waits for the debugger to connect before starting; for cases where you wanna debug a failing JVM

nicolaferraro:
This requires the base image to support the env variable `JAVA_DEBUG_SUSPEND` (https://github.com/fabric8io-images/run-java-sh/pull/21 and other PRs to come for the base images).

Each run of `fabric8:deploy -Dfabric8.debug.suspend` restarts the pod, so a debugger can attach at startup. Readiness probes are disabled.

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/DebugMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/DebugMojo.java
@@ -17,7 +17,10 @@ package io.fabric8.maven.plugin.mojo.develop;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.Random;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.CountDownLatch;
 
 import io.fabric8.kubernetes.api.Controller;
@@ -76,11 +79,15 @@ public class DebugMojo extends ApplyMojo {
     @Parameter(property = "fabric8.debug.port", defaultValue = "5005")
     private String localDebugPort;
 
+    @Parameter(property = "fabric8.debug.suspend", defaultValue = "false")
+    private boolean debugSuspend;
+
     private String remoteDebugPort = DebugConstants.ENV_VAR_JAVA_DEBUG_PORT_DEFAULT;
     private Watch podWatcher;
     private CountDownLatch terminateLatch = new CountDownLatch(1);
     private Pod foundPod;
     private Logger podWaitLog;
+    private String debugSuspendValue;
 
     @Override
     protected void applyEntities(Controller controller, KubernetesClient kubernetes, String namespace, String fileName, Set<HasMetadata> entities) throws Exception {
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/DebugMojo.java
```diff
@@ -185,13 +197,23 @@ public void onClose(KubernetesClientException e) {
...
-        throw new MojoExecutionException("Could not find a running pod with $" + envVarName + " = " + envVarValue);
+        throw new MojoExecutionException("Could not find a running pod with environment variables " + envVars);
     }
 
     private boolean isAddOrModified(Watcher.Action action) {
         return action.equals(Watcher.Action.ADDED) || action.equals(Watcher.Action.MODIFIED);
     }
 
+    private boolean podHasEnvVars(Pod pod, Map<String, String> envVars) {
+        for (String envVarName : envVars.keySet()) {
```
pepperbot: MAJOR Iterate over the "entrySet" instead of the "keySet". [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2864)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/DebugMojo.java
```diff
@@ -251,6 +273,23 @@ private boolean enableDebugging(HasMetadata entity, PodTemplateSpec template) {
                         container.setPorts(ports);
                         enabled = true;
                     }
+                    if (debugSuspend) {
+                        // Setting a random value to force pod restart
+                        this.debugSuspendValue = String.valueOf(new Random().nextLong());
+                        KubernetesResourceUtil.removeEnvVar(env, DebugConstants.ENV_VAR_JAVA_DEBUG_SUSPEND);
+                        KubernetesResourceUtil.setEnvVar(env, DebugConstants.ENV_VAR_JAVA_DEBUG_SUSPEND, this.debugSuspendValue);
```
rhuss: Shouldn't `KubernetesResourceUtil.setEnvVar()` overwrite an existing variable if already present ? I would prefer this over an explicit removeEnvVar before, so that its more aligned to common practice. Also, setEnvVar should also ensure that it does not add the same env var twice.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/DebugMojo.java
```diff
@@ -251,6 +273,23 @@ private boolean enableDebugging(HasMetadata entity, PodTemplateSpec template) {
...
+                        container.setEnv(env);
+                        if (container.getReadinessProbe() != null) {
+                            log.info("Readiness probe will be disabled on " + getKind(entity) + " " + getName(entity) + " to allow attaching a remote debugger during suspension");
+                            container.setReadinessProbe(null);
+                        }
+                        enabled = true;
+                    } else {
+                        if (KubernetesResourceUtil.removeEnvVar(env, DebugConstants.ENV_VAR_JAVA_DEBUG_SUSPEND)) {
```
rhuss: Could you always set the variable with 'true', 'false' (and evaluate it accordingly in the startup script). This would simplify the code here (you don't need an "if" then, but set it always) and it makes it more explicit for the user within the container whether suspending is switched on or off (since the env var is always set).

nicolaferraro:
[merge]

Issue #977, opened 2017-07-10
ConfigMaps content from external data (fragment/xml config)


jstrachan:
[here's an example](https://github.com/fabric8-quickstarts/spring-boot-webmvc/blob/master/src/main/fabric8/configmap.yml#L4-L7
) of where we embed an `application.properties` file inside a `ConfigMap`. We may want to maintain the contents separately - so its easier to test the file for example (e.g. parse it or use it in a junit test case).

So maybe we could replace this with something like:
```yaml
metadata:
  annotations:
    file-import.maven.fabric8.io/application.properties: ${basedir}/target/test-classes/application.properties
```
which would load the given key `application.properties` from the given local file system file and import it into the generated YAML like this https://github.com/fabric8-quickstarts/spring-boot-webmvc/blob/master/src/main/fabric8/configmap.yml#L4-L7

rhuss:
Agree that this makes much sense to prepare configmaps and secrets (see also #975 ) from files. Instead of using annotations for fragments though (which would have no meaning during runtime), I'd prefer to add an XML configuration for preparing such cm / secrets. It would make the definition of such imports more concise:

```xml
<configuration>
   <resources>
      <configmaps>      
          <cm>
               <!-- Plain entry -->
               <entry>
                    <name>myKey</name>
                     <value>myValue</value>
               </entry>
              <!-- Import with name -->
              <entry>
                    <name>config</name>
                    <file>${basedir}/target/test-classes/application.properties</file>
               </entry>
               <!-- Would import with the filename as key. Or just use an <entry> without name -->
              <import>${basedir}/target/test-classes/application.properties</import>
               .....
          <cm>
      <configmaps>
   </resources>
</configuration>
```

If we would go for support for yaml annotation, I would like then to remove the annotation befire adding it to kubernetes.yml (as it exposed build details which are of no use and maybe even unwanted during deployment time)

jstrachan:
Am happy with the XML and/or annotation approach. I guess the XML approach may be easier to tool one day (IDEs might be smart enough to grok <file> is-a file etc). 

Agreed that if we did go with annotations we should remove the `file-import.maven.fabric8.io/*` annotation during enrichment.

yuwzho:
[...]

PR #1013, opened 2017-08-04
Make apply secrets to kubernetes cluster security

yuwzho:
refer to issue: #977 

This pull request contains:

- [x] Add an enricher to resolve kubernetes docker-registry secret using serverId in yaml source fragment with annotation, this yaml source fragment file should be put under `src/main/fabric8` folder
- [X] Configure kubernetes docker-registry secret using serverId in `pom.xml` configuration's resource list
- [x] Use goal `fabric8:resource` to compile the above two ways' resource into `target/classes/META-INF/fabric8/` as well as other resources.
- [X] Use goal `fabric8:apply` to apply compiled kubernetes yaml file to cluster

**Not include in this PR**
- [ ] Prevent secret resource shared to helm charts
- [ ] environment specific


jstrachan:
@yuwzho thanks! will take a look first thing tomorrow

rhuss:
thanks @jstrachan ! please coordinate also with @nicolaferraro who also already started to review.

yuwzho:
Thanks @nicolaferraro 's review and comments. 

Agree that we should apply the secrets resources as well as other resources when call `apply` goal. 

But I don't think using filter flag to avoid writing secrets resource into yaml file is a perfect idea. Because:
* first, adding a flag increase the difficulty of usage. 
* Second, if user wants to package the resource to helm, it should set the filter-secret as true, while when he wants to apply to kubernets,it should set the filter-secret as false. This make the goals `resource` -> `helm` and `resource` -> `apply` must be two branches of execution. They are never be a continuous execute operation. Also, it is hard to configure in pom.xml.

Hence I want to make the following change:

* `apply` goal will apply the `target\...\fabric8\secrets\kubernetes.yml` and `target\...\fabric8\kubernetes.yml` to the cluster.
* `resource` goal will filter all secrets resources and writes into the `target\...\fabric8\secrets\` folder.

jstrachan:
Humble apologies for the delay responding; been a bit swamped lately!

So I agree we definitely need a way to generate `environment specific` resources that don't get packaged up with the app (could be `ConfigMaps` or `Secrets` plus things like `Services` too which link to external names/Endpoints) and a way to apply them.

We've some cases where we do generate Secrets as part of an App. e.g. to store the maven `~/.m2/settings.xml` when doing builds with maven inside Jenkins pods. I've seen other apps that include a Secret - but parameterise the username/password in a Template.

So I don't think its the case that all Secrets go to some place by default; its more a case we need to define environment specific stuff somewhere - which is then always kept outside of the app thats deployed. 

Maybe the simplest way to keep these things separated is to use separate folders; so maybe if you add stuff into `src/main/fabric8` it gets included in the app; if you added to somewhere else, say `src/main/fabric8-environment` then its kept separately and never included in the JAR and never versioned and released to a maven repository and whatnot.

e.g. an app developer may decide that a particular `ConfigMap` called `foo` is not included in an app but is managed on a per environment basis

I'm cool with the `mvn fabric8:resource` goal generating all the resources in one go as thats cheap and fast. I'm wondering if we need a separate goal though to apply _environmental_ resources. e.g. maybe that should be done via `mvn fabric8:deploy-env` or something? For example if we're generating Secrets randomly for an environment - we don't wanna redeploy that each time we redeploy the app.

One other thought; we just added an 'override' mechanism to let us easily tweak resources (e.g. adding env vars, adding/editing/removing annotations/configmap entries): https://github.com/fabric8io/fabric8-maven-plugin/pull/1023 - we may wanna do something similar for environment specific resources too?

e.g. lets consider if you want to use an environment specific `Service` that points to some external service for each environment. We may want to do something like this...

```
$ find .
src/
  main/
    fabric8-enviroment/
       my-service.yml
       prod/
          my-service.yml
       test/
          my-service.yml
       uat/
          my-service.yml

$ mvn fabric8:deploy-env -Denv=prod
Deploying the environment resources to the environment: prod
Applying resource Service: my

```
i.e. then folks can keep a list of overrides for each environment containing which hard coded values to use for env vars / Service.externalName, ConfigMap.data[key] or anontations - or use different files/envvars for defining Secrets etc

jstrachan:
after that ramble - let me comment more directly on the PR

* I'm wondering if we use `src/main/fabric8-environment` instead of `src/main/fabric8/secrets` as its really about environment specific stuff; not just Secrets. (Though Secrets are the main kind of resource that will differ on a per environment basis but ConfigMap + Services could too - maybe even Delpoyments / Ingress as well?)
* maybe we should rename the mvn goals too to be `environment / env` related instead of `secrets`. e.g. how about we added the 'env-' prefix to the [existing goals](https://maven.fabric8.io/) we have? 
  * `mvn fabric8:env-resource`
  * `mvn fabric8:env-deploy`
  * `mvn fabric8:env-resource-apply`
  * `mvn fabric8:env-apply`

yuwzho:
@jstrachan Of course, the environment specific is a super good idea for CI/CD. However I don't think need to add another folder to put several similar kubernetes yaml files for each environment.

First, for the docker registry secrets, you can find in the new added [sample](https://github.com/Microsoft/fabric8-maven-plugin/blob/master/samples/secret-config/pom.xml#L69), it uses the same way to refer a docker registry as fabric8 docker maven plugin feature. It can easily configure the docker username/password in `settings.xml` and refer it as a properties in `pom.xml` . It is a smooth way to build and push docker package and then apply the docker secrets to cluster, the cluster can use this secret to pull image from docker registry pushed just now. If we here make k8s docker secrets resource environment specific, why not we handle the docker maven plugin as well? So I don't think environment specific is a thing only about k8s resource.

Then, the kubernetes resource fragment's template should be all the same for each environment, but they hold different values. To get a different value, it can refer different values from project properties. To specific the properties, the `pom.xml` can import different properties file according to environment flag.

Above all, I think we should put the environment specific value in a properties file but not separate the resource file in different folder. And it is a totally different feature with this one.

yuwzho:
@nicolaferraro I update the `resource` goal and `apply` goal to merge the `secrets` and `apply-secrets`. Could you please review it?

rhuss:
@yuwzho : fyi, @nicolaferraro is away currently, but will be back next week. hope this is still fine for you.

nicolaferraro:
Hi @yuwzho, sorry for the delay but I was a bit off keyboard in this period.
The PR looks really good.

I've just one major remark. You see from previous comments that there's still no agreement on how to manage resources that should not be included in helm/maven-deploy and whether only secrets should be handled differently or also ConfigMap, Services, i.e. let users choose what not to export vs. do it automatically.

For this reason, I think we should separate the creation of secrets from the docker server id (that is the main topic of this PR) from the change in the packaging workflow (we will open another issue to track that).

This means we should **not define a special input folder** for secrets (`src/main/fabric8/secrets/` in this PR. Note that it's folder already used by profiles, so it would have been changed anyway), and let them be added to `src/main/fabric8/` like any other resources. And **no special output folder** for secrets.

If we add a special handling for secrets now, we cannot remove it when we find a better strategy.

I hope you agree with this change. Let me know if you've time to do it, otherwise I can work on it in next few days.

yuwzho:
@nicolaferraro , agree with you we should keep the logic as simple as we can before coming up with a better strategy. I now put the secret resource at  `src/main/fabric8/` as well as other resources. Also, there is no special for the output.

nicolaferraro:
@yuwzho I've done some minor changes and adjusted documentation.

I've changed the default namespace used for the secrets, which is now empty by default and not "default". It is the convention, and if we set it to "default", there's no way for the user to specify the *current namespace*.

If you have no more comments, I think we are ready to merge it! :medal_sports:

yuwzho:
@nicolaferraro I go through all the changes. I have no more comments. Please merge it! Thanks. 😄

nicolaferraro:
Thanks a lot @yuwzho!
:smiley:

nicolaferraro:
[merge]

PR #1018, opened 2017-08-10
Update pom property version.fabric8 to 2.3.4

fabric8cd:


fabric8cd:
[merge]

jstrachan:
updating in separate PR as we need to upgrade the kube client too #1022

Issue #1009, opened 2017-08-01
JAVA_OPTIONS not merged when using the vertx generator


cescoffier:
I'm not sure if it's the desired behavior or a bug...

### Description

When the user uses a `deployment.yml` setting `JAVA_OPTIONS`, the value is overridden by the generator (
https://github.com/fabric8io/fabric8-maven-plugin/blob/master/generator/vertx/src/main/java/io/fabric8/maven/generator/vertx/VertxGenerator.java#L59)

Anything that needs to be done to "append" both ?

### Info

* f-m-p version : 3.5.1
* Maven version (`mvn -v`) : 3.5.0
* Kubernetes / OpenShift setup and version : v1.5.1+7b451fc, kubernetes v1.5.2+43a9be4 (minishift)

nicolaferraro:
Definitely this is not the desired behavior. Similar issues have been raised in #957 and #937.

There's not a default way to "append" to a env variable in f-m-p. Usually env-vars having multiple "declarations" are subject to a sort of "priority" and only one of them is used. In this case it seems the "priority" for `JAVA_OPTIONS` is not what a user expects.

I think we can default to an "append" mode for some environment variables like `JAVA_OPTIONS`, and this can be configurable. Wdyt @rhuss ?

nicolaferraro:
@cescoffier I'm trying to investigate on the current behavior to understand what the plugin should do. I've added a IT test [here])https://github.com/nicolaferraro/fabric8-maven-plugin/blob/19b8f8985e671edcb47f0bfbdecf3e9a95609fde/it/src/it/vertx/expected/kubernetes.yml#L45-L46), but what I see is:
- The `JAVA_OPTIONS` created by the generator are put in the Docker image
- The `JAVA_OPTIONS` in the resource fragment are put in the kubernetes resources

At the end, the first `JAVA_OPTIONS` (`-Dvertx.cacheDirBase=/tmp`) are used during the build (S2I or Docker), while the one in `deployment.yml` is used at runtime in the Pod, and it overrides the one in the docker image (not vice-versa).

Is this correct?

cescoffier:
Right now I see the opposite. The options from the generator are used and not the one from the deployment.yml.

nicolaferraro:
Interesting... can you share your project, so I can check what's happening?

cescoffier:
Hi @nicolaferraro 

[...]

PR #1033, opened 2017-08-29
Allow to set JAVA_OPTIONS

nicolaferraro:
This relates to #1009. Also @kameshsampath is interested in fixing it.

This basically changes the way environment variables defined at generator level are put into the container. 

Previously:
- In **Kubernetes** mode: they were set in the docker image
- In **Openshift** mode: they were set in the resource fragment through a openshift s2i enricher

The problem is that if the user wanted to set his own value e.g. of `JAVA_OPTIONS` in Openshift mode, he couldn't because the openshift s2i enricher overrode the custom values.

Now:
- In **Kubernetes** mode: unchanged
- In **Openshift** mode: using the [.s2i/environment file](https://github.com/openshift/source-to-image/blob/master/README.md#build-workflow), the environment variables are encoded in the image, as done in Kubernetes mode

@hrishin, @rhuss  or @jstrachan ... can you have a look before merging?


hrishin:
@nicolaferraro really looks good from implementation standpoint :+1:

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -93,11 +107,57 @@ public void build(ImageConfiguration imageConfig) throws Fabric8ServiceException
...
+            return dockerTar;
+        } catch (MojoExecutionException e) {
+            throw new Fabric8ServiceException("Unable to create the build archive", e);
+        }
+    }
+
+    private ArchiverCustomizer getS2ICustomizer(ImageConfiguration imageConfiguration) throws Fabric8ServiceException {
+        try {
+            if (imageConfiguration.getBuildConfiguration() != null && imageConfiguration.getBuildConfiguration().getEnv() != null) {
+                final File environmentFile = new File(config.getBuildDirectory(), "s2i-env-" + imageConfiguration);
```
rhuss: Do you rely here on the toString() from the imageConfiguration (i.e. that it uses the default with a random memory address) ? I would do this is in a more controlled manner, also to avoid unexpected chars in the filename).

nicolaferraro: Uh, thanks for spotting this :+1:

Probably a refactoring issue...

nicolaferraro:
[merge]

cescoffier:
Great ! Thanks !

Issue #1032, opened 2017-08-28
Improvements of the Vert.x Generator and enrichers


cescoffier:
**Important** - this is more an umbrella issue for me, to track some work I would like to do in the Vert.x generator / enrichers. So it's not a bug, just features that will get implemented soon

### Description

[ ] - Detects cluster manager from the project classpath and adds `-cluster` to the command
[ ] - Enable JMX metrics registration when vertx-dropwizard-metrics is in the classpath
[ ] - Allows passing parameter to the application
[ ] - If the Infinispan Cluster Manager is used, force IPv4
[ ] - Disable async DNS resolution (as it is not able to resolve names exposed by other namespaces)

cescoffier:
One of the task is related to https://github.com/fabric8io/fabric8-maven-plugin/issues/976.

cescoffier:
PR submitted: https://github.com/fabric8io/fabric8-maven-plugin/pull/1035

Issue #1019, opened 2017-08-11
Custom liveness/readiness probes are not being created


nicolaferraro:
The plugin should be already configured to create liveness/readiness probes from custom user definition, as in the following example:

```
<resources>
   <readiness>
      <getUrl>/monitor/</getUrl>
   </readiness>
</resources>
```

Anyway no probes seem to be generated by this configuration.
We should:
- fix the issue (see discussion in #1014)
- create an integration test to avoid regression
- document the feature in the doc

cescoffier:
@nicolaferraro I'm not sure the current config model does not support nested structure. See my other PR where I'm using the Maven configuration directly.

Right now, this does not work because the `ResourceConfig` does not have the readiness and liveness attributes (io.fabric8.maven.core.config.ResourceConfig:100).

rhuss:
@nicolaferraro you are right we should value XML configuration over enricher config when reconsidering the XML configuration in #18 

@cescoffier afais this issue is about the XML config syntax in general, not for a an enricher config. Nevertheless we are thinking about a deeper configuration for enrichers via #1375

PR #1035, opened 2017-09-02
Vertx improvements

cescoffier:
This PR provides a set of improvements around the Vert.x support (generator and enricher).

This PR fixes https://github.com/fabric8io/fabric8-maven-plugin/issues/1032. 

* Detects cluster manager from the project classpath and adds -cluster to the command
* Enable JMX metrics registration when vertx-dropwizard-metrics is in the classpath
* Allows passing parameter to the application
* If the Infinispan Cluster Manager is used, force IPv4
* Disable async DNS resolution (as it is not able to resolve names exposed by other namespaces)
* Implement ITs
* Allow full configuration of the health checks
* Update Vert.x generator doc
* Update Vert.x health check enricher doc


nicolaferraro:
@cescoffier this is a really good work! I've given it a brief look at it and I will do a in-depth review later.

I see you've looked a bit into #1019. I don't know the exact reasons why that doesn't work, but I wonder if we can use what you've done for vert.x to define global probes instead of a vert.x-related configuration. I see most of the work on the probes is not related to vert.x, so I think it would be awesome if we can extract the logic to use it for other platforms as well. Wdyt?

cescoffier:
@nicolaferraro Yes, definitely.

I've encountered a couple of issues around the configuration management in the plugin. While I understand why it was made like this, I, as a Maven "veteran", prefer going back to the roots ;-). 

Typically, I needed nested structures (a list for the exec command, a map for the HTTP headers...). I've built a parser for these. These methods can be moved somewhere else and shared among enrichers. 

I also agree that we can reuse the structure I've used to describe the probes in the other "Health" related enrichers. That would bring homogeneity and cover almost all cases described in the Kubernetes doc (I voluntarily don't support the `host` option on HTTP probe as it should be passed in the headers).

Let me know if you need my help. I've started having a (not so - still need a couple of hours) good understanding of the plugin now.

nicolaferraro:
Yeah, you're right about the current configuration model used inside enrichers, that is flat and cannot be used effectively to declare a probe.

But the hierarchical maven configuration using classes and `@Parameter`fields seemed to work to me in other parts of the plugin... In fact I've done some experiments abuot #1019 and it seems that the the bean `ProbeConfig` is filled in by maven if you declare a configuration like:

```
<!-- Root of plugin config -->
<configuration>
  <resources>
    <liveness>
      <getUrl>/thepath</getUrl>
    </liveness>
  </resources>
</configuration>
```

The problem is that the bean is not passed to the container handler for some reasons (need to investigate).

So, instead of providing a health check config in *enricher -> vertx-health-check -> *, I'd like to have **only one** global health check for any kind of project, so that if I want to add a custom probe to a simple java application I don't have to add another enricher and I don't need to change the spring-boot enricher if I'm using spring-boot.

The `liveness` (and `readiness`) object under *configuration -> resources* is not documented and it is not working since some time. I expect nobody is using it for this reason and we can make breaking changes to adapt it to the full Kubernetes model you've provided inside the vertx enricher.

Vertx users will not change the enricher configuration anymore, instead, they will change the global config (if this doesn't create problems).

This will also keep the enricher configuration model simple.

Do you agree with this change?

cescoffier:
@nicolaferraro Yes, I agree.

But to avoid breaking changes, we would need to be sure the specific health enrichers are still functional (with a warning notice) for a bit (let's say 2 months or something like that). 

About @Parameter, this works only on the first level of configuration. In the case of the `EnricherConfig` it's created using a builder, not using the Maven IOC mechanism. That's actually something I would like to discuss too: what about using @Component. But that's another subject.

nicolaferraro:
Agree on avoiding breaking changes. We can leave the vertx-helth-check enricher *as is* and deprecate it, also with a warning. And we can add a generic `custom-health-check` enricher, that has priority in the default-profile over all other health-check-enrichers that will read the configuration from the top-level xml config and create the probes.

Yes `@Parameters` cannot be used for specific enricher config (we can discuss about it later), but since we want to add a plugin top level configuration they should be ok. The only limitation iirc of `@Parameters` is that autocompletion works for the top-level classes declared in the mojo and not for nested classes, but injection (Maven IOC) works also in nested classes (there are 2-3 levels of `@Parameters` in the resource mojo).

rhuss:
I don't think it makes much sense for me to dive in a review right now, so I would like to follow @nicolaferraro assessment whether this PR can be merged or not. 'hope this is ok.

nicolaferraro:
We didn't have time to work on the generic health check enricher, but I still consider this PR good to be merged. I don't know if it is compatible with current way of handling vert.x applications or it introduces breaking changes. If @cescoffier agrees on merging this, I'd do it.

cescoffier:
Yes please merge.

nicolaferraro:
[merge]

nicolaferraro:
[merge]

nicolaferraro:
Thanks @cescoffier for the PR, it should have been merged a lot of time ago. My mistake :cry:

PR #1056, opened 2017-09-22
fixes #1055

### LINKED ISSUE #1055: cannot override a Route from a dependent jar
I'm trying to provide 2 options of a YAML; one with TLS enabled and one without. So I tried creating a separate maven project to override the `Route` resource:
https://github.com/jstrachan/fabric8-tenant-che/tree/malarkey/apps/che-http

though right now the custom Route is ignored https://github.com/jstrachan/fabric8-tenant-che/blob/malarkey/apps/che-http/src/main/fabric8/che-route.yml

I've got a PR coming with a fix...

jstrachan:
lets only replace an openshift specific dependent resource if the current resource is not already an openshift specific resource. i.e. so we can override openshift specific resources from dependents

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
@@ -713,13 +713,15 @@ public class ResourceMojo extends AbstractResourceMojo {
             // kubernetes App Catalog so we use a Template instead on OpenShift
             return null;
         }
+        // TODO-F8SPEC: ^^^^^ (end)
 
         // lets check if there's an OpenShift resource of this name already from a dependency...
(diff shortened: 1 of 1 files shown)


nicolaferraro:
Thanks, going to review it soon...

jstrachan:
@nicolaferraro agreed; things would be much simpler if we kept the kubernetes-specific and openshift specific resources completely separate at all times (from dependency / overloading / generating / enriching)

rhuss:
Tend to refer to my favourite issue here, too -->  https://github.com/fabric8io/fabric8-maven-plugin/issues/678 ;-) I really hope to get this soon (and also hope that I can jump on the fmp train again in a foreseeable future).

See also https://github.com/fabric8io/fabric8-build for an even more ambitious approach.

PR #1060, opened 2017-09-26
Based upon the code in BuildMojo, it seems that the property name sho…

jsight:
…uld be fabric8.build.strategy


nicolaferraro:
Thanks for fixing it!

nicolaferraro:
[merge]

PR #1067, opened 2017-10-10
upgrade to 3.x of the kubernetes-model

jstrachan:


jstrachan:
one change with the 3.x schema is the generated `apiVersion` values for `Route` and `DeploymentConfig` are now `route/v1` and `apps/v1`.  Am not sure if thats to be expected now and we need to amend the integration tests or if its a regression ;) e.g.

```yaml
---
apiVersion: apps/v1
kind: DeploymentConfig
---
apiVersion: route/v1
kind: Route
```
cc @hrishin

core/src/main/java/io/fabric8/maven/core/handler/ProbeHandler.java
```diff
@@ -83,6 +83,8 @@ private HTTPGetAction getHTTPGetAction(String getUrl) {
     }
 
     private TCPSocketAction getTCPSocketAction(String port) {
+        // TODO what to do about host?
```
hrishin: How about this?
probhandeler

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
```diff
@@ -89,7 +89,7 @@ public void visit(PodSpecBuilder builder) {
                 }
                 String jksSecretVolumeName = getConfig(Config.jksVolumeName);
                 if (!isVolumeAlreadyExists(builder.buildVolumes(), jksSecretVolumeName)) {
-                    builder.addNewVolume().withName(jksSecretVolumeName).withNewEmptyDir("Memory").endVolume();
+                    builder.addNewVolume().withName(jksSecretVolumeName).withNewEmptyDir().withMedium("Memory").endEmptyDir().endVolume();
```
hrishin: Is this change needed?

jstrachan: there's no parameters on `withEmptyDir()` any more - `withMedium()` was the only place I could find a string parameter ;) no clue really what we should do with the `"Medium"` string really

core/src/main/java/io/fabric8/maven/core/handler/ProbeHandler.java
jstrachan: sounds good to me. Fancy udpating this PR? :)

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/AutoTLSEnricher.java
hrishin: strange, in my case it's `builder.addNewVolume().withName(jksSecretVolumeName).withNewEmptyDir("Memory").endVolume();` still getting valid without any compilation error. 
I'm doing something wrong here then.

jstrachan: whoah - maybe its somehting wrong with my step; I tried that and it didn't compile for me. Wanna have a go at a PR yourself? :)

hrishin: Sure

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -593,7 +594,14 @@ public int compare(Pod p1, Pod p2) {
     public static Date getCreationTimestamp(HasMetadata hasMetadata) {
         ObjectMeta metadata = hasMetadata.getMetadata();
         if (metadata != null) {
-            return parseTimestamp(metadata.getCreationTimestamp());
+            return parseTimestamp(metadata.getCreationTimestamp().getTime());
+        }
+        return null;
+    }
+
+    private static Date parseTimestamp(Time time) {
```
pepperbot: MAJOR Remove this unused private "parseTimestamp" method. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUnusedPrivateMethod)

hrishin:
#1068

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -68,6 +68,7 @@
 import io.fabric8.kubernetes.api.model.PodTemplateSpec;
 import io.fabric8.kubernetes.api.model.ReplicationController;
 import io.fabric8.kubernetes.api.model.ReplicationControllerSpec;
+import io.fabric8.kubernetes.api.model.Time;
```
pepperbot: MINOR Remove this unused import 'io.fabric8.kubernetes.api.model.Time'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

jstrachan:
LGTM are you happy @nicolaferraro  or @ro14nd ?

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -593,7 +593,7 @@ public int compare(Pod p1, Pod p2) {
     public static Date getCreationTimestamp(HasMetadata hasMetadata) {
         ObjectMeta metadata = hasMetadata.getMetadata();
         if (metadata != null) {
-            return parseTimestamp(metadata.getCreationTimestamp());
+            return parseTimestamp(metadata.getCreationTimestamp().getTime());
```
nicolaferraro: `CreationTimestamp` is null for client generated objects. This will throw a `NullPointerException`.

jstrachan:
@nicolaferraro ideally the kube-model + kube-client should work for 3.5.x, 3.6.x and 3.7.x.

I wonder if there's a bug and it should be using `apiVersion: "v1"` rather than `apiVersion: "image/v1"`. I grok the new URI scheme of `/apis/image.openshift.io/v1/namespaces/fis/imagestreams` and all along with the different `kind` but I'm not sure yet what the `apiVersion` should be yet for the new `*.openshift.io` kinds yet I'm afraid

hrishin:
@jstrachan seems to kind of bug. 
In OpenShift alpha 3.7 client throwing exceptions with "ImageStream in version "v1" cannot be handled as a ImageStream: no kind "ImageStream" is registered for version "image/v1" while building the images.

jstrachan:
yeah - I remember now the apiVersion stuff. So the old version was "v1" for openshift resources. The new version is `apiVersion: (app|build|route|...).openshift.io/v1`. So the failing test cases lately have been due to the generation of an apiVersion of `apiVersion: route/v1` rather than `apiVersion: route.openshift.io/v1` - so thats definitely a bug - the new apiVersion should end with `.openshift.io/v1`

We may want to add some property on fabric8-maven-plugin to override the new version mechanism so that folks can generate YAML for old OpenShift clusters using the old "v1"?

nicolaferraro:
Yes, that's a problem that needs to be solved with a flag or we can also think to detect which version schema to use basing on the openshift version, so that users can specify the target version of Openshift and generate compatible yml/json artifacts. This can be done on f-m-p but also on other layers.

But there's more than resource generation. I've seen issues also in Apache Camel when we upgraded to latest kubernetes-client 3.x. There isn't yaml generation in camel-kubernetes, but the client was not able to read and update a configmap, so we had to downgrade to version 2.x.

jstrachan:
@nicolaferraro yeah, I had to downgrade fabric8-generator too (the 4.x version of forge) which creates/updates BuildConfigs. The fabric8-maven-plugin tests also broke too due to the `apiVersion` issue as well. 

Somehow the upgrade to the new schema has had an impact on the support of `apiVersion` which is a little odd really - as we moved to supporting the new API groups and the `apiVersion build.openshift.io/v1` stuff a few months ago on the 2.3.x client and we didn't have these failures.

jstrachan:
I've fixed the underlying issue with incorrect OpenShift API group versions (so `apps/v1 => apps.openshift.io/v1`).

@nicolaferraro I've tested the new master of the 3.x of kubernetes-client and its able to load/create/update ConfigMaps just fine too https://github.com/fabric8io/kubernetes-client/pull/892

Once the new release is out I'll update this PR to use it (and fix up the tests to use the correct `apps.openshift.io/v1` version)

jstrachan:
I've just updated to the latest kubernetes-model / client and fabric8 release which fixes the bad api versions; so they now use the proper `apps.openshift.io/v1` values now.

@nicolaferraro any chance you could retest kubernetes-client 3.0.1 sometime to see if things are working better now?

nicolaferraro:
Yes @jstrachan will do some tests today

hrishin:
@jstrachan looks good. 
We have tested these changes with OpenShift 3.7 but build request is throwing some exception.
```
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://192.168.42.225:8443/apis/build.openshift.io/v1/namespaces/myproject/buildconfigs/fabric8-maven-sample-spring-boot-s2i/instantiatebinary?commit=. Message: no kind is registered for the type build.BinaryBuildRequestOptions. Received status: Status(apiVersion=v1, code=400, details=null, kind=Status, message=no kind is registered for the type build.BinaryBuildRequestOptions, metadata=ListMeta(resourceVersion=null, selfLink=null, additionalProperties={}), reason=BadRequest, status=Failure, additionalProperties={}).
```
Issue is `instantiatebinary?commit=`. If we removed `commit` API is working fine. Need to fix request URI.

jstrachan:
Yeah - we hit a few similar issues a while back when we first started supporting the new API Groups APIs; some query parameters are now no longer supported. Lets raise an issue and fix it in KubernetesClient maybe?

jstrachan:
@nicolaferraro on your comment about the new client/model causing issues updating a ConfigMap - I hit that today ;) https://github.com/fabric8io/kubernetes-model/issues/241 should be fixed now though

nicolaferraro:
@jstrachan I think we can merge this. It seems to be working now with latest kubernetes-client. Is it ok for you?

jstrachan:
@nicolaferraro yeah - am totally happy to merge this

nicolaferraro:
[merge]

Issue #1070, opened 2017-10-17
Using its own BuildConfig when using f-m-p


zonArt:
Is it a way to define its own `BuildConfig` when using f-m-p (like adding a bc.yml file in src/main/fabric8) ?

### Description
Got my `BuildConfig` always overridden by fabric8 plugin and would really like to tune it and use it 
### Info

* f-m-p version : v3.5.31
* Maven version (`mvn -v`) : 3.3.9
* Kubernetes / OpenShift setup and version : 
openshift v1.5.1
kubernetes v1.5.2+43a9be4
* If it's a bug, how to reproduce :
having this kind of file in src/main/fabric8:
```
apiVersion: v1
kind: BuildConfig
metadata:
  name: mybuildconfig
  namespace: projectname
spec:
  source:
    binary: {}
    type: Binary
  output:
    pushSecret:
      name: artifactory
    to:
      kind: ImageStream
      name: myimagestream
  strategy:
    dockerStrategy: {}
    type: Docker
  triggers:
  - generic:
      secret: dummysecret
    type: Generic
  nodeSelector:
    kubernetes.io/hostname: hostname
```
It gets overridden when executing `mvn fabric8:build` with especially annoying removal of `pushSecret` because of this:
https://github.com/fabric8io/fabric8-maven-plugin/blob/9102af60e232c1ac6e2cc29d81ffd8f13f562bab/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java#L179-L182

Thank you in advance for your kind help

FlorianBois:
I'm asking myself if the plugin is really designed to have  BuildConfig and ImageStream really overridable , my understanding of the lifecycle is : build > push > resource > apply > deploy > ...  However, with custom yml for BuildConfig and ImageStream , we are force to do the resource > apply before the build, which somhow do unwanted deployment (deployment done before the build)

Either, the plugin is able to deal with pushSecret and make its own BuildConfig and ImageStream before the build, forbidding the custom yaml for BuildConfig and ImageStream, or either , we can provide custom BuildConfig and ImageStream but we need specifc goals for build like build-resource  build-apply dedicated to the preparation of the build. 

what the bestthe strategy with openshift here ? this is maybe a topic for story : Fill in the gaps and decouple (https://github.com/fabric8io/fabric8-maven-plugin/projects/1 ) where openshift has a different way to do the build.

rhuss:
[...]

PR #1071, opened 2017-10-17
Keep BuildConfig configuration

### LINKED ISSUE #1054: Support DockerImage as output for Openshift builds
Pushing images from Openshift to external registries is not trivial. A blog post describing the procedure can be found here:

https://blog.openshift.com/pushing-application-images-to-an-external-registry

So, it seems that a build that has an output like:

    spec:
      output:
        to:
          kind: DockerImage   
          name: docker.io/veermuchandi/mytime:latest
        pushSecret:
          name: dockerhub

can directly push to an external registry as part of the build.

It would be fantastic, if we could optionally configure that option on the fabric8-maven-plugin, so that we can ease up the process of exporting images.

zonArt:
If wanting to update `buildConfig` with existing one
it could be good to keep the spec defined in it otherwise it
is the same than the recreate option
May fix #1054 and #1070

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
@@ -245,8 +245,8 @@ public class OpenshiftBuildService implements BuildService {
         if (!Objects.equals(buildStrategy, spec.getStrategy()) || !Objects.equals(buildOutput, spec.getOutput())) {
             client.buildConfigs().withName(buildName).edit()
                     .editSpec()
-                    .withStrategy(buildStrategy)
-                    .withOutput(buildOutput)
+                    .withStrategy(spec.getStrategy())
+                    .withOutput(spec.getOutput())
                     .endSpec()
                     .done();
             log.info("Updating BuildServiceConfig %s for %s strategy", buildName, buildStrategy.getType());


hrishin:
@nicolaferraro could you please?

FlorianBois:
While reviewing the commit, please check if my comment on the issue, if the fix really makes sens : https://github.com/fabric8io/fabric8-maven-plugin/issues/1070#issuecomment-338187679

zonArt:
I've maybe misunderstood the plugin's philosophy and if this is the case I'm sorry. But if not is it possible to help me finding a way to customize `BuildConfigs` and `ImageStreams` within f-m-p ?
Thanks

zonArt:
Any news on this one ?

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -245,8 +245,8 @@ private String updateBuildConfig(OpenShiftClient client, String buildName, Build
         if (!Objects.equals(buildStrategy, spec.getStrategy()) || !Objects.equals(buildOutput, spec.getOutput())) {
             client.buildConfigs().withName(buildName).edit()
                     .editSpec()
-                    .withStrategy(buildStrategy)
-                    .withOutput(buildOutput)
+                    .withStrategy(spec.getStrategy())
+                    .withOutput(spec.getOutput())
```
rhuss: I'm really not sure why this should fix something, as this is a no-op after all: You are re-setting the same strategy and output as it was there already. The whole idea of this method is to _change_ the strategy (docker or binary s2i build) and the output stream if changed.

@zonArt could you please elaborate what are you trying to achieve ?

rohanKanojia: @rhuss : I read the related issue file by @zonArt , I think he wants to have his pushSecrets kept in `BuildConfig` to push/pull from some private registry, maybe that's why he is setting same strategy and output itself.

Agreed that this doesn't align with plugin's flow at all. But is there some way to configure plugin to include these pushSecrets into BuildConfig?

zonArt: @rohanKanojia You got it totally right, being able to keep the configuration of the buildconfig was the needed behavior thus this PR which was maybe not intended to be merged as is but more a way to request support on how to customize those objects or to find a way to keep provided ones (maybe with a flag)

FlorianBois:
Here another proposal, to add pushSecret and pullSecret to the builld config.

https://github.com/fabric8io/fabric8-maven-plugin/compare/master...ElcaIT:Feature/ThirdPartyRepositoryWithSecret

However, this probably need some rework : 
- The build time out of the client has been increased
- this line withNewSpec().withDockerImageRepository(imageConfiguration.getRegistry()+"/"+imageName.getNameWithoutTag())
on OpenshiftBuildService is highly to be wrongly coded I took the registry from the imageConfiguration, but it is possible that it should come from an additional configuration or reuse another one (currently, there is many ways to tell the plugin there is a repository),  taking the imageName without hte image tag looks also more like a workaround than a real solution.

rhuss:
@zonArt @FlorianBois Let me try to somerise the issue you are trying to solve. It's all about being able to configure pull and push secrets for an OpenShift S2I build, right ?

If so, we should indeed extend the configuration and implement this like @FlorianBois suggests. 

I suggest to close this PR and work on the diff referenced above. @FlorianBois could you please create a PR and we can then discuss over there about the naming, timeout configuration and registry handling ?

PR #1096, opened 2017-11-15
Fixes #1042 Switch to using Deployments rather than DeploymentConfig on OpenShift when not using ImageStreams

### LINKED ISSUE #1042: switch to using Deployments rather than DeploymentConfig on OpenShift when not using ImageStreams
DeploymentConfig's are quite a bit worse to use than Deployments due to the extra deploy pods that sometimes timeout/fail, stay around, add confusion - whereas Deployments seem pretty rock solid! I've never seen a Deployment ever not deploy or timeout - whereas I've seen that happen lots with DCs.

So I wonder should we avoid DeploymentConfigs when not using ImageStreams on OpenShift? I guess we could make it a flag so folks could decide to keep DCs if they really want; but I'd have thought sticking with standard kubernetes resources OOTB sounds a nicer default (providing folks are not using ImageStreams)

rohanKanojia:
Fabric8-maven-plugin should always use DeploymentConfigs if using ImageStreams; but if not we should probably default to Deployments.

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
@@ -49,6 +49,7 @@ import io.fabric8.maven.plugin.enricher.EnricherManager;
 import io.fabric8.maven.plugin.generator.GeneratorManager;
 import io.fabric8.openshift.api.model.DeploymentConfig;
 import io.fabric8.openshift.api.model.Template;
+import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.utils.Lists;
 import io.fabric8.utils.Strings;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -174,6 +175,8 @@
     @Parameter(property = "fabric8.build.strategy")
     private OpenShiftBuildStrategy buildStrategy = OpenShiftBuildStrategy.s2i;
 
+    @Parameter(property = "fabric8.build.switchToDeployment", defaultValue = "true")
```
hrishin: Would it be better to keep the default preference to `false`, so that existing users will not get surprising behavior :)
@jstrachan WDYT?

PR #1112, opened 2017-11-24
feature: Don't use a default for skipBuildPom

rhuss:
Instead, try to detect the default by checking whether there has been
some image configurations with build sections resolved.

This can always overwritten by explicitly specifying fabric8.skip.build.pom

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/BuildMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/BuildMojo.java
@@ -17,7 +17,6 @@
 package io.fabric8.maven.plugin.mojo.build;
 
 
-import java.awt.*;
 import java.io.File;
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/BuildMojo.java
```diff
@@ -17,6 +17,7 @@
 package io.fabric8.maven.plugin.mojo.build;
 
 
+import java.awt.*;
```
hrishin: If possible can we import specific packages?

rhuss: Oops, good point. That shouldn't be there at all ....

rhuss:
> Shall we document this change and update the doc?

I don't think that it is documented, but let me check.

hrishin:
Thanks @rhuss

hrishin:
[test]

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/BuildMojo.java
rhuss: @metacosm btw, the bogus import has been already removed within this PR at --> https://github.com/fabric8io/fabric8-maven-plugin/pull/1112/commits/a5a72194e8789812a048145e1fffc46839950c56

Issue #1134, opened 2017-12-12
improve unit test for 'core' module 'handlers' part


hrishin:
Implement the unit tests for classes falling under the [handlers](https://github.com/fabric8io/fabric8-maven-plugin/tree/master/core/src/main/java/io/fabric8/maven/core/handler) directory.

hrishin:
part of #1102 #1107

PR #1136, opened 2017-12-12
Added Test for Container Handler, ServiceHandler, PodTemplateHandler, DeploymentHandler, DaemonSetHandler, EnvVarHandler, ProbeHandler

piyush1594:
#1134


core/src/test/java/io/fabric8/maven/core/handler/ContainerHandlerTest.java
```diff
@@ -0,0 +1,324 @@
...
+        ImageConfiguration imageConfiguration2 = new ImageConfiguration.Builder().
+                name("test").alias("test-app").buildConfig(buildImageConfiguration2).registry("docker.io").build();
+
+        List<ImageConfiguration> images = new ArrayList<>();
+        images.add(imageConfiguration1);
+        images.add(imageConfiguration2);
+
+        containers = handler.getContainers(config, images);
+
+        assertEquals(2,containers.size());
```
hrishin: Can we assert other important container attributes as well? e.g. container name

piyush1594: @hrishin getImageName function is using ImageConfiguration as input to derive the image name, in ImageConfiguration class, there are only two relevant fields `name` and `registry`. ImageConfiguration class is from docker-maven-plugin. You can check the getImageName() code [here](https://github.com/fabric8io/fabric8-maven-plugin/blob/master/core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java#L96)

piyush1594:
@hrishin I added some more assertions for ContainerPart you asked and also added one-case for FQDN. Please review. Test for ServiceHandler is also added. Thanks

core/src/test/java/io/fabric8/maven/core/handler/ContainerHandlerTest.java
piyush1594: @hrishin This has been resolved in the new commit. Thanks

core/src/main/java/io/fabric8/maven/core/handler/ProbeHandler.java
```diff
@@ -72,22 +67,22 @@ public Probe getProbe(ProbeConfig probeConfig)  {
     // ========================================================================================
 
     private HTTPGetAction getHTTPGetAction(String getUrl) {
-        if (getUrl == null) {
+        if (getUrl == null || !getUrl.subSequence(0,4).toString().equalsIgnoreCase("http")) {
```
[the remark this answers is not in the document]
piyush1594: @hrishin PTAL. I divided the test cases into further smaller test cases and refactored the code.

And the point you mentioned regarding https, that is already included, the same will check both `http` and `https`.

Thanks

core/src/test/java/io/fabric8/maven/core/handler/JobHandlerTest.java
```diff
@@ -0,0 +1,172 @@
...
+                    .controllerName("TesTing")
+                    .withServiceAccount("test-account")
+                    .volumes(volumes1)
+                    .build();
+
+            jobHandler.getJob(config, images);
+        }
+        //asserting the exact message because
+        // it throws the same exception in case controller name is null
+        catch(IllegalArgumentException exception) {
```
hrishin: Instead of try-catch use "@Test(expected=IllegalArgumentException.class) annotation"

hrishin:
@piyush1594 this looks good now!
Thanks!

PR #1153, opened 2018-01-09
Fix VertxConfigmapBooster test

rohanKanojia:
A recent change in vertx-configmap-booster project has caused the build
process to automatically create resources during build(see src/main/fabric8
in that project). So it's better to edit configmap after build process.
For more details, see https://github.com/openshiftio-vertx-boosters/vertx-configmap-booster/commit/2afa3e1a66e8e895d9a6dc91d94880480a478f28


rohanKanojia:
@hrishin : Hmm, agreed. I also feel that these booster regression tests should be coupled with those booster projects rather than fabric8 maven plugin.
Roland was also asking about moving these tests out of Fabric8 maven plugin, so maybe we can add them to a repo on that Github org, or maybe add them individually? But that would require configuring CI on each project.

hrishin:
@rohanKanojia agreed

rohanKanojia:
@hrishin : Thanks :) . Could we merge this small fix for now? I have created a separate issue for it. Let's move our discussion [here](https://github.com/fabric8io/fabric8-maven-plugin/issues/1155)

rohanKanojia:
@piyush1594 : Can you merge this, please?

piyush1594:
Thanks @rohanKanojia 

LGTM, it is required as all the Travis builds for PR are getting failed, I am merging this. @hrishin

PR #1160, opened 2018-01-17
Fix #1149 : Handle problems in fmp regression tests

### LINKED ISSUE #1149: regression test for Vertx http booster is failing
### Description
The regression test for vertx http booster is failing on Travis CI. This behavior is sporadic and usually happens with OpenShift v3.6.0., v3.6.1
Its an issue with the assertion that verifies application endpoint response. Its observed that creating application route may take some time. Before that assertion invokes the application endpoint.

rohanKanojia:
+ Added a try catch block around makeHttpRequest(..); in case nip.io is down
+ Added retry() logic while checking vertx-http-booster's endpoints.


hrishin:
This looks good, lets merge it.

PR #1167, opened 2018-01-25
support for docker.pull.registry

MichaelSp:
In this Pull request the registry-handling is implemented according to the documentation: https://maven.fabric8.io/#registry

I discovered that for the `fabric8:resource` step, the `docker.pull.registry` is ignored. So here is the fix.

Signed-off-by: Michael Sprauer <Michael.Sprauer@sap.com>

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
+++ b/core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
@@ -16,29 +16,22 @@
 
 package io.fabric8.maven.core.handler;
 
-import java.util.ArrayList;
-import java.util.List;
-
-import io.fabric8.kubernetes.api.model.Container;
-import io.fabric8.kubernetes.api.model.ContainerBuilder;
-import io.fabric8.kubernetes.api.model.ContainerPort;
-import io.fabric8.kubernetes.api.model.ContainerPortBuilder;
-import io.fabric8.kubernetes.api.model.Probe;
-import io.fabric8.kubernetes.api.model.SecurityContext;
-import io.fabric8.kubernetes.api.model.SecurityContextBuilder;
-import io.fabric8.kubernetes.api.model.VolumeMount;
-import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
+import io.fabric8.kubernetes.api.model.*;
 import io.fabric8.maven.core.config.ResourceConfig;
 import io.fabric8.maven.core.config.VolumeConfig;
 import io.fabric8.maven.core.util.KubernetesResourceUtil;
 import io.fabric8.maven.docker.access.PortMapping;
 import io.fabric8.maven.docker.config.BuildImageConfiguration;
 import io.fabric8.maven.docker.config.ImageConfiguration;
+import io.fabric8.maven.docker.util.EnvUtil;
+import io.fabric8.maven.docker.util.ImageName;
 import io.fabric8.utils.Strings;
-
 import org.apache.maven.project.MavenProject;
 import org.json.JSONArray;
 import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * @author roland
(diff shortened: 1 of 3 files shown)


rohanKanojia:
Sorry for late review. Thanks a lot for this :+1: 

Hello @hrishin @nicolaferraro, Please see if you have time to review this. Shouldn't this change go in for docker maven plugin? Since fabric8-maven-plugin uses dmp as a dependency; and that deals with push/pull registries.

piyush1594:
@MichaelSp Can you do a little change to create an issue or fine if its already there and put the reference link in CHANGELOG.md and push again with squashed commits.

hrishin:
@MichaelSp thanks for the contribution :+1: 



> Shouldn't this change go in for docker maven plugin

Yes, that makes sense. But its small fix, shall we allow this exception?

@rohanKanojia  @piyush1594

core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
```diff
@@ -94,14 +87,19 @@ private String getImagePullPolicy(ResourceConfig config) {
...
     private String getImageName(ImageConfiguration imageConfiguration) {
-        if (Strings.isNullOrBlank(imageConfiguration.getName())) {
-            return imageConfiguration.getName();
-        }
+        if (Strings.isNullOrBlank(imageConfiguration.getName()))
+            return null;
+        String configuredRegistry = EnvUtil.findRegistry(
+                Strings.isNullOrBlank(imageConfiguration.getName()) ? null : new ImageName(imageConfiguration.getName()).getRegistry(),
+                imageConfiguration.getRegistry(),
+                project.getProperties().getProperty("docker.pull.registry"),
```
hrishin: Not sure if its good idea to use `pull.registry ` as a general registry(for pull or push).

MichaelSp:
rebased, sqashed, with changelog-entry and issue #1334

rhuss:
> Shouldn't this change go in for docker maven plugin? Since fabric8-maven-plugin uses dmp as a dependency

No, this change is specific for the fabric8-maven-plugin for calculating the image name in the resource descriptor. I.e. that's only relevant for `fabric8:resource` not `fabric8:build`. For `fabric8:build` the integrated d-m-p already evaluates this property properly.

core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
rhuss: Ideally we should also use system properties. You can use `Configs.getPropertyWithSystemAsFallback(project.getProperties(), "docker.pull.registry"`

core/src/test/java/io/fabric8/maven/core/handler/ContainerHandlerTest.java
```diff
@@ -257,6 +257,23 @@ public void getImageNameTest(){
         assertNull(containers.get(3).getImage());
     }
 
+    @Test
+    public void getRegistryTest() {
+        ContainerHandler handler = new ContainerHandler(project1, envVarHandler, probeHandler);
```
rhuss: sure, not an issue of this test, but I really wonder why this test class keeps all setup in member variables and not in local variables (e.g. a local MavenProject() instead of the member var). As it is now every test needs to be very careful to cleanup afterwards, as all the test method share state. It could be also that the tests become flaky as test methods are running in different order. 

Actually I highly recommend to use no shared state between test methods, except when you need extra features like injected mocks.

core/src/test/java/io/fabric8/maven/core/handler/ContainerHandlerTest.java
```diff
@@ -257,6 +257,23 @@ public void getImageNameTest(){
...
+        ImageConfiguration imageConfig = new ImageConfiguration.Builder().
+                name("test").alias("test-app").buildConfig(buildImageConfiguration1).build();
+
+        images.clear();
+        images.add(imageConfig);
+
+        project1.getProperties().setProperty("docker.pull.registry", "push.me");
+        containers = handler.getContainers(config1, images);
+
+        project1.getProperties().remove("docker.pull.registry");
```
rhuss: If `project1` would be a local variable, then you don't have to clean it up.

core/src/test/java/io/fabric8/maven/core/handler/ContainerHandlerTest.java
rhuss: @piyush1594 any particular reason why you introduced project, project1, project2 as member variables and don't create them as local variables when you need them ? It makes maintaining such tests quite hard.

core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
```diff
@@ -94,14 +87,19 @@ private String getImagePullPolicy(ResourceConfig config) {
     }
 
     private String getImageName(ImageConfiguration imageConfiguration) {
-        if (Strings.isNullOrBlank(imageConfiguration.getName())) {
-            return imageConfiguration.getName();
-        }
+        if (Strings.isNullOrBlank(imageConfiguration.getName()))
+            return null;
```
rhuss: please added braces also for single line ifs.

MichaelSp: done

MichaelSp:
👏

Issue #1166, opened 2018-01-24
Fix the Jenkins release workflow 


hrishin:
Current workflow F8MP release on Jenkins is 
checkout project snapshot -> tag release on Github-> build and test project -> publish the project to maven central

Change it to
checkout project snapshot -> build and test project -> tag release on Github ->publish the project to maven central

piyush1594:
@hrishin I am closing this issue as #1170 is merged. Thanks

PR #1169, opened 2018-01-29
Test for both CI and CD flows

piyush1594:
After adding this, the tests will run for both CI and CD process.
Ref #1166.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,12 +23,13 @@ clientsTemplate{
         sh "git remote set-url origin git@github.com:fabric8io/fabric8-maven-plugin.git"
 
         def pipeline = load 'release.groovy'
+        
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -23,12 +23,13 @@ clientsTemplate{
         sh "git remote set-url origin git@github.com:fabric8io/fabric8-maven-plugin.git"
 
         def pipeline = load 'release.groovy'
+        
+        // Test for both CI and CD flows till we have confidence in the process.
+        sh "mvn clean install"
```
hrishin: This will work.
It would be better if we can run `mvn verify` or equivalent in CD block.

piyush1594:
i am closing this as this is resolved in #1170

PR #1175, opened 2018-02-08
Fixes #1174: use image stream local lookup policy by default

### LINKED ISSUE #1174: Set lookupPolicy to local on Openshift imagestreams
### Description
Openshift ImageStreams currently can be used with DeploymentConfig, but not with Deployments or StatefulSets. The reason is that the ImageStream created by f-m-p has lookupPolicy not set to "local", so images required by Deployments or StatefulSets are fetched from docker hub instead of the internal Openshift registry.

Lookup policy is allowed since at least Openshift 3.6, so we should not have problems.
https://docs.openshift.com/container-platform/3.6/rest_api/openshift_v1.html#v1-imagestreamspec

I'll work on this if there are no problems.

### Info
* f-m-p version : 3.5.34

nicolaferraro:
This add the local lookup policy to image streams (fixes #1174).

@hrishin can you do a review?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ We use semantic versioning in some slight variation until our feature set has st
 
 After this we will switch probably to real [Semantic Versioning 2.0.0](http://semver.org/)
 
-
+###3.5.35
+* Feature 1174: ImageStreams use local lookup policy by default to simplify usage of Deployment or StatefulSet resources on Openshift
 
 ###3.5.34
 * Feature 1003: Added suspend option to remote debugging 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ We use semantic versioning in some slight variation until our feature set has st
 After this we will switch probably to real [Semantic Versioning 2.0.0](http://semver.org/)
 
 ###3.5.35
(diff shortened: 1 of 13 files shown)


rohanKanojia:
Thanks a lot for this :+1: .Could you please update `CHANGELOG.md` also for the upcoming release of fabric8 maven plugin 3.5.35?

nicolaferraro:
It is unlikely to break existing functionality. It could if you name your application with the same `imageId` of an existing "official" image in docker hub (only the ones without group) and you want to use both images in the namespace. e.g. you name your application "mysql" and you want to deploy also a official mysql container in the same namespace.
In that case there's ambiguity on which image corresponds to "mysql:latest" (we use only latest on the imagestream). You can disambiguate using the full name of the docker image: "mysql:latest" is the app, "docker-library:mysql:latest" is the database.

By contrast, if you don't put the default to `true`, when you try to use a "Deployment" or "StatefulSet" (or "CronJob" or anything except "DeploymentConfig"), the application doesn't start and you don't know why. Up to now, such resources worked with f-m-p only in kubernetes mode.

nicolaferraro:
@rohanKanojia I've updated the changelog.

piyush1594:
ok to test

rohanKanojia:
I've fixed the conflict manually and merged this into master, See [fabric8io:master](https://github.com/fabric8io/fabric8-maven-plugin/commits/master). So closing this now.

Thanks

PR #1204, opened 2018-02-20
Bump f-m-p version in parent POM to last released version for docs ge…

oscerd:
…neration

@nicolaferraro @hrishin


nicolaferraro:
Thanks @oscerd. This update should be done every time we do a release..

Issue #1209, opened 2018-02-21
Allow generating service without a route


chmouel:
Currently when generating a service with f-m-p it would generate a Route along the way which is not needed, it would be useful to have a flag so it the Route doesn't get generated along the Service.

piyush1594:
Hey @chmouel 

You need to do some little changes in pom.xml and probably it should work.

```
            <plugin>
                <groupId>io.fabric8</groupId>
                <artifactId>fabric8-maven-plugin</artifactId>
                <version>3.5-SNAPSHOT</version>
                <configuration>
                    <enricher>
                        <excludes>
                            <exclude>f8-expose</exclude>
                        </excludes>
                    </enricher>
                </configuration>
            </plugin>
```

Can you please try this configuration. Please let us know if this doesn't resolve your problem.

Thanks

chmouel:
Ah that's neat but i can't seem to make it works, i have this : 

![image](https://user-images.githubusercontent.com/98980/36486861-dcf2d4a4-171f-11e8-86b8-a91159919f13.png)

but the Route for content-repository still gets generated : 

![image](https://user-images.githubusercontent.com/98980/36486934-03be4a14-1720-11e8-95cc-11c1d76066ca.png)

piyush1594:
Hi @chmouel,

The file you are showing is the resource file which is getting generated. When you hit fabric8 goal, it generates :
- deployment and service for kubernetes.
- deploymentConfig, service, and route for OpenShift

The above configuration I provided will generate all the resource yaml, no change in that behavior but I will not create a route resource on your OpenShift cluster when you hit fabric8: deploy goal. If you hit fabric8: deploy goal without this configuration it will generate a route on your cluster.

Above configuration is wrt to resources on the cluster, not the yaml.

I think this will resolve your problem. Let us know if you have some other doubt.

Thanks

hrishin:
Even making https://github.com/fabric8-services/fabric8-tenant-jenkins/blob/master/apps/content-repository/src/main/fabric8/content-repository-svc.yml#L14 to `false` generating the route resource.

The requirement here is `to not generate route resource` at all.

piyush1594:
As #1211 is merged, I am closing this.

PR #1211, opened 2018-02-23
Adds flag for Route Generation

piyush1594:
if fabric8.openshift.generateRoute is set to false no route.yaml will be generated and if set true or not set it will generate route.yml
#1209

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,16 @@ We use semantic versioning in some slight variation until our feature set has st
 
 After this we will switch probably to real [Semantic Versioning 2.0.0](http://semver.org/)
 
+###3.5.36
+* Fix 1209: Added flag fabric8.openshift.generateRoute which if set to false will not generate route.yml and also will not add Route resource in openshift.yml. If set to true or not set, it will generate rou  te.yml and also add Route resource in openshift.yml. By default its value is true.
+
 ###3.5.35
-* Fix 1130: Added flag fabric8.openshift.trimImageInContainerSpec which would set the container image reference to "", this is done to handle weird 
+* Fix 1130: Added flag fabric8.openshift.trimImageInContainerSpec which would set the container image reference to "", this is done to handle weird
   behavior of Openshift 3.7 in which subsequent rollouts lead to ImagePullErr.
 * Feature 1174: ImageStreams use local lookup policy by default to simplify usage of Deployment or StatefulSet resources on Openshift
 
 ###3.5.34
-* Feature 1003: Added suspend option to remote debugging 
+* Feature 1003: Added suspend option to remote debugging
 * Remove duplicate tenant repos from downstream version updates and add in tjenkins platform
 * Fix 1051: resource validation was slow due to online hosted schema. The fix uses the JSON schema from kubernetes-model project
 * Fix 1062: Add a filter to avoid duplicates while generating kubernetes template(picking the local generated resource ahead of any dependency). Added a resources/ folder in enricher/standard/src/test/ directory to add some sample yaml and jar resource files for DependencyEnricherTest.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ We use semantic versioning in some slight variation until our feature set has st
 
 After this we will switch probably to real [Semantic Versioning 2.0.0](http://semver.org/)
 
-###3.5.36
-* Fix 1209: Added flag fabric8.openshift.generateRoute which if set to false will not generate route.yml and also will not add Route resource in openshift.yml. If set to true or not set, it will generate rou  te.yml and also add Route resource in openshift.yml. By default its value is true.
+###3.5.38
+* Feature 1209: Added flag fabric8.openshift.generateRoute which if set to false will not generate route.yml and also will not add Route resource in openshift.yml. If set to true or not set, it will generate rou  te.yml and also add Route resource in openshift.yml. By default its value is true.
 
 ###3.5.35
 * Fix 1130: Added flag fabric8.openshift.trimImageInContainerSpec which would set the container image reference to "", this is done to handle weird
--- a/doc/src/main/asciidoc/inc/goals/build/_fabric8-resource.adoc
+++ b/doc/src/main/asciidoc/inc/goals/build/_fabric8-resource.adoc
@@ -190,7 +190,7 @@ Resource goal also validates the generated resource descriptors using API specif
 [[resource-route-geneartion]]
 === Route Generation
 
-If you hit fabric8:resource goal, it will also generate route along with service for both kubernetes and Openshift, deployment, deployment Config. If you do not want to generate route.yml, you can do the same using the below configuration.
+If you hit fabric8:resource goal, it will also generate route along with service for OpenShift. If you do not want to generate route.yml, you can do the same using the below configuration.
 
 .Route Generation Configuration
 [cols="1.6.1"]
@@ -203,7 +203,9 @@ If you hit fabric8:resource goal, it will also generate route along with service
 |===
 
 If you do now want to generate route.yml and want to configure it in pom.xml then you can do the same using following configuration in pom.xml
+
 .Example for not generating route resource by configuring it in pom.xml
+
 [source,xml,indent=0,subs="verbatim,quotes,attributes"]
 ----
 <plugin>
@@ -215,6 +217,7 @@ If you do now want to generate route.yml and want to configure it in pom.xml the
 ----
 
 If you are using resource fragments, then also you can configure it in service.yml. You need to add a label `expose` in the metadata of the service and need to set it false.
+
 .Example for not generating route resource by configuring it in resource fragments
 
 [source.yaml]
--- a/doc/src/main/asciidoc/inc/goals/build/_fabric8-resource.adoc
+++ b/doc/src/main/asciidoc/inc/goals/build/_fabric8-resource.adoc
@@ -202,18 +202,30 @@ If you hit fabric8:resource goal, it will also generate route along with service
 | true
 |===
 
-If you want to generate route.yml but do not want to create route resource on your OpenShift Cluster. You can do the same by doing the following configuration in pom.xml
-.Example for not generating route resource on your cluster
+If you do now want to generate route.yml and want to configure it in pom.xml then you can do the same using following configuration in pom.xml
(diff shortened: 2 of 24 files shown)


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
```diff
@@ -92,30 +93,56 @@ protected void writeResources(KubernetesList resources, ResourceClassifier class
         }
     }
 
-    public static File writeResourcesIndividualAndComposite(KubernetesList resources, File resourceFileBase, ResourceFileType resourceFileType, Logger log) throws MojoExecutionException {
+    public static File writeResourcesIndividualAndComposite(KubernetesList resources, File resourceFileBase, ResourceFileType resourceFileType, Logger log, Boolean generateRoute) throws MojoExecutionException {
+
+        List<HasMetadata> oldItemList = new ArrayList<>();
```
pepperbot: MAJOR Remove this useless assignment to local variable "oldItemList". [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1854)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
```diff
@@ -92,30 +93,56 @@ protected void writeResources(KubernetesList resources, ResourceClassifier class
...
+            resources.setItems(newItemList);
+
+        }
+
         Object entity = resources;
+
         // if the list contains a single Template lets unwrap it
         Template template = getSingletonTemplate(resources);
         if (template != null) {
+            System.out.println("In");
```
pepperbot: MAJOR Replace this use of System.out or System.err by a logger. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS106)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
```diff
@@ -92,30 +93,54 @@ protected void writeResources(KubernetesList resources, ResourceClassifier class
...
             String name = KubernetesHelper.getName(item);
             if (Strings.isNullOrBlank(name)) {
                 log.error("No name for generated item %s", item);
                 continue;
             }
             String itemFile = KubernetesResourceUtil.getNameWithSuffix(name, item.getKind());
-            File itemTarget = new File(targetDir, itemFile);
-            writeResource(itemTarget, item, resourceFileType);
+
+            if (!(item.getKind().equalsIgnoreCase("Route") && generateRoute.equals(false))){
```
rohanKanojia: You can have a cleaner check via `item instanceof Route`

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
```diff
@@ -92,30 +93,54 @@ protected void writeResources(KubernetesList resources, ResourceClassifier class
         }
     }
 
-    public static File writeResourcesIndividualAndComposite(KubernetesList resources, File resourceFileBase, ResourceFileType resourceFileType, Logger log) throws MojoExecutionException {
+    public static File writeResourcesIndividualAndComposite(KubernetesList resources, File resourceFileBase, ResourceFileType resourceFileType, Logger log, Boolean generateRoute) throws MojoExecutionException {
+
+        List<HasMetadata> oldItemList = resources.getItems();
```
rohanKanojia: I think possibly we can get rid of this `oldItemList`; Since you're simply modifying resource object in your route generate condition. Or is there something I'm missing here?

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
```diff
@@ -72,11 +73,11 @@ protected static Template getSingletonTemplate(KubernetesList resources) {
         return null;
     }
 
-    protected void writeResources(KubernetesList resources, ResourceClassifier classifier) throws MojoExecutionException {
+    protected void writeResources(KubernetesList resources, ResourceClassifier classifier, boolean generateRoute) throws MojoExecutionException {
         // write kubernetes.yml / openshift.yml
         File resourceFileBase = new File(this.targetDir, classifier.getValue());
 
-        File file = writeResourcesIndividualAndComposite(resources, resourceFileBase, this.resourceFileType, log);
+        File file = writeResourcesIndividualAndComposite(resources, resourceFileBase, this.resourceFileType, log, generateRoute);
```
rohanKanojia: Nitpick: You're passing `boolean` from here which is being received as `Boolean` from `writeResourcesIndividualAndComposite`. I don't have any problem with this but It would be nice if we stay consistent with these.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
piyush1594: Here I am storing the old list because I am deleting the Route for generating openshift.yml. Later I will need all the resources and that's why it is again added to resources [here](https://github.com/fabric8io/fabric8-maven-plugin/pull/1211/files/4774dd63e61ab3c70b64e6e898112ff53346aa12#diff-cc525cbb3a0475474fadc5ce80911c1cR126)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
piyush1594: I think this will not make much difference and here items are the types of HasMetadata which are indirectly every OpenShift resource. Accidentally, we may not end up deleting any other resource, i used kind value which will be unique.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
piyush1594: That's a typo on my end I will fix that

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
rohanKanojia: I think we can use `instanceof` for both Kubernetes and Openshift resources.

piyush1594: Problem is not we can use instance of with both Kubernetes and Openshift. My case is only related to Openshift as Route is only available in OpenShift. We may not end up deleting some other resource I am using very more accurate defined condition where Kind is equal to Route

CHANGELOG.md
```diff
@@ -10,13 +10,16 @@ We use semantic versioning in some slight variation until our feature set has st
 
 After this we will switch probably to real [Semantic Versioning 2.0.0](http://semver.org/)
 
+###3.5.36
```
hrishin: Its going to be 3.5.38, check project release page

doc/src/main/asciidoc/inc/goals/build/_fabric8-resource.adoc
```diff
@@ -185,4 +185,35 @@ Resource goal also validates the generated resource descriptors using API specif
...
+| *fabric8.openshift.generateRoute*
+| If value is set to 'false' then it will not generate the route.yml file. By default it is set to 'true', which will create route.yml and also add resource route in openshift.yml.
+| true
+|===
+
+If you want to generate route.yml but do not want to create route resource on your OpenShift Cluster. You can do the same by doing the following configuration in pom.xml
+.Example for not generating route resource on your cluster
+[source,xml,indent=0,subs="verbatim,quotes,attributes"]
+----
+<plugin>
```
hrishin: Along with this let's also note that, if route and service fragment this exclude goal will not work. 
 
Removing route yml and making  `service` metadata -> label -> expose: "false" will prevent the route generation

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/AbstractResourceMojo.java
```diff
@@ -92,30 +93,69 @@ protected void writeResources(KubernetesList resources, ResourceClassifier class
...
+            //if flag is set false, this will remove the Route resource from resources list
+            for (HasMetadata item : resources.getItems()) {
+                if (item.getKind().equalsIgnoreCase("Route")) {
+                    continue;
+                }
+                newItemList.add(item);
+            }
+
+            //update the resource with new list
+            resources.setItems(newItemList);
```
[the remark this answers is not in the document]
piyush1594: the new item list is just the same old item lits after removing route object if the flag is set to false. then the resource with new item list is getting used in writing openshift.yml and kubernetes.yml.

After that, I am again putting old item list in resource if that is getting used further anywhere then do not modify it and send the original object in function calls.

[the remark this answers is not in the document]
hrishin: I see. I think there should be a `single source of truth` for the resource list. Whatever written to file (ymls) should get process further into pipeline .

samples/spring-boot-without-route/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+                <artifactId>spring-boot-maven-plugin</artifactId>
+            </plugin>
+
+            <plugin>
+                <groupId>io.fabric8</groupId>
+                <artifactId>fabric8-maven-plugin</artifactId>
+                <version>3.5-SNAPSHOT</version>
+
+                <configuration>
+                    <generateRoute>false</generateRoute>
```
hrishin: Nice!

samples/spring-boot-without-route/src/main/deployments/bin/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+ls -ltr /
```
hrishin: remove shell script

piyush1594: done

hrishin:
@piyush1594 LGTM

Issue #1173, opened 2018-02-05
WildFly Swarm based applications do not work FMP


rareddy:
This is a regression of https://github.com/fabric8io/fabric8-maven-plugin/issues/320

I tested with 3.5.34 version of FMP, the last version this worked is with 3.5.25

Ladicek:
We (WildFly Swarm team) are seeing the issue as well, see https://issues.jboss.org/browse/SWARM-1859

As I describe in that issue, I believe this started hapenning in FMP 3.5.34 (i.e. 3.5.33 works fine -- could you verify?), and the reason isn't a regression of #320 (#320 was caused by Jolokia Java agent, and that has been fixed since Jolokia 1.3.6). The reason is actually caused by the Prometheus Java agent, which was newly added in the Fabric8 Java S2I image 2.1.

I believe that until the Prometheus Java agent is fixed (similarly to how the Jolokia Java agent was fixed), FMP should just switch it off by setting the `AB_PROMETHEUS_OFF` environment variable.

Ladicek:
Submitted https://github.com/fabric8io/fabric8-maven-plugin/pull/1216

rhuss:
In order to fix this for Prometheus, the [jmx_exporter](https://github.com/prometheus/jmx_exporter) needs to adapt its initialization behaviour. It suffers from the same problem as Jolokia did since they switched to the JVM internal Java HTTP server which uses jul logging early before the JBoss log subsystem is setup (as jmx_exporter is also designed as  JVM agent which kicks in early).

The fix within Jolokia was quite involved, requiring some internal lifecycle changes, so not sure how easy and, even more important, whether the Prometheus folks are willing to change this for Wildfly. Tbh, I still believe this is a Wildfly issue, not that of the agents. In call always be that any JVM agent (e.g. also profiler) use java.util.logging which will cause this symptoms: Its not correct from Wildfly to assume its the first one hitting java.util.logging. 

I raised these concerns earlier on the original Jolokia related tickets, but they told me, its easier to it is easier for a camel to go through the eye of a needle than to get such a change into Widlfly. So we changed Jolokia, but you still have this issue for any other JVM agent.

Ladicek:
Personally, I'd rather abstain from such discussion, because I can see the pain on both sides. I just think that we should disable the Prometheus agent for WildFly Swarm applications for the time being :-)

rhuss:
np for me ;-) 

[...]

PR #1216, opened 2018-02-27
disable Prometheus and re-enable Jolokia for WildFly Swarm

Ladicek:
The Jolokia agent, which used to have problems with WildFly Swarm due to
how Java LogManager is set, works fine since Jolokia 1.3.6. This means that
the Fabric8 Java S2I image has been OK since version 2.0. Hence I believe
it is safe to re-enable Jolokia for WildFly Swarm applications now.

At the same time, the Fabric8 Java S2I image version 2.1 added a Prometheus
agent, which suffers from the same problem. Hence I believe it should be
disabled for WildFly Swarm applications (for now).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 
 ###3.5.39
 * Feature 1206: Added support for spring-boot 2 health endpoint
+* Fix 1173: disable the Prometheus agent for WildFly Swarm applications, because it uses Java logging too early; also reenable the Jolokia agent, which was disabled due to the same problem but was fixed a long time ago
 
 ###3.5.38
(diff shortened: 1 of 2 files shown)


hrishin:
@Ladicek will review it as soon as possible.

Created issue for the same #1217

Ladicek:
Thanks. I actually believe this would fix https://github.com/fabric8io/fabric8-maven-plugin/issues/1173, but asked the original reporter to check.

Ladicek:
Bump?

hrishin:
LGTM

hrishin:
@piyush1594 @rohanKanojia ?

rohanKanojia:
@Ladicek : Could you please squash your commits into one?

Ladicek:
Oh I intentionally made it 2 commits, as each fixes something different (even if in a single class), but sure, squashed.

piyush1594:
@Ladicek can you please update CHANGELOG.md regarding this PR

Ladicek:
Rebased and added a changelog entry.

piyush1594:
LGTM, merging this.

Ladicek:
Thanks!

PR #1199, opened 2018-02-16
Fixes #1198 Fix CI setup of fabric8-maven-plugin

rohanKanojia:
Since we have extended build timeout of Fabric8 maven plugin on
Travis CI, so reverting back to Travis from CircleCI

PR #1220, opened 2018-03-05
fix: use correct dependency of junit

dipak-pawar:
**Changes Proposed**
* Renamed `VertxHealthchecksBooster` to `VertxHealthchecksBoosterIT` to catch it for failsafe plugin
* Removed unnecessary `testng` dependency, adds `junit` for it.
* Adds test scope for all dependency from `rt/pom.xml`

### DIFF
--- a/rt/src/test/java/io/fabric8/maven/rt/SpringbootHttpBoosterIT.java
+++ b/rt/src/test/java/io/fabric8/maven/rt/SpringbootHttpBoosterIT.java
@@ -16,19 +16,16 @@
 
 package io.fabric8.maven.rt;
 
-import io.fabric8.openshift.api.model.*;
+import io.fabric8.openshift.api.model.Route;
+import java.util.concurrent.TimeUnit;
 import org.apache.http.HttpStatus;
 import org.eclipse.jgit.lib.Repository;
 import org.junit.After;
-import org.testng.annotations.Test;
-
-import java.util.concurrent.TimeUnit;
+import org.junit.Test;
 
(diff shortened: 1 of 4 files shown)


rohanKanojia:
Thanks a lot :+1:

rt/src/test/java/io/fabric8/maven/rt/SpringbootHttpBoosterIT.java
```diff
@@ -16,19 +16,16 @@
...
-
-import java.util.concurrent.TimeUnit;
+import org.junit.Test;
 
 import static io.fabric8.kubernetes.assertions.Assertions.assertThat;
 
-/**
- * Created by hshinde on 11/23/17.
- */
+
```
rohanKanojia: Any specific reason to remove this comment? I think it's perfectly fine to have this.

piyush1594:
Hi @dipak-pawar 

Thanks for the PR, LGTM. Please clarify the concern raised by @rohanKanojia 

We will get this in after proper testing when #1199 will be merged.

Thanks

rt/src/test/java/io/fabric8/maven/rt/SpringbootHttpBoosterIT.java
dipak-pawar: I cannot see it for any other test under `rt/`. I removed it to make it consistent across all tests.

rohanKanojia: Hm, okay. I also don't have any problems with it being gone :-)

PR #1225, opened 2018-03-08
fix 2505 : Pipeline fails to deploy changes to an application

hrishin:
- fixes ImageStream service to pick the version of image SHA to reflect latest version to S2I build

fix https://github.com/openshiftio/openshift.io/issues/2505

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix 1177: Added flag fabric8.openshift.enableAutomaticTrigger which would be able to enable/disable automatic deployments whenever there is new image
   generated.
 * Fix 1184: MultiModule projects were not getting deployed using FMP 3.5.34 onwards. This was working after adding an extra flag which was breaking the previous behaviour in patch release. We will make these change again in minor release and will add the notes regarding that. For the time being, we have reverted the change.
+* Fix #1226: Plugin fails to deploy changes to an application with S2I build. It used to pick first image verion always. This fix pick the right image tag by comaring created attribute of image tag
 
 ###3.5.35
 * Fix 1130: Added flag fabric8.openshift.trimImageInContainerSpec which would set the container image reference to "", this is done to handle weird
--- a/core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
@@ -18,6 +18,8 @@ package io.fabric8.maven.core.service.openshift;
 
 import java.io.File;
 import java.io.IOException;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.*;
 
 import io.fabric8.kubernetes.api.KubernetesHelper;
@@ -42,6 +44,7 @@ import org.apache.maven.plugin.MojoExecutionException;
 public class ImageStreamService {
 
 
+    public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
     private final OpenShiftClient client;
     private final Logger log;
 
@@ -184,6 +187,7 @@ public class ImageStreamService {
 
     private String findTagSha(OpenShiftClient client, String imageStreamName, String namespace) throws MojoExecutionException {
         ImageStream currentImageStream = null;
+
         for (int i = 0; i < IMAGE_STREAM_TAG_RETRIES; i++) {
             if (i > 0) {
                 log.info("Retrying to find tag on ImageStream %s", imageStreamName);
@@ -205,29 +209,62 @@ public class ImageStreamService {
             if (tags == null || tags.isEmpty()) {
                 continue;
             }
-            // latest tag is the first
+
+            // Iterate all imagestream tags and get the latest one by 'created' attribute
+            TagEvent latestTag = null;
+
             TAG_EVENT_LIST:
             for (NamedTagEventList list : tags) {
                 List<TagEvent> items = list.getItems();
-                if (items == null) {
+                if (items == null || items.isEmpty()) {
                     continue TAG_EVENT_LIST;
                 }
 
-                // latest item is the first
-                for (TagEvent item : items) {
-                    String image = item.getImage();
-                    if (Strings.isNotBlank(image)) {
-                        log.info("Found tag on ImageStream " + imageStreamName + " tag: " + image);
-                        return image;
-                    }
+                for (TagEvent tag : items) {
+                    latestTag = latestTag == null ? tag : newerTag(tag, latestTag);
                 }
             }
+
+            if (latestTag != null && Strings.isNotBlank(latestTag.getImage())) {
+                String image = latestTag.getImage();
+                log.info("Found tag on ImageStream " + imageStreamName + " tag: " + image);
+                return image;
+            }
         }
+
         // No image found, even after several retries:
         if (currentImageStream == null) {
             throw new MojoExecutionException("Could not find a current ImageStream with name " + imageStreamName + " in namespace " + namespace);
         } else {
             throw new MojoExecutionException("Could not find a tag in the ImageStream " + imageStreamName);
         }
+    }
+
+    public TagEvent newerTag(TagEvent tag1, TagEvent tag2) {
+        Date tag1Date = extractDate(tag1);
+        Date tag2Date = extractDate(tag2);
+
+        if(tag1Date == null) {
+            return tag2;
+        }
+
+        if(tag2Date == null) {
+            return tag1;
+        }
+
+        return tag1Date.compareTo(tag2Date) > 0 ? tag1 : tag2;
+    }
+
+    private Date extractDate(TagEvent tag) {
+        try {
+            return new SimpleDateFormat(DATE_FORMAT).parse(tag.getCreated());
+        } catch (ParseException e) {
+            log.error("parsing date error : " + e.getMessage(), e);
+            return null;
+        } catch (NullPointerException e) {
+            log.error("tag date is null : " + e.getMessage(), e);
+            return null;
+        }
+
     }
 }
--- a/core/src/test/java/io/fabric8/maven/core/service/openshift/ImageStreamServiceTest.java
+++ b/core/src/test/java/io/fabric8/maven/core/service/openshift/ImageStreamServiceTest.java
@@ -29,6 +29,8 @@ import mockit.Expectations;
 import mockit.Mocked;
 import mockit.integration.junit4.JMockit;
 import org.apache.maven.plugin.MojoExecutionException;
+import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.yaml.snakeyaml.Yaml;
@@ -137,5 +139,33 @@ public class ImageStreamServiceTest {
             .addToTags(list)
             .endStatus()
             .build();
+    }
+
+    @Test
+    public void should_return_newer_tag() throws Exception {
+        // GIVEN
+        ImageStreamService service = new ImageStreamService(client, log);
+        TagEvent oldTag = new TagEvent("2018-03-09T03:27:05Z\n", null, null, null);
+        TagEvent latestTag = new TagEvent("2018-03-09T03:28:05Z\n", null, null, null);
+
+        // WHEN
+        TagEvent resultedTag = service.newerTag(oldTag, latestTag);
+
+        // THEN
+        Assert.assertEquals(latestTag, resultedTag);
+    }
+
+    @Test
+    public void should_return_first_tag() throws Exception {
+        // GIVEN
+        ImageStreamService service = new ImageStreamService(client, log);
+        TagEvent first = new TagEvent("2018-03-09T03:27:05Z\n", null, null, null);
+        TagEvent latestTag = null;
+
+        // WHEN
+        TagEvent resultedTag = service.newerTag(first, latestTag);
+
+        // THEN
+        Assert.assertEquals(first, resultedTag);
     }
 }


core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -230,4 +238,19 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
             throw new MojoExecutionException("Could not find a tag in the ImageStream " + imageStreamName);
         }
     }
+
+    private boolean isDate1AfterDate2(String date1, String date2) {
```
sthaha: Isn't this a static method?

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,23 +209,27 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
             if (tags == null || tags.isEmpty()) {
                 continue;
             }
-            // latest tag is the first
+
+            // Iterate overall tags ang get the latest one by 'created' attribute
```
sthaha: typo ang -> and

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,23 +209,27 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
             TAG_EVENT_LIST:
             for (NamedTagEventList list : tags) {
                 List<TagEvent> items = list.getItems();
                 if (items == null) {
                     continue TAG_EVENT_LIST;
                 }
 
-                // latest item is the first
                 for (TagEvent item : items) {
```
sthaha: Have you considered simplifying the iterations and filtering using java8 stream?

piyush1594:
@sthaha Tests need to be added, I and Rohaan are working on that.

sthaha:
@piyush1594 thank you.

May I also suggesting adding more details in the commit message e.g. what is the problem with the current implementation and how the patch addresses it?

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -230,4 +238,19 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
             throw new MojoExecutionException("Could not find a tag in the ImageStream " + imageStreamName);
         }
     }
+
+    private boolean isDate1AfterDate2(String date1, String date2) {
+        SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+
+        if(Strings.isNotBlank(date1) || Strings.isNotBlank(date2)) {
```
piyush1594: @sthaha @hrishin This statement looks like not correct to me. Maybe I am wrong. Can you please check.

sthaha: Good catch!
Agree, I think we will hit this condition all the time, don't we?

piyush1594: Yaa, I also think the same. Only when both are blank then we are going further. It seems like this to me.

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,23 +209,27 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
                     continue TAG_EVENT_LIST;
                 }
 
-                // latest item is the first
                 for (TagEvent item : items) {
-                    String image = item.getImage();
-                    if (Strings.isNotBlank(image)) {
-                        log.info("Found tag on ImageStream " + imageStreamName + " tag: " + image);
-                        return image;
+                    if(latestTag == null || isDate1AfterDate2(item.getCreated(), latestTag.getCreated())) {
```
piyush1594: @sthaha This always returns the first item and the second function is always returning the false. so no changes after that,

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
rohanKanojia: This check seems redundant to me. Even if `SimpleDateFormat` fails to parse the string; it should throw ParseException.

piyush1594: Format needs to be `yyyy-MM-dd'T'HH:mm:ss'Z'`

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -230,4 +238,19 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
             throw new MojoExecutionException("Could not find a tag in the ImageStream " + imageStreamName);
         }
     }
+
+    private boolean isDate1AfterDate2(String date1, String date2) {
+        SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
```
rohanKanojia: Hmm, the reason why it was giving false all the time was due to the date format we're using in `SimpleDateFormat`; it should be `yyyy-MM-dd'T'HH:mm:ss'Z'`

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -240,16 +240,18 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
     }
 
     private boolean isDate1AfterDate2(String date1, String date2) {
-        SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+        SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
 
-        if(Strings.isNotBlank(date1) || Strings.isNotBlank(date2)) {
-            return false;
-        }
+        if(Strings.isNotBlank(date1) && Strings.isNotBlank(date2)) {
```
hrishin: @piyush1594 Thats perfect! :+1:

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -230,4 +238,21 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
             throw new MojoExecutionException("Could not find a tag in the ImageStream " + imageStreamName);
         }
     }
+
+    private boolean isDate1AfterDate2(String date1, String date2) {
```
sthaha: Also the method doesn't access instance/member variables which is good indication that you may want to turn this into a Comparator .. Please note my java is rusty and dated.

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -230,4 +238,21 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
+        SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
+
+        if(Strings.isNotBlank(date1) && Strings.isNotBlank(date2)) {
```
sthaha: Please choose to exit early to keep the nesting fairly flat e.g.

```java
if (String.isBlank(date1) || String.isBlank(date2) {
    return false;
}

try {
   ..
} catch () {
   ...
  return false
}

File: core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java [hunk 10]
```diff
@@ -205,29 +208,59 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
             if (tags == null || tags.isEmpty()) {
                 continue;
             }
-            // latest tag is the first
+
+            // Iterate overall tags and get the latest one by 'created' attribute
```
sthaha: nitpick: typo overall -> over all

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,29 +208,59 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
-                    String image = item.getImage();
-                    if (Strings.isNotBlank(image)) {
-                        log.info("Found tag on ImageStream " + imageStreamName + " tag: " + image);
-                        return image;
-                    }
+                for (TagEvent tag : items) {
+                    latestTag = latestTag == null ? tag : newerTag(tag, latestTag);
                 }
             }
+
```
sthaha: It feels to me that  this block `if (latestTag != null ...)` should move outside the `for { }`  just like `currentImageStream`

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,29 +208,59 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
+        }
+
+        if(tag2Date == null) {
+            return tag1;
+        }
+
+        return tag1Date.compareTo(tag2Date) > 0 ? tag1 : tag2;
+    }
+
+    private Date extractDate(TagEvent tag) {
```
rohanKanojia: There is an almost similar function [KubernetesResourceUtil.parseDate](https://github.com/fabric8io/fabric8-maven-plugin/blob/master/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java#L622); I think we should use that instead of this; via minor changes of course ;)

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
hrishin: apparently, if we take this out of loop, logic will keep iterating till max iteration even after find the latest tag. Its better to return ith iteration after getting latest tag?

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
hrishin: @rohanKanojia good suggestion. Just wanted to handle `Null` check and string formate is different.

piyush1594:
This LGTM @hrishin, Can we add this to CHANGELOG.MD to keep the track. Thanks

hrishin:
fix #1226

piyush1594:
Lets merge this.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix 1177: Added flag fabric8.openshift.enableAutomaticTrigger which would be able to enable/disable automatic deployments whenever there is new image
   generated.
 * Fix 1184: MultiModule projects were not getting deployed using FMP 3.5.34 onwards. This was working after adding an extra flag which was breaking the previous behaviour in patch release. We will make these change again in minor release and will add the notes regarding that. For the time being, we have reverted the change.
+* Fix #1226: Plugin fails to deploy changes to an application with S2I build. It used to pick first image verion always. This fix pick the right image tag by comaring created attribute of image tag
```
maxandersen: a few typos: *version* and *comparing*

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,29 +209,62 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
                 }
 
-                // latest item is the first
-                for (TagEvent item : items) {
-                    String image = item.getImage();
-                    if (Strings.isNotBlank(image)) {
-                        log.info("Found tag on ImageStream " + imageStreamName + " tag: " + image);
-                        return image;
-                    }
+                for (TagEvent tag : items) {
```
maxandersen: Do we know why latest item is no longer the first ? That seems like a behavior that must have changed.

Also, moving to search through all tags to find the latest based on created image do not really solve the issue  does it ? 
another build could be happening for a PR or similar that would create something even later but is not related to this ongoing build - correct ?

maxandersen: in other words - using created attribute is better than latest tag now; but ultimately it is as broken as it always been. The code relies on there being zero other builds happening at the same time.

maxandersen: in a deterministic flow there should be a passing of a id/label or similar to ensure uniquness.

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
```diff
@@ -205,29 +209,62 @@ private String findTagSha(OpenShiftClient client, String imageStreamName, String
...
                 continue;
             }
-            // latest tag is the first
+
+            // Iterate all imagestream tags and get the latest one by 'created' attribute
+            TagEvent latestTag = null;
+
             TAG_EVENT_LIST:
             for (NamedTagEventList list : tags) {
                 List<TagEvent> items = list.getItems();
```
kpiwko: I would replace this code with:

```
List<TagEvent> items = list.getItems()!=null ? list.getItems() : Collections.EMPTY_LIST;
```
And remove if handling & continue with goto label altogether - if collection is empty, for loop would be noop.

core/src/main/java/io/fabric8/maven/core/service/openshift/ImageStreamService.java
hrishin: Absolutely missed these comments somehow. (_thats bad_)
Creating one more issue and spinning the discussion there.

PR #1238, opened 2018-03-22
Refactor Regression Tests

piyush1594:
Spring boot Configmap Booster in a multimodule project so it required fmp version as a command line argument too
Otherwise, it is fetching the latest version for other modules

### DIFF
--- a/rt/src/test/java/io/fabric8/maven/rt/BaseBoosterIT.java
+++ b/rt/src/test/java/io/fabric8/maven/rt/BaseBoosterIT.java
@@ -192,6 +192,10 @@ public class BaseBoosterIT {
         return repository;
     }
 
+    protected String getCurrentFMPVersion() throws Exception{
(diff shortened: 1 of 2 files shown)


piyush1594:
ok to test

piyush1594:
@rohanKanojia done

rt/src/test/java/io/fabric8/maven/rt/BaseBoosterIT.java
```diff
@@ -192,6 +192,10 @@ protected Repository setupSampleTestRepository(String repositoryUrl, String rela
         return repository;
     }
 
+    protected String getCurrentFMPVersion() throws Exception{
```
rohanKanojia: Actually in this case also we would end up querying project model again and again for fmp's version. I was talking about having it as a member element, like `testsuiteNamespace`

PR #1241, opened 2018-03-23
Fix extra column generated

piyush1594:
This will fix the extra column generated here https://maven.fabric8.io/#compatibility-with-OpenShift-and-Kubernetes


rohanKanojia:
[merge]

Issue #1247, opened 2018-03-26
Unable to build empty project due to artifact-is.yml file not found


nicolaferraro:
### Description
Tracking https://issues.jboss.org/browse/OSFUSE-709.

When trying to install a empty project, the plugin fails with a misleading message.

```
Caused by: org.apache.maven.plugin.MojoExecutionException: Failed to install artifact org.jboss.qe:java-helloworld:yml:is:1.0-SNAPSHOT: /home/nferraro/git/stuff/java-helloworld/target/java-helloworld-is.yml (No such file or directory)
        at org.apache.maven.plugin.install.InstallMojo.execute(InstallMojo.java:153)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:208)
        ... 24 more
Caused by: org.apache.maven.artifact.installer.ArtifactInstallationException: Failed to install artifact org.jboss.qe:java-helloworld:yml:is:1.0-SNAPSHOT: /home/nferraro/git/stuff/java-helloworld/target/java-helloworld-is.yml (No such file or directory)
        at org.apache.maven.artifact.installer.DefaultArtifactInstaller.install(DefaultArtifactInstaller.java:111)
        at org.apache.maven.plugin.install.InstallMojo.execute(InstallMojo.java:145)
        ... 26 more
Caused by: org.eclipse.aether.installation.InstallationException: Failed to install artifact org.jboss.qe:java-helloworld:yml:is:1.0-SNAPSHOT: /home/nferraro/git/stuff/java-helloworld/target/java-helloworld-is.yml (No such file or directory)
        at org.eclipse.aether.internal.impl.DefaultInstaller.install(DefaultInstaller.java:280)
        at org.eclipse.aether.internal.impl.DefaultInstaller.install(DefaultInstaller.java:195)
        at org.eclipse.aether.internal.impl.DefaultInstaller.install(DefaultInstaller.java:152)
        at org.eclipse.aether.internal.impl.DefaultRepositorySystem.install(DefaultRepositorySystem.java:377)
        at org.apache.maven.artifact.installer.DefaultArtifactInstaller.install(DefaultArtifactInstaller.java:107)
        ... 27 more
[...]

PR #1248, opened 2018-03-26
Fix 1247: do not try to install non-existent imagestream yml file

nicolaferraro:
With this change the plugin will not try to install into maven repository the yml imagstream if not generated.
#1247


hrishin:
LGTM

cc: @rohanKanojia @piyush1594

nicolaferraro:
Apart from the issue, I find it useful every time I want the plugin to just deploy some resources from `src/main/fabric8` without building a container image.

I think tests are stuck for other reasons..

piyush1594:
Yaa some issues with test @nicolaferraro , Fixing it here #1253

nicolaferraro:
[merge]

PR #1251, opened 2018-03-26
feat: make  it possible to detect a redhat build of the plugin even if the plugin was repackaged under a different groupid.

chirino:


nicolaferraro:
[merge]

Issue #886, opened 2017-03-15
Introduce "extends" for profiles


rhuss:
would be nice if profiles could extend another profile to avoid repetition e.g. of generators if the profile is only about including certain enrichers.

E.g. for a profile like

```
- name: minimal
   extends: default
   enricher:
     includes:
     - fmp-name
     - fmp-controller
     - fmp-service
     - fmp-image
     - fmp-project
     - fmp-debug
```

one then wouldn't need to repeat all generators as they are inherited from the `default` profile.

christian-posta:
+1 to this!! any chance this is high on anyone's list to implement? :)

rhuss:
Not yet, but if there's demand I can move it up in the backlog. Hope to start fmp works again, now that it's post-summit :)

rhuss:
Sorry, no luck. had to give up on f-m-p, but I still think its a good idea.

PR #1252, opened 2018-03-26
Add enricher for adding space labels to resources

ebaron:
This PR creates a new enricher that applies a label to all resources, which identifies the OSIO space that they belong to. Previously this task was handled by the launcher, which would edit the fabric8-maven-plugin configuration in users' GitHub repositories. See: https://github.com/ebaron/myspringboot/commit/3b8f7ff68d740bf5ee316e5dc68b3751dd0999ea

There are a few problems with this approach. The main issue we're encountering right now, is that Spring Boot quickstarts have moved their fabric8-maven-plugin configuration to a new parent POM. This makes the launcher unable to find/add the space label to the FMP configuration. This causes features that rely on the space label to break: https://github.com/openshiftio/openshift.io/issues/2360. Other problems include requiring the user to maintain their POM file in a certain format, and potential problems if a repository is shared between spaces.

This enricher aims to apply the correct space label to resources without any additional configuration or assistance from the user. It works primarily through a change to fabric8-pipeline-library (https://github.com/fabric8io/fabric8-pipeline-library/pull/390), which reads the space label previously set on the Build resource corresponding to the running pipeline. When the pipeline library invokes the fabric8:resource goal from mavenCanaryRelease, it passes the space name to FMP with the `fabric8.enricher.fmp-space-label.space` system property. This property is read by the enricher and returns a label for all resources.

I have included test cases for this new enricher. I can also add documentation, if you think an enricher is the way to go. I added the new enricher to the list of those enabled by the default profile, I was unsure whether to add it to any other profiles.


hrishin:
Fixes #1264

piyush1594:
@ebaron It will be good if you can add info about the new enricher in the docs of website http://maven.fabric8.io

piyush1594:
LGTM :+1:

ebaron:
@rhuss That sounds fine with me. I made the necessary changes, but I'm not sure how best to define the profile. Ideally, you would want it to include everything in `default` plus the `osio-space-label` enricher. If I do the following:
```
# Includes enrichers only useful to OpenShift.io
- name: osio
  enricher:
    includes:
      - osio-space-label
```
only the osio-space-label enricher is used. I would prefer to avoid duplicating all the enrichers/generators in `default` in this `osio` profile. That seems like it would be difficult to maintain. Do you have any suggestions?

rhuss:
@ebaron sorry, that was planned last year with https://github.com/fabric8io/fabric8-maven-plugin/issues/886 but has not been implemented before I left the fabric8 team.

ebaron:
@rhuss I've revised my PR to put this new enricher under a new `osio` profile. In the absence of a better way to do it, I just copied everything in `default` for the `osio` profile. I added a note under the `default` profile to propagate any changes to the `osio` profile. I also moved the enricher out of the fabric8 enrichers and created a new osio module/category for it.

rhuss:
Sounds good to me. 'hopefully someone will jump on #886 soon, to avoid duplication.

ebaron:
Is it possible for this PR to make it into 3.5.39?

ebaron:
@rhuss @hrishin @piyush1594 Would somebody be able to merge this PR? I do not have access.

ebaron:
@hrishin I have revised the PR to use the "extends" functionality from #886 for the "osio" profile. How does it look?

ebaron:
@hrishin @piyush1594 @rhuss @rohanKanojia Sorry for the mass ping, could someone please take a look at the revised PR with the `osio` profile extending `default`?

I see this has now missed 3.5.39, I have moved the ChangeLog message accordingly. Is there a planned timeframe for the next release? This feature would fix some issues for us in OpenShift.io that we are eager to have fixed. See: https://github.com/openshiftio/openshift.io/issues/2360 https://github.com/openshiftio/openshift.io/issues/2724

Issue #1246, opened 2018-03-26
Wrap docker:start/docker:stop from docker-maven-plugin


giannello:
### Description
According to the documentation, this plugin include docker-maven-plugin for most of the heavy lifting of creating/pushing images. It's a huge benefit, as the existing docker-maven-plugin configuration can be reused for this plugin to work, however the docker:start and docker:stop goals are not working in this context, and fabric8:start/stop are not working as expected, as they need to deploy the application to a cluster to work.
It would be great if the old docker:start/stop/watch could be use with this plugin too, as they are pretty useful for local development/testing.

### Info

* f-m-p version: 3.5.38
* Maven version (`mvn -v`): 3.5.2

rohanKanojia:
Closing it for the moment, feel free to re-open if anyone has any objection. See discussion on #1260

PR #1260, opened 2018-03-29
Wrap docker:start/stop

giannello:
Adds wrappers for d-m-p docker:start and docker:stop goals.
This allows switching from docker-maven-plugin to fabric8-maven-plugin and retaining the capability of launching containers on a local docker instance for integration tests purposes.

The goals can be invoked with `mvn fabric8:local-start` and `mvn fabric8:local-stop` and are by default bound to PRE_INTEGRATION_TEST and POST_INTEGRATION_TEST

Solves #1246

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/LocalStartMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/LocalStartMojo.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package io.fabric8.maven.plugin.mojo.build;
+
+
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+
+/**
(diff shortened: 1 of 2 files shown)


rohanKanojia:
I'm not sure whether this change is required or not(I might be wrong). Can't you just use docker-maven-plugin for that purpose?

giannello:
@rohanKanojia @hrishin @nicolaferraro ping

rohanKanojia:
Unfortunately I don't have much insight about any use cases for this change. Perhaps @rhuss / @nicolaferraro could share their thoughts on it.

rhuss:
Sorry, I can't seriously comment on this as I'm not sure what direction you guys want this plugin to go (i've no insight in the future roadmap of fmp). tbh, personally I think fmp already lost its focus, adding even more goals seems to blur the line even further. The use case might be questionable anyway, because running a container in isolation (without other images) not often makes sense. And when you configure additional images, these will end up in the same pod when the resources are created (e.g. db and app), but started seperately as container with `docker:run`.

The runtime model of dmp and fmp are similar but not the same.
Just my 2 cents...

hrishin:
@giannello I do agree with others members here. Why not include fabric8 docker maven (DMP) plugin in a project and invoke docker:start/stop goals?  Could not work?

giannello:
Fair enough, if you project leads think that this is not useful, feel free to close without merging.
In our specific case, we're heavy d-m-p users, and we figured out that f-m-p could be its natural progress as we move our projects over to kubernetes, due to the automagic yaml generation, and compatibility with the existing d-m-p configuration. However, we still wanted to retain the ability to run containers in docker for testing purposes.
Using d-m-p together with f-m-p requires duplicating a big chunk of configuration in our poms, which we could easily avoid by using that simple wrapper.
Again, feel free to drop if deemed not necessary or out of scope.

rhuss:
@giannello thanks again for your contribution. We are currently in a phase where we are about to consolidate the feature around fabric8-maven-plugin. 

Having a `docker:run` actually in the context of a Kubernetes application doesn't make so much sense as only in rare cases the container could work without any of the services Kubernetes offers (config-maps, secrets, persistent volumes). Even for a service lookup via DNS might work differently.

As we are moving more and more to native configuration formats like Dockerfiles or docker-compose.yml, it will be easier in the future to share configuration between fabric8-maven-plugin and docker-maven-plugin so that for your use case you then could consider enabling both plugins in your pom.xml

Issue #1245, opened 2018-03-26
Use a released version in place of master for boosters


piyush1594:
Right now we are using the master of boosters GitHub for regression rest. This will sometimes break because of changes in master. 

You can check this here https://github.com/fabric8io/fabric8-maven-plugin/blob/master/rt/src/test/java/io/fabric8/maven/rt/SpringbootConfigmapBoosterIT.java#L36

We should always use a released version of booster in regression tests and should do an upgrade of booster version whenever required. We should not always rely on the master. This will help us regularly fixing the regression tests.

This will also help in fixing #1198

piyush1594:
Need your inputs, comments or suggestions @hrishin @rohanKanojia

piyush1594:
I am not sure whether a release will happen for boosters. 

@Ladicek @cescoffier Any inputs on this will be really helpful. Thanks

hrishin:
@piyush1594 I second that. Let's do it. At the same time can we make a plan to move regression test from f8-m-p?

rohanKanojia:
There is already a discussion about separating these tests from fmp : #1155 . Feel free to share your thoughts on it.

piyush1594:
@cescoffier @Ladicek Can you please guide if there are any released version of boosters. If yes, then where are released. From Github, it seems like released happened a long time back. TIA

Ladicek:
Yes, boosters are being released. The last releasees happened a couple of weeks ago. You might also want to take a look at the booster catalog (https://github.com/fabric8-launcher/launcher-booster-catalog) -- whatever is recorded there is available through the Launcher.

piyush1594:
Hey @Ladicek 

Thanks for the help. Just one more thing, if you have an idea, which version (community or redhat) is used for OSIO.

Thanks

piyush1594:
I think its the community version that is getting used in OSIO, not sure.

gastaldi:
@Ladicek @piyush1594 it depends on which UI you are running the tests. The legacy UI in OSIO uses the community versions. In the new UI the user is free to select the version.

piyush1594:
Thanks, @gastaldi @Ladicek Right now I have switched the regressions test for latest released community version. #1262 is the PR which will fix this.

piyush1594:
As #1262 is merged, closing this.

PR #1262, opened 2018-04-03
Fixes use released version of boosters in FMP regression tests

piyush1594:
Fixed use a released version of all the Boosters
in place of always using master.
This will also help in regression tests breaking regularly
#1245

### DIFF
--- a/rt/src/test/java/io/fabric8/maven/rt/SpringbootConfigmapBoosterIT.java
+++ b/rt/src/test/java/io/fabric8/maven/rt/SpringbootConfigmapBoosterIT.java
@@ -21,8 +21,10 @@ import okhttp3.Response;
 import org.eclipse.jgit.lib.Repository;
 import org.json.JSONObject;
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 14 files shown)


rt/src/test/java/io/fabric8/maven/rt/SpringbootConfigmapBoosterIT.java
```diff
@@ -35,6 +34,8 @@
 
     private final String SPRING_BOOT_CONFIGMAP_BOOSTER_GIT = "https://github.com/snowdrop/spring-boot-configmap-booster.git";
 
+    private final String RELEASED_VERSION_TAG = "tags/1.5.10-5";
```
lordofthejars: Couldn't we just put here the number part (`1.5.10-5`) and just append `tags/` inside the `setupSampleTestRepository`? So it is easier to track.

piyush1594:
@lordofthejars Changes done. Thanks for the time.

piyush1594:
I right now just used the last released version of the booster. I am not sure how we should do that.

piyush1594:
@hrishin I just checked with launcher team and answer is use the latest released version of boosters in OSIO.

piyush1594:
@hrishin @lordofthejars @rohanKanojia 

Request you to go through one more round of review. Added the fix for the concern raised by @hrishin 

The versions of Booster used by OSIO are mentioned here. https://github.com/fabric8-launcher/launcher-booster-catalog/blob/master/metadata.yaml and the respective git URL and release branch is mentioned in their respective yaml like https://github.com/fabric8-launcher/launcher-booster-catalog/blob/master/spring-boot/1.5.10-redhat/rest-http/booster.yaml

I have parsed the yaml and used the URL and git tag for regression tests. Right now, they are using the Redhat version as the info I received. This may change in the future. We just need to maintain booster.yaml URL if that changes in any case.

piyush1594:
Using the git URL and tag from these boosters also means we are using the booster versions in sync with OSIO.

PR #1270, opened 2018-04-12
Changes the timestamp tag to latest for Docker Images

### LINKED ISSUE #1093: Revert default image name back to "latest" for snapshots
Using timestamps as tags breaks so often, especially when running `fabric8:resource` and `fabric8:build` in different invocations of Maven (where different timestamps are calculated). Also it turned out that in the recent versions even when using `mvn fabric8:resource fabric8:build` it does work, but for `mvn fabric8:build fabric8:resource` the timestamp of the image is different then created in the resource files.

This is so fragile that at least for the default I would revert back to `%g/%a:%l` in

https://github.com/fabric8io/fabric8-maven-plugin/blob/76ee023bd976b246d64715d6f063fad1a0208718/generator/api/src/main/java/io/fabric8/maven/generator/api/support/BaseGenerator.java#L194

Other issues are the excessive creation of Docker images (which needs to be cleaned up manually). For me updating existing images also works without using timestamps, so I really would reconsider the original use case for using a timestamp as tag. And maybe consider to implement a different solution.

piyush1594:
This will fix the FMP goals broken
Fixes #1093


hrishin:
@piyush1594 could you please fix this? You may like update CHANGELOG.md as well.

rhuss:
Looks good !

As I'm going to create the 3.5 branch tomorrow, lets hold back all 4.0 PRs for now. That will make my life easier ;-) After this (and when we resolved the conflict), we should be fine.

rhuss:
@rohanKanojia a booster check seems to be constantly failing:

```
Results :
Tests in error: 
  VertxHealthchecksBoosterIT.redeploy_vertx_app:80->assertDeployment:101->testHealthChecks:119->assertApplicationRecovery:131->BaseBoosterIT.makeHttpRequest:273 » IO
Tests run: 12, Failures: 0, Errors: 1, Skipped: 0
```

as I can't see how this PR affects the test I'm going to merge it still.

Issue #1132, opened 2017-12-11
build logs and status is not working with OpenShift


hrishin:
### Description
For OpenShiftv3.7 build logs not coming with build goal. It keeps throwing following `backoff` messages 

### Info
-> [INFO] F8: Creating ImageStream http-vertx
-> [INFO] F8: Starting Build http-vertx-s2i
-> [INFO] F8: Waiting for build http-vertx-s2i-1 to complete...
-> [INFO] Current reconnect backoff is 1000 milliseconds (T0)
-> [INFO] Current reconnect backoff is 2000 milliseconds (T1)
-> [INFO] Current reconnect backoff is 4000 milliseconds (T2)
-> [INFO] Current reconnect backoff is 8000 milliseconds (T3)
-> [INFO] Current reconnect backoff is 16000 milliseconds (T4)
-> [INFO] Current reconnect backoff is 32000 milliseconds (T5)

* f-m-p version : v3.5.34 
* Maven version (`mvn -v`) : 
Apache Maven 3.3.9 (NON-CANONICAL_2016-07-01T11:53:38Z_mockbuild; 2016-07-01T17:23:38+05:30)
Maven home: /usr/share/maven
Java version: 1.8.0_131, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-1.b12.fc25.x86_64/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.11.10-200.fc25.x86_64", arch: "amd64", family: "unix"
```

```
* Kubernetes / OpenShift setup and version : v3.7.0
* If it's a bug, how to reproduce : run mvn fabric8:build
* If it's a feature request, what is your use case : issue
* Sample project : https://github.com/fabric8io/fabric8-maven-plugin/tree/master/samples/spring-boot

hrishin:
### Description
For OpenShiftv3.7 build logs not coming with build goal. It keeps throwing following `backoff` messages 

### Info
 java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@399df1a2 rejected from java.util.concurrent.ScheduledThreadPoolExecutor@7c9f68f5[Shutting down, pool size = 1, active threads = 1, queued tasks = 0, completed tasks = 2]
[...]

PR #1147, opened 2017-12-27
Fix #1132 build logs and status is not working with OpenShift

rohanKanojia:
doing watchLog() straightway seems to causing problems on OpenShift 3.7.0,
let's watch over the build pod until it's ready.

PR #1274, opened 2018-04-17
Fix #1263 : Fix fmp's build wait section according to k8-client

### LINKED ISSUE #1263: Problems during s2i build on upgrading kubernetes-client to latest version
**Please consider this whenever upgrading fabric8-maven-plugin's kubernetes-client dependency; Right now fmp uses a pretty old version of kubernetes client 3.0.3.**

I was trying to test something using the latest SNAPSHOT version of kubernetes-client on fabric8 maven plugin. But after configuring fmp to use latest client, I saw this while trying to deploy a simple application in samples/spring-boot:
```
[INFO] Copying files to /home/rohan/work/repos/fabric8-maven-plugin/samples/spring-boot/target/docker/fabric8-maven-sample-spring-boot/latest/build/maven
[INFO] Building tar: /home/rohan/work/repos/fabric8-maven-plugin/samples/spring-boot/target/docker/fabric8-maven-sample-spring-boot/latest/tmp/docker-build.tar
[INFO] F8: [fabric8-maven-sample-spring-boot:latest] "spring-boot": Created docker source tar /home/rohan/work/repos/fabric8-maven-plugin/samples/spring-boot/target/docker/fabric8-maven-sample-spring-boot/latest/tmp/docker-build.tar
[INFO] F8: Using BuildServiceConfig fabric8-maven-sample-spring-boot-s2i for Source strategy
[INFO] F8: Creating ImageStream fabric8-maven-sample-spring-boot
[INFO] F8: Starting Build fabric8-maven-sample-spring-boot-s2i
[INFO] F8: Waiting for build fabric8-maven-sample-spring-boot-s2i-34 to complete...
[INFO] F8: ==================================================================
[INFO] F8: Starting S2I Java Build .....
[INFO] F8: S2I binary build from fabric8-maven-plugin detected
[INFO] F8: Copying binaries from /tmp/src/maven to /deployments ...
[INFO] F8: Checking for SpringBoot archive...
[INFO] F8: Found fabric8-maven-sample-spring-boot-3.5-SNAPSHOT.jar...
[INFO] F8: ... done
[WARNING] Ignoring error for already closed/closing connection
[ERROR] F8: Failed to execute the build [Unable to build the image using the OpenShift build service]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 19.026 s
[INFO] Finished at: 2018-04-04T16:22:15+05:30
[INFO] Final Memory: 52M/471M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.5-SNAPSHOT:build (default) on project fabric8-maven-sample-spring-boot: Failed to execute the build: Unable to build the image using the OpenShift build service: NullPointerException -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

rohanKanojia:
#1263

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
@@ -374,31 +374,30 @@ public class OpenshiftBuildService implements BuildService {
 
     private void waitForOpenShiftBuildToComplete(OpenShiftClient client, Build build) throws MojoExecutionException, InterruptedException {
         final CountDownLatch latch = new CountDownLatch(1);
+        final CountDownLatch readyLatch = new CountDownLatch(1);
         final CountDownLatch logTerminateLatch = new CountDownLatch(1);
         final String buildName = KubernetesHelper.getName(build);
-
-        final AtomicReference<Build> buildHolder = new AtomicReference<>();
+        final String buildPodName = buildName + "-build";
 
         // Don't query for logs directly, Watch over the build pod:
-        waitUntilPodIsReady(buildName + "-build", 20, log);
-        log.info("Waiting for build " + buildName + " to complete...");
-        try (LogWatch logWatch = client.pods().withName(buildName + "-build").watchLog()) {
-            KubernetesClientUtil.printLogsAsync(logWatch,
-                    "Failed to tail build log", logTerminateLatch, log);
-            Watcher<Build> buildWatcher = getBuildWatcher(latch, buildName, buildHolder);
-            try (Watch watcher = client.builds().withName(buildName).watch(buildWatcher)) {
+        Watcher<Pod> buildWatcher = getBuildWatcher(latch, readyLatch, buildPodName);
+        try (Watch watcher = client.pods().withName(buildPodName).watch(buildWatcher)) {
+            log.info("Waiting for build " + buildName + " to complete...");
+            waitForBuildEvent(readyLatch);
+            try (LogWatch logWatch = client.pods().withName(buildName + "-build").watchLog()) {
+                KubernetesClientUtil.printLogsAsync(logWatch,
+                        "Failed to tail build log", logTerminateLatch, log);
                 // Check if the build is already finished to avoid waiting indefinitely
                 Build lastBuild = client.builds().withName(buildName).get();
                 String lastStatus = KubernetesResourceUtil.getBuildStatusPhase(lastBuild);
                 if (Builds.isFinished(lastStatus)) {
                     log.debug("Build %s is already finished", buildName);
-                    buildHolder.set(lastBuild);
                     latch.countDown();
                 }
 
-                waitUntilBuildFinished(latch);
+                waitForBuildEvent(latch);
                 logTerminateLatch.countDown();
-                build = buildHolder.get();
+                build = client.builds().withName(buildName).get();
                 String status = KubernetesResourceUtil.getBuildStatusPhase(build);
                 if (Builds.isFailed(status) || Builds.isCancelled(status)) {
                     throw new MojoExecutionException("OpenShift Build " + buildName + ": " + KubernetesResourceUtil.getBuildStatusReason(build));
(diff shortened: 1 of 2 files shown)


nicolaferraro:
It seems a permutation of old code, why does it fix the error?

rohanKanojia:
Yeah, It's just a result of several hit and trials. I wasn't sure whether it's the right thing that's why I
added you as a reviewer since you wrote this file. I was trying to upgrade to latest kubernetes-client
when I faced this. After looking at the code problem seem to be coming from `waitUntilPodIsReady()` method.
I'm afraid I don't quite have an exact answer to why this is working, but I would explain what I observed
while doing this. We are getting `onClose()` callback from Watcher with null cause before `eventReceived` callbacks(in which `buildHolder` is being set) here : https://github.com/fabric8io/fabric8-maven-plugin/blob/9986dc520d140aeccaf1c2927506d51cb0f27b0e/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java#L471

When I remove that `waitUntilPodIsReady()` method from that wait build section(which I had added in [PR#1147](https://github.com/fabric8io/fabric8-maven-plugin/pull/1147)
, the error is gone but then I get that `container in pod waiting to start` warning. I think here the problem
is that we're watching both `Build Pod`(which I added to fix #1132) and `Build` resource itself. Now by changing
the order we still get `onClose()` callback, but we wait for additional time for build pod in which `eventReceived` callback comes and sets `buildHolder`.

piyush1594:
retest this please

rohanKanojia:
@nicolaferraro : I've realized that the previous approach is not the right solution. It just avoids the error. I've refactored that code to use one watch only for build logs and `Build` resource. Does this change make any sense to you?

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -374,31 +374,30 @@ private Build startBuild(OpenShiftClient client, File dockerTar, String buildNam
...
-        log.info("Waiting for build " + buildName + " to complete...");
-        try (LogWatch logWatch = client.pods().withName(buildName + "-build").watchLog()) {
-            KubernetesClientUtil.printLogsAsync(logWatch,
-                    "Failed to tail build log", logTerminateLatch, log);
-            Watcher<Build> buildWatcher = getBuildWatcher(latch, buildName, buildHolder);
-            try (Watch watcher = client.builds().withName(buildName).watch(buildWatcher)) {
+        Watcher<Pod> buildWatcher = getBuildWatcher(latch, readyLatch, buildPodName);
+        try (Watch watcher = client.pods().withName(buildPodName).watch(buildWatcher)) {
+            log.info("Waiting for build " + buildName + " to complete...");
+            waitForBuildEvent(readyLatch);
```
nicolaferraro: This blocks the thread until the latch is released.

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -447,21 +417,23 @@ private void waitUntilBuildFinished(CountDownLatch latch) {
...
+        return new Watcher<Pod>() {
 
             String lastStatus = "";
 
             @Override
-            public void eventReceived(Action action, Build build) {
-                buildHolder.set(build);
-                String status = KubernetesResourceUtil.getBuildStatusPhase(build);
-                log.verbose("BuildWatch: Received event %s , build status: %s", action, build.getStatus());
+            public void eventReceived(Action action, Pod buildPod) {
```
nicolaferraro: This and the `onClose` method are the points where the latch is released. But does the kubernetes client guarantee that they are called? In the past we added additional because in some scenarios we missed the event when the build was too fast (e.g. the base image does not exist), and the latch was not released.

Plus, the onClose does not mean that the build is finished. It can be fired also when the build is not existent. Previously there was a check on the pod to ensure that we start watching after that the build is started.

rohanKanojia: In my opinion kubernetes-client is ensuring that. In [WatchConnectionManager](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L225 ) during the OkHttp's websocket `onMessage()` callback, we're calling `watcher.eventReceived` in two cases and  `watcher.onClose` in `onClose()`, `onFailure()` and `onClosing()` callbacks.

You're right about onClose, I'm going to close `readlyLatch` in that block also.

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
rohanKanojia: Yes, the intention is to wait until pods gets into `Ready` state so that logs can be queried into it. Since Openshift 3.7 additional pod states:
```
fabric8-maven-sample-spring-boot-s2i-1-build   0/1       Pending   0         0s
fabric8-maven-sample-spring-boot-s2i-1-build   0/1       Init:0/2   0         0s
fabric8-maven-sample-spring-boot-s2i-1-build   0/1       Init:0/2   0         2s
fabric8-maven-sample-spring-boot-s2i-1-build   0/1       Init:1/2   0         3s
fabric8-maven-sample-spring-boot-s2i-1-build   0/1       PodInitializing   0         4s
fabric8-maven-sample-spring-boot-s2i-1-build   1/1       Running   0         5s

```

rohanKanojia:
@nicolaferraro : The root cause of why we're getting `onClose()` callback from Watcher with `null` cause before `eventReceived()` callbacks is due to watcher getting closed here: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L711

This was added to avoid thread leak in WatchConnectionManager, You can have a look at this PR: fabric8io/kubernetes-client/pull/974

nicolaferraro:
I think there's a error in the client, going to open a PR to fix it.

rohanKanojia:
Closing this as actual fix is in fabric8io/kubernetes-client#1074

PR #1281, opened 2018-04-23
Fix #886 : Introduce extends for profiles

### LINKED ISSUE #886: Introduce "extends" for profiles

rohanKanojia:
#886

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
@@ -26,6 +26,8 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.fabric8.maven.core.config.ProcessorConfig;
 import io.fabric8.maven.core.config.Profile;
 import org.apache.maven.shared.utils.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Helper class for dealing with profiles.
@@ -37,7 +39,9 @@ public class ProfileUtil {
 
     private ProfileUtil() {}
 
-    // Alowed profile names
+    private static final Logger log = LoggerFactory.getLogger(ProfileUtil.class);
+
+    // Allowed profile names
     public static final String[] PROFILE_FILENAMES = {"profiles%s.yml", "profiles%s.yaml", "profiles%s"};
 
     // Mapper for handling YAML formats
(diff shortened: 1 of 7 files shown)


nicolaferraro:
LGTM

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
```diff
@@ -62,6 +62,9 @@ public static Profile findProfile(String profileArg, File resourceDir) throws IO
             String profile = profileArg == null ? DEFAULT_PROFILE : profileArg;
             Profile profileFound = lookup(profile, resourceDir);
             if (profileFound != null) {
+                if(profileFound.getParentProfile() != null) {
+                    profileFound = inheritFromParentProfile(profileFound, resourceDir);
```
hrishin: Would be better to add log debug/info statement so that user can get visibility of extended profile is being used

core/src/main/java/io/fabric8/maven/core/util/ProfileUtil.java
```diff
@@ -71,6 +74,18 @@ public static Profile findProfile(String profileArg, File resourceDir) throws IO
...
     }
 
+    private static Profile inheritFromParentProfile(Profile aProfile, File resourceDir) throws IOException {
+        Profile aParentProfile = lookup(aProfile.getParentProfile(), resourceDir);
+        if(aParentProfile != null) {
+            aProfile.setEnricherConfig(ProcessorConfig.mergeProcessorConfigs(aProfile.getEnricherConfig(), aParentProfile.getEnricherConfig()));
+            aProfile.setGeneratorConfig(ProcessorConfig.mergeProcessorConfigs(aProfile.getGeneratorConfig(), aParentProfile.getGeneratorConfig()));
+            aProfile.setWatcherConfig(ProcessorConfig.mergeProcessorConfigs(aProfile.getWatcherConfig(), aParentProfile.getWatcherConfig()));
+        } else {
+            throw new IllegalArgumentException("No profile '" + aProfile.getParentProfile() + "' defined");
```
hrishin: It would be better to log "No parent profile xyz defined". Let it be explicit for user. WDYT?

core/src/test/java/io/fabric8/maven/core/util/ProfileUtilTest.java
```diff
@@ -125,4 +125,17 @@ public void blendProfiles() throws Exception {
         assertFalse(mergeConfig.use("spring.swarm"));
 
     }
+
+    @Test
+    public void testExtends() throws Exception {
```
hrishin: Is it better to rename the test method to `shouldExtend()` ?

rohanKanojia:
@hrishin: I've updated it according to your comments. Please see if I've missed anything.

piyush1594:
I am merging this.

PR #1285, opened 2018-04-26
[java-exec] set permissions on fatjar

### LINKED ISSUE #1282: Permission bits from host system are transferred kept in container, breaking executiion
Currently the Dockerfile just copies the `.jar` artifact from maven into the docker container.
In case the permissions of the artifact do not allow read access for group members the execution fails.
(See reproduction case)

I have found the [permissions settings of the assembly descriptor](https://maven.fabric8.io/#config-image-build-assembly).
However I think that the `.jar` file should be handled directly by `f-m-p`, as it concerns a core aspect of `f-m-p.`

### Info

* f-m-p version : 3.5.38
* Maven version (`mvn -v`) :
```
Apache Maven 3.3.9
Maven home: /usr/share/maven
Java version: 1.8.0_162, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.13.0-37-generic", arch: "amd64", family: "unix"
```
* Kubernetes / OpenShift setup and version : N/A

### Reproduction steps

```sh
$ umask
077
$ mvn install
$ ls ls -hl target/docker/foo/bar/latest/build/maven/baz-0.0.1-SNAPSHOT.jar
-rw------- 1 user users 53M Apr 24 18:27 target/docker/foo/bar/latest/build/maven/baz-0.0.1-SNAPSHOT.jar
$ docker run foo/bar
exec java -javaagent:/opt/agent-bond/agent-bond.jar=jolokia{{host=0.0.0.0}},jmx_exporter{{9779:/opt/agent-bond/jmx_exporter_config.yml}} -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/baz-0.0.1-SNAPSHOT.jar
Error: Unable to access jarfile /deployments/baz-0.0.1-SNAPSHOT.jar

$ docker run foo/bar /bin/sh
sh-4.2$ id
uid=1000(jboss) gid=0(root) groups=0(root),1000(jboss)
sh-4.2$ ls -l /deployments/
-rw------- 1 root root 54540342 Apr 24 16:08 /deployments/baz-0.0.1-SNAPSHOT.jar
^D

$ umask 000
$ mvn install
$ ls ls -hl target/docker/foo/bar/latest/build/maven/baz-0.0.1-SNAPSHOT.jar
-rw-rw-rw- 1 user users 53M Apr 24 18:27 target/docker/foo/bar/latest/build/maven/baz-0.0.1-SNAPSHOT.jar
$ docker run foo/bar
# lots of spring boot output
```

PS: Thank you for `f-m-p`!

t-8ch:
Fixes #1282

If the system running maven has a restrictive umask the generated fatjar may
not be executable inside the container.

Signed-off-by: Thomas Weißschuh <thomas.weissschuh@de.amadeus.com>

### DIFF
--- a/generator/java-exec/src/main/java/io/fabric8/maven/generator/javaexec/JavaExecGenerator.java
+++ b/generator/java-exec/src/main/java/io/fabric8/maven/generator/javaexec/JavaExecGenerator.java
@@ -191,6 +191,7 @@ public class JavaExecGenerator extends BaseGenerator {
                     fileSet.setDirectory(toRelativePath(buildDir, project.getBasedir()));
                     fileSet.addInclude(toRelativePath(fatJar.getArchiveFile(), buildDir));
                     fileSet.setOutputDirectory(".");
+                    fileSet.setFileMode("0640");
                     assembly.addFileSet(fileSet);
                 }
                 assembly.addFileSet(createFileSet("src/main/fabric8-includes/bin","bin","0755","0755"));


t-8ch:
Should I do something about the test failures?
I don't think they are related to the changes here.

rohanKanojia:
@t-8ch : No, I don't think so. They don't seem to be related to your change

rohanKanojia:
Thanks

PR #1296, opened 2018-05-24
Fix 1265: WildFly Swarm health check enricher now supports detecting MicroProfile Health

Ladicek:
The old `monitor` fraction (which is now deprecated) and the new
`microprofile-health` fraction both fortunately use the same default
endpoint for the health check: `/health`. Because of that, the only
change that is really required here is ability to detect presence
of the new fraction.


piyush1594:
ok to test

Ladicek:
Bump, could someone please take a look and megre if OK? Thanks!

piyush1594:
ok to test

Ladicek:
Rebased to resolve a conflict in `CHANGELOG.md`.

Ladicek:
Bump?

Ladicek:
Bump?

PR #1303, opened 2018-06-06
fixed unizp issue

hrishin:
- droped unsed method that coulod cause CEV issue

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/MavenUtil.java
@@ -71,15 +71,6 @@ public class MavenUtil {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


piyush1594:
@rohanKanojia you need to wait for CI to pass so that we are confident enough.

rohanKanojia:
@piyush1594 : This PR is just removing an unused function. I fail to see how can we lose confidence from this.

piyush1594:
I agree that we are removing an unused function but I think we should wait for it. That's the purpose we worked on that.

piyush1594:
It's not about losing confidence. It's about proving in the better way.

rohanKanojia:
Not sure what you mean. I thought the change was trivial enough to be merged. Sorry if I broke standard process. Will take care from next time.

hrishin:
Thanks, @rohanKanojia! Let's adore the process.

PR #1311, opened 2018-06-27
dummy change to test PR build with Maven 3.5.2

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ The full documentation can be found in the [User Manual](http://maven.fabric8.io
 ### fabric8-maven-plugin 3 vs. 2
 
 > This is a complete rewrite of the former fabric8-maven plugin. It does not share the same configuration syntax,
-> but migration should be straight forward - please use the [fabric8:migrate goal from 2.x of the plugin](http://fabric8.io/guide/mavenFabric8Migrate.html).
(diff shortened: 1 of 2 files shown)


piyush1594:
ok to test

apupier:
it has not worked as i expected.
Maven version has been changed in the job config for jenkins but for Travis it is still using Maven 3.3.9 https://travis-ci.org/fabric8io/fabric8-maven-plugin/jobs/397348506#L1454
I don't know how to modify it, I found no reference in codebase to 3.3.9 and [one for 3.5.2](https://github.com/fabric8io/fabric8-maven-plugin/blob/master/.mvn/wrapper/maven-wrapper.properties) (which seems not used in the build then?)

rohanKanojia:
@apupier : Right now we don't specify maven version in these integration tests. I think it can be configured here using option `useMaven3Version`: https://github.com/fabric8io/fabric8-maven-plugin/blob/e7ea8bff53bb7f8035d27b0e587df0810d71b144/rt/src/test/java/io/fabric8/maven/rt/BaseBoosterIT.java#L204-L209

apupier:
I'm closing this issue. it shows that the problem was not in fabric8-maven-plugin itself.
Based on that I was able to find that the issue is related to an incompatibility within the embedded Maven version in Eclipse and fabric8-maven-plugin. I think the issue is more on Eclipse side.
Thanks for your help!

PR #1317, opened 2018-07-04
Fix #1155 : Decouple regression test module from Fabric8 Maven Plugin

### LINKED ISSUE #1155: Decouple regression test module from Fabric8 Maven Plugin
These regression tests were basically written to test build and deployment, redeployment aspects of fabric8 maven plugin. It sounds to have some overlap with integration tests in general for f8m-p. These were written to test overall integration of fabric8 maven plugin with fabric8io. These booster projects are some getting started applications which are on the Openshift.io page. Each test basically clones the repository from Github and test all the deployment/redeployment scenarios of fabric8 maven plugin.

But as pointed out by @rhuss in [this](https://github.com/fabric8io/fabric8-maven-plugin/pull/1115) pull request:
_If I understand right, the regression tests are using various other booster projects which are in turn using f-m-p. This is a circular dependency which must not be there. It's not problem if f-m-p would have external test projects which are under full control of f-m-p and are created only for the purpose of testing f-m-p. However, testing other projects which can be developed independently must not be the scope of tests within f-m-p as they can break the f-m-p build any time (and without even being f-m-p fault). Also, f-m-p must not depend on any project using f-m-p, if you want to keep it as an independent product._

If anything changes in these booster projects, something breaks in these regression tests and whole CI starts breaking for no reason. It has happened two times already. So, It's not a good idea to keep them in fabric8 maven plugin source code. I have the following doubts:

- So even if we want to move this module outside fabric8 maven plugin, where should we keep it? Shall we keep it in fabric8io Github org? Or maybe the org which is maintaining those booster projects?

- How shall they be run? I can't think of any way of integrating them again with CI running on pull requests. Or maybe we can have some job which would run these tests on a daily/weekly basis triggering email alarms if something breaks.

rohanKanojia:
#1155


rhuss:
@rohanKanojia not sure that I'm the right person to review, as I left fmp before the regression tests were introduced and I have no idea of boosters (yet). Maybe someone with an incentive for the booster tests might be a better fit for the review ?

rohanKanojia:
@hrishin @piyush1594 : Could you please share your thoughts on this?

rhuss:
@rohanKanojia Do you already have build up the replacement repo holding the regression tests ? AFAIU this PR is about removing the tests,  right ? For this purpose it looks good to me ;-)

rohanKanojia:
@rhuss : Yes, This is replacement repo: https://github.com/rohanKanojia/fabric8-maven-plugin-rt

rohanKanojia:
@rhuss : Yes, I think we can merge once I resolve conflicts.

rhuss:
Ok, let's get that merged. We always can fix issues later then.

Thanks a lot @rohanKanojia !!

PR #1321, opened 2018-07-16
Fix #1312 : Container name should not be generated from maven group Id

### LINKED ISSUE #1312: Container name should not be generated from maven group Id
### Description

When generating Kubernetes configurations, if username is omitted then maven group Id is inserted instead. Maven group id's often have periods, which are illegal characters in Kubernetes container names. Therefore, defaulting to an unmodified group ID seems like the wrong thing to do. Either remove the periods, or default to something else.

The erroneous function is extractContainerName in KubernetesResourceUtil. Decompiled, roughly:

```
    private static String extractImageUser(String image, MavenProject project) {
        ImageName name = new ImageName(image);
        String imageUser = name.getUser();
        return imageUser != null ? imageUser : project.getGroupId();
    }
```

### Info

* f-m-p version : 3.5.40
* Maven version (`mvn -v`) : Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T11:41:47-05:00)
```

```
* Kubernetes / OpenShift setup and version : minikube version: v0.27.0
* If it's a bug, how to reproduce : create a maven project with a period in the group Id, and configure the plugin as follows:

```
                <configuration>
                    <registry>registry.gitlab.com/deltavstudios/flightcontrol</registry>
                    <images>
                        <image>
                            <!-- I am using a private registry here -->
                            <name>192.168.99.100:32449/%a:%l</name>
                            <build>
                               <!-- whatever -->
                            </build>
                        </image>
                    </images>
                </configuration>
```

rohanKanojia:
#1312

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
@@ -133,6 +133,8 @@ public class KubernetesResourceUtil {
 
     public static final HashSet<Class<?>> SIMPLE_FIELD_TYPES = new HashSet<>();
 
+    public static final String CONTAINER_NAME_REGEX = "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$";
+
     protected static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ssX";
 
 
@@ -402,7 +404,17 @@ public class KubernetesResourceUtil {
     private static String extractImageUser(String image, MavenProject project) {
         ImageName name = new ImageName(image);
         String imageUser = name.getUser();
-        return imageUser != null ? imageUser : project.getGroupId();
+        String projectGroupId = project.getGroupId();
+        if(imageUser != null) {
+            return imageUser;
+        } else {
+            if(projectGroupId == null || projectGroupId.matches(CONTAINER_NAME_REGEX)) {
+                return projectGroupId;
+            }
+            else {
+                return projectGroupId.replaceAll("[^a-zA-Z0-9-]", "").replaceFirst("^-*(.*?)-*$","$1");
+            }
+        }
     }
 
     public static Map<String, String> removeVersionSelector(Map<String, String> selector) {
--- a/core/src/test/java/io/fabric8/maven/core/util/KubernetesResourceUtilTest.java
+++ b/core/src/test/java/io/fabric8/maven/core/util/KubernetesResourceUtilTest.java
@@ -26,6 +26,10 @@ import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesList;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 
+import io.fabric8.maven.docker.config.ImageConfiguration;
+import mockit.Expectations;
+import mockit.Mocked;
+import org.apache.maven.project.MavenProject;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
@@ -44,6 +48,9 @@ import static org.junit.Assert.fail;
 public class KubernetesResourceUtilTest {
 
     private static File fabric8Dir;
+
+    @Mocked
+    final MavenProject project = new MavenProject();
 
     @BeforeClass
     public static void initPath() throws UnsupportedEncodingException {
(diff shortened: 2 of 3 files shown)


rohanKanojia:
@nicolaferraro @hrishin : Could you please review whenever you get time?

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -402,7 +404,17 @@ public static String extractContainerName(MavenProject project, ImageConfigurati
...
-        return imageUser != null ? imageUser : project.getGroupId();
+        String projectGroupId = project.getGroupId();
+        if(imageUser != null) {
+            return imageUser;
+        } else {
+            if(projectGroupId == null || projectGroupId.matches(CONTAINER_NAME_REGEX)) {
+                return projectGroupId;
+            }
+            else {
+                return projectGroupId.replaceAll("[^a-zA-Z0-9]", "");
```
hrishin: the regex `[^a-zA-Z0-9]` would also replace `-` which is desired?

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -402,7 +404,17 @@ public static String extractContainerName(MavenProject project, ImageConfigurati
...
-        return imageUser != null ? imageUser : project.getGroupId();
+        String projectGroupId = project.getGroupId();
+        if(imageUser != null) {
+            return imageUser;
+        } else {
+            if(projectGroupId == null || projectGroupId.matches(CONTAINER_NAME_REGEX)) {
+                return projectGroupId;
+            }
+            else {
+                return projectGroupId.replaceAll("[^a-zA-Z0-9-]", "");
```
rhuss: This replacement still does not ensure that a name does not start or end with a `-`. Ok, a groupId propably never starts with a minus, but maybe it could end with one ?

I would add something like `.replace("^-*(.*)-*$","\\1")` ...

rhuss:
@hrishin if the fix is good enough for you, could you please approve the PR ? thx!

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
rohanKanojia: Sorry, please correct me if I'm wrong. I tested my regex with groupId ending with a `-`, I think it handles. Plus how does your regex trims `.` from groupId? Does `replace` work with regex? I tried your regex with `replaceAll` but it seems to be replacing everything with 1

rhuss: @rohanKanojia I mean to add the regexp I suggest in _addition_ to yours of course (and it was not tested, therefore that _something like_ ...)

```
projectGroupId.replaceAll("[^a-zA-Z0-9-]", "").replaceFirst("^-*(.*)-*$","\\1")
```

The idea is to remove leading and trailing `-` from the name as these are not allowed.

core/src/main/java/io/fabric8/maven/core/util/KubernetesResourceUtil.java
```diff
@@ -402,7 +404,19 @@ public static String extractContainerName(MavenProject project, ImageConfigurati
...
+        String projectGroupId = project.getGroupId();
+        if(imageUser != null) {
+            return imageUser;
+        } else {
+            if(projectGroupId == null || projectGroupId.matches(CONTAINER_NAME_REGEX)) {
+                return projectGroupId;
+            }
+            else {
+                return projectGroupId.replaceAll("[^a-zA-Z0-9-]", "")
+                        .replaceFirst("^-", "\\1")
```
rhuss: Are you sure that this works ? You have no capturing group, so what should `\1` refer to ? Actually you could use `.replaceAll("^-*","")` if you want to do it two steps.

rohanKanojia: I tested it on groupIds like `io.fabric8-test-`, `io.fabric8-test-` and `-io.fabric8-test-`, Sorry I'm not an expert of regexes so might be doing some  mistake. We can replace `-` with any character in the desired container characters range, I just thought replacing it with `1`. Since you also suggested this in previous comment.

I tried your regex also i.e `projectGroupId.replaceAll("[^a-zA-Z0-9-]", "").replaceFirst("^-*(.*)-*$","\\1")` but it seems to be replacing everything with 1

rhuss: No, that `\1` references the content capturing group in the previous expression `(...)`. Its not "1" literally ;). Let me suggest a working solution and push it to your branch ...

rhuss: If you are interested how this works, look for "regular expression capturing group"

core/src/test/java/io/fabric8/maven/core/util/KubernetesResourceUtilTest.java
```diff
@@ -143,6 +150,25 @@ public void invalidExtension() throws IOException {
...
+            result = "fabric8-maven-plugin-dummy";
+        }};
+
+        ImageConfiguration imageConfiguration = new ImageConfiguration.Builder()
+                .name("dummy-image")
+                .registry("example.com/someregistry")
+                .name("test")
+                .build();
+        String containerName = KubernetesResourceUtil.extractContainerName(project, imageConfiguration);
+        assertTrue(containerName.matches(KubernetesResourceUtil.CONTAINER_NAME_REGEX));
```
lordofthejars: assertj is already in classpath why not starting writing assertions using it?

rhuss:
Looks good, thanks a lot ! And sorry for the confusion ...

PR #1325, opened 2018-07-19
feature: Add "path" to webapp generator

### LINKED ISSUE #1276: webapp generator requires <finalName> to be set
### Description

The build must contain a 

```xml
 <build>
    <finalName>${project.artifactId}</finalName>
    ....
```

because otherwise the "webapp" assembly used in the generator doesn't find the war file to add. The generator should be smarter and really consider ${project.build.finalName} for adding the proper war-file

### Info

* f-m-p version : 3.5.38
* Maven version (`mvn -v`) :

```
Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T09:58:13+02:00)
Maven home: /Users/roland/.sdkman/candidates/maven/current
Java version: 1.8.0_171, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home/jre
Default locale: en_DE, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.2", arch: "x86_64", family: "mac"
```

rhuss:
and also fix the assembly to include the proper artefact, regardless
of the 'final' name.

Fixes #1276.

### DIFF
--- a/generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
+++ b/generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
@@ -53,6 +53,9 @@ public class WebAppGenerator extends BaseGenerator {
         // Command to execute. If null, the base image default command is used
         cmd,
 
+        // Context path under which the app will be available
+        path {{ d = "/"; }},
+
         // Ports to expose as a command separated list
         ports;
 
@@ -74,7 +77,8 @@ public class WebAppGenerator extends BaseGenerator {
     @Override
     public List<ImageConfiguration> customize(List<ImageConfiguration> configs, boolean prePackagePhase) {
         if (getContext().getMode() == PlatformMode.openshift &&
-            getContext().getStrategy() == OpenShiftBuildStrategy.s2i) {
+            getContext().getStrategy() == OpenShiftBuildStrategy.s2i &&
+            !prePackagePhase) {
             throw new IllegalArgumentException("S2I not yet supported for the webapp-generator. Use -Dfabric8.mode=kubernetes or " +
                                                "-Dfabric8.build.strategy=docker for OpenShift mode. Please refer to the reference manual at " +
                                                "https://maven.fabric8.io for details about build modes.");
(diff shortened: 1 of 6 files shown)


rhuss:
Added a new feature 'path' which use the root context path as default ("/"). This is also a perfect fit to the generate route on OpenShift (which otherwise wouldn't find the service because it does not know about the context path).

Currently, it doesn't work with the Tomcat base image because of permission problems when overwriting the ROOT dir. This will be fixed asap with https://github.com/fabric8io-images/tomcat/issues/2

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/WebAppGenerator.java
```diff
@@ -130,6 +134,11 @@ private AppServerHandler createCustomAppServerHandler(String from) {
     }
 
     private AssemblyConfiguration createAssembly(AppServerHandler handler) {
+        String path = getConfig(Config.path);
+        if (path.equals("/")) {
+            path = "ROOT";
```
rohanKanojia: Question, Just to help me understand for this fix. So we're validating <finalName> here, right? final name of jar would default to ROOT.war , am I right?

rohanKanojia:
@rhuss : Did you tested it on Openshift/Kubernetes? Actually, I was testing it on Openshift(in `-Dfabric8.mode=kubernetes)` but application pod goes in `ImagePullBackOff` state(Might be some problem with my Openshift)

rhuss:
I tested it with minishift and it worked as expected.

rhuss:
btw, ne Tomcat images has been fixed, released and f-m-p uses them. These images fixes the problem with root context.

I'm going probably to update Jetty and Wildfly Images soon, too, to use the latest upstream servers.

rhuss:
@rohanKanojia if you use openshift, please be sure that the user part of your image is the same as the projectname in OpenShift.

That's what I did in samples/webapp

```
mvn package fabric8:build fabric8:resource fabric8:apply -Dfabric8.generator.webapp.server=jetty -Dfabric8.generator.name="myproject/%a:%l"
```

assuming that I'm connected to a project "myproject". Also note, that `%l` as there are still problems with the default %t which should have been changed already as it really gives you all sort of problems. See #1093 and and #1270

rohanKanojia:
@rhuss : Works, Thanks.

PR #1339, opened 2018-07-27
Add more details info related to liveness and readiness probes HTTPS

dmvolod:
Add more details info related to liveness and readiness probes HTTPS as this info is not clear for docs readers and can be detected in source code only.


dmvolod:
@nicolaferraro , thanks for pointing this. I thought that latest fabric8 plugin supports latest SB only and docs need to backport to oldest one.
I will add different management port properties to docs.

nicolaferraro:
[merge]

nicolaferraro:
Thanks :+1:

PR #1347, opened 2018-08-01
Support repackaged maven plugins for Karaf

nicolaferraro:
Not touching the others, but Karaf has repackaged versions of the plugin, so the detection should be flexible. I've also fixed the detection of the spring-boot maven plugin in the icon enricher.


rhuss:
I'm going to merge the PR now before branching off 3.5.x. Unfortunately it didn't made it into 3.5.41 which was just released two hours before. 

But it's included in 3.5.42 (and 4.0.x)

PR #1355, opened 2018-08-14
Fix #825 : More flexible S2I binary build data format

### LINKED ISSUE #825: More flexible S2I binary build data format
### Description

Currently for an S2I binary build f-m-p sends a `docker-build.tar` as binary input during a binary build. This is the same archive which would be send to a Docker daemon for building an image (and e.g. contains also an (here unused) Dockerfile). 

This works fine with the S2I builder image we have under control (like the FIS images). However other S2I Builder images (like EAP or Wildfly) expect the binary input in a different format. For Java Servlet container (including Java EE server) its the plain artifact (`.war` or probably also `.ear`). 

There should be a way for generators and the build configuration in general to influence this format send to OpenShift. Initially this could be determined by some sort property ("tar", "bin" or so), but could be then even more flexible.

At least for the war generator it should be possible to send the war file as binary to the builder image, because currently the war generation is useless when using s2i.

rohanKanojia:
#825

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
@@ -16,9 +16,7 @@
 package io.fabric8.maven.core.service.openshift;
 
 import java.io.File;
-import java.io.FileWriter;
 import java.io.IOException;
-import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
 import java.util.HashMap;
 import java.util.List;
@@ -39,16 +37,16 @@ import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.LogWatch;
 import io.fabric8.maven.core.config.OpenShiftBuildStrategy;
+import io.fabric8.maven.core.service.BinaryInputArchiveBuilder;
 import io.fabric8.maven.core.service.BuildService;
 import io.fabric8.maven.core.service.Fabric8ServiceException;
-import io.fabric8.maven.core.util.IoUtil;
+import io.fabric8.maven.core.util.FileUtil;
 import io.fabric8.maven.core.util.ResourceFileType;
 import io.fabric8.maven.core.util.kubernetes.KubernetesClientUtil;
 import io.fabric8.maven.core.util.kubernetes.KubernetesHelper;
 import io.fabric8.maven.core.util.kubernetes.KubernetesResourceUtil;
 import io.fabric8.maven.core.util.kubernetes.OpenshiftHelper;
 import io.fabric8.maven.docker.access.AuthConfig;
-import io.fabric8.maven.docker.assembly.ArchiverCustomizer;
 import io.fabric8.maven.docker.assembly.DockerAssemblyManager;
 import io.fabric8.maven.docker.config.AssemblyConfiguration;
 import io.fabric8.maven.docker.config.BuildImageConfiguration;
@@ -87,27 +85,37 @@ public class OpenshiftBuildService implements BuildService {
     private BuildServiceConfig config;
     private RegistryService.RegistryConfig registryConfig;
     private AuthConfigFactory authConfigFactory;
+    private BinaryInputArchiveBuilder binaryInputArchiveBuilder;
 
 
-    public OpenshiftBuildService(OpenShiftClient client, Logger log, ServiceHub dockerServiceHub, BuildServiceConfig config) {
+    public OpenshiftBuildService(OpenShiftClient client, Logger log, ServiceHub dockerServiceHub, BuildServiceConfig config, BinaryInputArchiveBuilder binaryInputArchiveBuilder) {
         Objects.requireNonNull(client, "client");
         Objects.requireNonNull(log, "log");
         Objects.requireNonNull(dockerServiceHub, "dockerServiceHub");
         Objects.requireNonNull(config, "config");
+        Objects.requireNonNull(binaryInputArchiveBuilder, "binaryInputArchiveBuilder");
 
         this.client = client;
         this.log = log;
         this.dockerServiceHub = dockerServiceHub;
         this.config = config;
+        this.binaryInputArchiveBuilder = binaryInputArchiveBuilder;
     }
 
     @Override
     public void build(ImageConfiguration imageConfig) throws Fabric8ServiceException {
         String buildName = null;
         try {
-            ImageName imageName = new ImageName(imageConfig.getName());
+            // Create an empty directory
+            File targetDir = createEmptyTargetDir();
 
-            File dockerTar = createBuildArchive(imageConfig);
+            // Fill the directory with the data in the right directory structure
+            binaryInputArchiveBuilder.createBinaryInput(dockerServiceHub, targetDir, imageConfig);
+
+            // Create the tar archive by tarring the directory
+            File dockerTar = binaryInputArchiveBuilder.getBinaryInputTar() != null ? binaryInputArchiveBuilder.getBinaryInputTar() :  createTar(targetDir);
+
+            ImageName imageName = new ImageName(imageConfig.getName());
 
             KubernetesListBuilder builder = new KubernetesListBuilder();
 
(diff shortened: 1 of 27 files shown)


rhuss:
Awesome, thanks a lot @rohanKanojia ! Unfortunately I only can have a look at it as I'm on PTO for the rest of the week (+ Monday next week). I hope that I can then work fulltime on fmp as the last two weeks were eaten by my other work (we are about to release Fuse 7.1).

sorry for that.

generator/api/src/main/java/io/fabric8/maven/generator/api/Generator.java
```diff
@@ -45,6 +47,8 @@
      * @return list of image configurations
      */
     List<ImageConfiguration> customize(List<ImageConfiguration> existingConfigs, boolean prePackagePhase) throws MojoExecutionException;
+
```
rhuss: Could you please add Javadock to the method ? For at least all interfaces, Javadoc should exist. Would be good for classes, too, but for interfaces its really important to add docs and to keep them updated.

generator/karaf/src/main/java/io/fabric8/maven/generator/karaf/KarafGenerator.java
```diff
@@ -71,6 +82,37 @@ public boolean isApplicable(List<ImageConfiguration> configs) {
                MavenUtil.hasPluginOfAnyGroupId(getProject(), KARAF_MAVEN_PLUGIN_ARTIFACT_ID);
     }
 
+    @Override
+    public BinaryInputArchiveBuilder getBinaryInputArchiveBuilder(final BuildService.BuildServiceConfig config) {
+        return new BinaryInputArchiveBuilder() {
+            @Override
+            public void createBinaryInput(ServiceHub hub, File binaryInputDirectory, ImageConfiguration imageConfiguration) throws Fabric8ServiceException {
+                try {
+                    File targetDirectory = new File(config.getBuildDirectory());
```
rhuss: The code here and in the WebApp generator look quite the same. Shouldn't we extract it into an external class which then is used here and in the Webapp generator ?

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/handler/OpenshiftWildFlyS2IHandler.java
```diff
@@ -0,0 +1,66 @@
...
+package io.fabric8.maven.generator.webapp.handler;
+
+import io.fabric8.maven.core.config.PlatformMode;
+import io.fabric8.maven.core.util.MavenUtil;
+import org.apache.maven.project.MavenProject;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class OpenshiftWildFlyS2IHandler extends AbstractAppServerHandler {
```
rhuss: Not really sure about the name. What's the difference the WildFlyAppServerHandler ?

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/handler/OpenshiftWildFlyS2IHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    public OpenshiftWildFlyS2IHandler(MavenProject project) {
+        super("tomcat", project);
```
rhuss: "tomcat" sounds fishy here.

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/handler/OpenshiftWildFlyS2IHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public boolean isApplicable() {
+        if(PlatformMode.isOpenShiftMode(project.getProperties())) {
+            return hasOneOf("**/META-INF/context.xml") ||
+                    MavenUtil.hasPlugin(project, "org.apache.tomcat.maven:tomcat6-maven-plugin") ||
```
rhuss: why is a tomcat maven plugin an indicator for this wildfly handler ?

parent/pom.xml
```diff
@@ -73,6 +73,7 @@
     <!-- Upstream -->
     <version.image.java.upstream.docker>1.5</version.image.java.upstream.docker>
     <version.image.java.upstream.s2i>2.3</version.image.java.upstream.s2i>
+    <version.image.webapp.upstream.s2i>latest</version.image.webapp.upstream.s2i>
```
rhuss: we shouldn't use latest in our deps if not for a very good reason. Can't we stick to a version, like for the other images ?

rohanKanojia: not sure, there doesn't seem to be any other tag here: https://hub.docker.com/r/openshift/wildfly-101-centos7/tags/

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -144,49 +152,26 @@ public void build(ImageConfiguration imageConfig) throws Fabric8ServiceException
...
-                        return tarArchiver;
-                    }
-                };
-            } else {
-                return null;
-            }
-        } catch (IOException e) {
-            throw new Fabric8ServiceException("Unable to add environment variables to the S2I build archive", e);
-        }
+        File tarFile = File.createTempFile(config.getArtifactId().toString(), ".tar");
```
rhuss: Not sure if I understand this: `tarFile` is a temporary tar filename but the tar archive itself is created with the `destFile` as name. The return value is then a the temporary name (which doesn't point to any tar archive if I understood the code right). Could you please check this again ?

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -144,49 +152,26 @@ public void build(ImageConfiguration imageConfig) throws Fabric8ServiceException
...
+        File destFile = new File(targetDir, config.getArtifactId() + ".tar");
+        tarArchiver.setDestFile(destFile);
```
rhuss: Also not sure that this works if you put the destination file into the directory that you tar up (as it would be tar up itself, too)

rhuss: Naiively I would expect a

```
tarArchiver.setDestFile(tarFile);
tarArchiver.createArchive();
return tarFile;
```

and removing `destFile` ....

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -87,27 +85,37 @@
...
     @Override
     public void build(ImageConfiguration imageConfig) throws Fabric8ServiceException {
         String buildName = null;
         try {
-            ImageName imageName = new ImageName(imageConfig.getName());
+            // Create an empty directory
+            File targetDir = createEmptyTargetDir();
+
+            // Fill the directory with the data in the right directory structure
+            binaryInputArchiveBuilder.createBinaryInput(dockerServiceHub, targetDir, imageConfig);
```
rhuss: What's about adding a `File` return value to the `BinaryInputArchiveBuilder` which then holds the the TAR archive ? 

Advantages:

* Get rid of `createTar()` method
* Simplify interface to one method
* No need for the `targetDir` parameter. This just would be an implementation details.

If you don't mind I would change it like this and push it to your branch (just to speed up things and to clarify my point).

@rohanKanojia would this be ok for you ?

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/handler/TomcatAppSeverHandler.java
```diff
@@ -34,9 +35,13 @@ public TomcatAppSeverHandler(MavenProject project) {
 
     @Override
     public boolean isApplicable() {
-        return hasOneOf("**/META-INF/context.xml") ||
-                MavenUtil.hasPlugin(project, "org.apache.tomcat.maven:tomcat6-maven-plugin") ||
-                MavenUtil.hasPlugin(project, "org.apache.tomcat.maven:tomcat7-maven-plugin");
+        if(!PlatformMode.isOpenShiftMode(project.getProperties())) {
```
rhuss: What is the motivation of this check ?

generator/webapp/src/main/java/io/fabric8/maven/generator/webapp/handler/WarS2IHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class WarS2IHandler extends AbstractAppServerHandler {
+    public WarS2IHandler(MavenProject project) {
+        super("webapp", project);
+    }
+
+    @Override
+    public boolean isApplicable() {
+        if(PlatformMode.isOpenShiftMode(project.getProperties())) {
+            return hasOneOf("**/META-INF/context.xml") || project.getPackaging().equals("war");
```
rhuss: Ok, got it. You have a dedicated handler only for s2i build ?. But shouldn't we just check then for the packaging ? Not sure why a JAR archive with a "**/META-INF/context.xml" should be considered as a war file.

rhuss:
@rohanKanojia Thanks a lot for your changes which adhere to the comments I did. However while reviewing and some minor refactorings, I thought again about this PR and I now believe, that a Generator should decide actually be responsible for the creation of the binary input.

Why? Well, because a generator is only about creating the _configuration_ of an image which then can be created in multiple ways (docker, s2i binary, s2i docker, ...). So the generator should be agnostic of the concrete creation of the image.

What now? The generator selects a base image and the base image is actually that part which in the s2i binary input case determines the format of the binary input. So my suggestion would be to either examine the generated images for the 'from' images or evaluate a certain meta-data (label) which could be set by the generator. Then a kind of 'BinaryBuildSelector' would examine this base image and creates the binary input which fits this s2i builder image. 

Does this make sense ? If so, I would like to postpone this story for the next sprint implement it in a generator independent way.

rohanKanojia:
@rhuss : ah, okay

Issue #1271, opened 2018-04-12
CronJob creation fails during fabric8:apply goal


smoyer64:
FMP accepts a manifest (or template) for a ``kind: CronJob`` but fails to deploy it to the Kubernetes cluster.  Apparently the kubernetes-client library used by FMP doesn't support ``CronJob`` yet, so while FMP processes it, pushing it to the cluster causes the following error:

``` (text)
[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.5.38:apply (deploy) on project academics-student-importer: Failed to create CronJob from kubernetes.yml. io.fabric8.kubernetes.client.KubernetesClientException: No handler found for object:CronJob(
<the generated manifest>), status=null, additionalProperties={}) -> [Help 1]
```

See https://github.com/fabric8io/kubernetes-client/issues/1018 for the kubernetes-client roadmap

smoyer64:
Here's the (slightly anonymized) YAML for several of our projects:

```yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: application-importer
spec:
  schedule: "*/5 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: application-importer
            image: application-importer:${timestamp}
            imagePullPolicy: IfNotPresent
            envFrom:
            - configMapRef:
                name: application-config
            - secretRef:
                name: application-secret
          restartPolicy: Never
```

rohanKanojia:
@smoyer64 : okay, I see you're using `v1beta1`. Cool ;-) . I think we'll fix this change before next release.

rohanKanojia:
Since fmp now has latest kubernetes client i.e v4.0.4, this issue can be closed ;-)

PR #1361, opened 2018-08-21
Fix #1308 : upgrade kubernetes client to support OpenShift 3.9 and Kubernetes 1.8/1.9

### LINKED ISSUE #1308: upgrade kubernetes client to support OpenShift 3.9 and Kubernetes 1.8/1.9
### Description
Upgrade kubernetes client to support Kubernetes 1.8/1.9 and OpenShift 3.9

cc: @rohanKanojia @piyush1594

rohanKanojia:
Fixes #1308 #1271


core/src/test/java/io/fabric8/maven/core/handler/ReplicationControllerHandlerTest.java
```diff
@@ -111,7 +111,7 @@ public void replicationControllerHandlerTest() {
         assertFalse(replicationController.getSpec().getTemplate().getSpec().getVolumes().isEmpty());
         assertEquals("test",replicationController.getSpec().getTemplate().getSpec().
                 getVolumes().get(0).getName());
-        assertEquals("/test/path",replicationController.getSpec().getTemplate()
+        assertEquals("path",replicationController.getSpec().getTemplate()
```
rhuss: Are you sure that you want to change the assert only ?

I mean, in `before()` you set up the path to "/test/path", so I'm not sure why you would expect a relative directory now in the generated objects.

rohanKanojia: Oops, I interchanged parameters here ;-) . Thanks for noticing :smile: 
https://github.com/fabric8io/fabric8-maven-plugin/blob/d9d74f21c2289d0c0f85ec0b4f37d65869553b58/core/src/main/java/io/fabric8/maven/core/config/VolumeType.java#L32

PR #1366, opened 2018-08-27
Fix #894: Service with NodePort always gets a new port after 'fabric8:apply'

### LINKED ISSUE #894: Service with NodePort always gets a new port after 'fabric8:apply'
Actually this happens always, but can have two causes, both resulting in the existing service and the service to apply to be considered different [here](https://github.com/fabric8io/fabric8/blob/e7e0ff8c38cad2b818a8cfba1c9f9a629e57f796/components/kubernetes-api/src/main/java/io/fabric8/kubernetes/api/UserConfigurationCompare.java#L49). The difference is always in the MetaData as checked [here](https://github.com/fabric8io/fabric8/blob/e7e0ff8c38cad2b818a8cfba1c9f9a629e57f796/components/kubernetes-api/src/main/java/io/fabric8/kubernetes/api/UserConfigurationCompare.java#L164-L167)

* When `exposecontroller` is used, then the exposecontroller adds two extra annotation (timestamp and a version). This make the annotation part different as these field obviously are missing from kubernetes.yml and so a [replacement of the service happens](https://github.com/fabric8io/fabric8/blob/e7e0ff8c38cad2b818a8cfba1c9f9a629e57f796/components/kubernetes-api/src/main/java/io/fabric8/kubernetes/api/Controller.java#L389)
* When no `exposecontroller` is used and no namespace is set directly in kubernetes.yml, then both objects are also considered to be different, since the namespace gets set to the default namespace by kubernetes (so that this field is never empty).

The second issue can be easily solved by ignoring the namespace if set to null when doing the comparisation.

For the first issue, I can't see any easy solution. @davsclaus @jstrachan @rawlingsj @iocanel any idea how to fix this (without hardcoding the exposecontroller annotations into this check ?)

rohanKanojia:
#894 
+ Whenever we detect that a resource with a certain name exists, edit()
  instead of doing replace()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix 1346: karaf-maven-plugin is detected under any groupId
 
 ###4.0-SNAPSHOT
+* Feature: Move to Java 1.8 as minimal requirement
 * Refactor 1344: Removed unused Maven goals. Please contact us if something's missing for you.
 * Refactor 949: Remove dependency from fabric8/fabric8
 * Feature 1214: Don't use a default for skipBuildPom. This might break backwards compatibility, so please specify the desired value in case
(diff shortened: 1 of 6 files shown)


rhuss:
Thanks a lot ! Let me review this tomorrow to give you feedback.

PR #1368, opened 2018-09-03
Fix url for listing of resource kind mappings

agriesbaum:
Path to util class changed but docs weren't updated.


rhuss:
Let's merge this, thanks @agriesbaum for the heads up !

For a future solution I opened #1372 to make it much easier to maintain this mapping.

PR #1371, opened 2018-09-06
chore: Update samples + add thorntail sample

rhuss:


rhuss:
Let's merge it. its not crucial.

Issue #1330, opened 2018-07-23
Remove fabric8-console specific enrichment


rhuss:
### Description

Actually a lot of annotations are added which make sense only for the fabric8-console. Depending on whether the fabric8-console is still supported/maintained, these annotations should be remvoed. At least they should be moved out of the default profile.

rhuss:
Looks like that everything is already done, a full text search only reveals some doc places where the console is mentioned. 

Before releasing 4.0 we might consider getting rid of the deprecated enrichers to save us some time in refactoring and cleanup.

Issue #1375, opened 2018-09-11
Allow deep configuration for enrichers


rhuss:
### Description

Currently only flat configuration is supported by default (i.e. a simple key-value map). However, there are already enrichers (e.g. the `VertxHealthCheckEnricher`) which examines the plugin configuration on is own by parsing the XML. However those enrichers are tightly coupled to maven and can't easily reused outside a Maven context.

This should not be required, but the enrichers should support nested configurations by default, which actually should be a plain map of maps with plain string values (could also contain lists).

lordofthejars:
@rhuss You reopened this issue to improve how we can pass the custom configuration to an enricher. We didn't talk about but what could be this improvement.  Do you have for example something in mind like an xpath for objects, so you say I want to attribute as `string` `/a/b/c` and the method takes the map and navigates through it, or you are thinking in using for example reflection to fill up a POJO?

PR #1376, opened 2018-09-11
refactor(enricher): Cleanup up fabric8 specific enrichers

rhuss:
Enrichers which are specific to the fabric8 platform (CD + UI) has been
moved into a "deprecated" enricher module. This can be kept as reference
or later easily removed. No enricher in this module is include in any
default profile and hence is disabled.

The enrichers that were disabled that way are:

* CdEnricher
* DocLinkEnricher
* GrafanaLinkEnricher
* IconEnricher

Some other refactoring consolidates on the naming of enrichers, so that
everything with a `f8-` prefix currently is about auto detection of
specific tech stacks (like specialized healthchecks or the prometheus
enricher)

All other are in the "standard" module and have a "fmp" prefix.

The MavenScmEnricher has been simplified a bit to remove duplicated SCM
information ("connection" and "developerConnection")

All integration tests have been adapted accordingly.


rhuss:
There is still one thing which I would like to improve in this PR: Currently in the "fabric8" enricher module, there are mostly healthcheck enrichers, one for each tech stack.

I'd like to move these health check enrichers to something more general like e.g. a "SpringBootEnricher" which is responsible for all Spring Boot specific enrichment (which currently is only for health checks). This would have the benefit to reduce the name of the enricher as it is also used in the configurtion to "spring-boot". 

For this to work properly though, we should be able to allow nested configurtion from the very beginning so that a configuration for such an enricher could look like:

```xml
<enricher>
  <config>
    <spring-boot>
      <health>
         ...
      </health>
    </spring-boot>
  </config>

</enricher>

```

This could also map nicely to properties like it is done in the vertx-healtcheck enricher.

See also #1375 for this deep configuration aspect.

Of course, this probably would make sense for genertors, too.

rhuss:
Fixes #1330

doc/src/main/asciidoc/inc/_enricher.adoc
```diff
@@ -754,7 +730,7 @@ For example:
                 <path>/ping</path>
                 <port>8080</port>
             </readiness>
-        </vertx-health-check>
+        </f8-healthcheck>
```
rohanKanojia: Shouldn't this be f8-healthcheck-vertx?

Issue #1372, opened 2018-09-07
Filename to type mappings should be more flexible than a string array


rhuss:
### Description
Currently we have all mapping of resource kinds to file name extensions in a static String array: https://github.com/fabric8io/fabric8-maven-plugin/blob/5e52eded75357bc8df3ac4ce93e931bd39c75e23/core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java#L198-L244

Ideally, this should be externalized into a text file.

My idea is to move it into an .adoc file, which contains this as a table. This file then could be

* included by the documentation with a usual asciidoc include
* parsed by the code to build up the map type to kind dynamically during startup of the plugin

That way, the mapping would be directly documented in the docs and never be out of date. Plus it can be super easily updated from outside the Java code.

PR #1381, opened 2018-09-13
fix(#1372): Set Kind/Filename mapping as AsciiDoc file

lordofthejars:
The mapping between Kind and Filename is moved to an AsciiDoc file, so it can be reused in documentation as well.

By default, it loads this mapping from `/fabric8/default-kind-filename-mapping.adoc` classpath location, but using env variable or system property named `fabric8.mapping` you can specify your own file.

Initially will inspect classpath and if it returns null, then it will treat `fabric8.mapping` value as a disk location.


lordofthejars:
Yes, I'll add the line in CHANGELOG.

Regarding this, I think we could create some method that automatically updates the `CHANGELOG` and about licenses, I have found some files with and without the headers, but at the same time, this could be also automatized. 

So for now, I will do it manually but it will be great to think about some kind of automation of the PR.

rhuss:
We could consider using [commitizen](https://commitizen.github.io/cz-cli/) for creating the [changelogs](https://github.com/conventional-changelog/conventional-changelog), but it requires quite some discipline when doing commit messages (and also every PR needs to be squashed). 

We tried that for Syndesis, but ended up to manage the changelogs manually again, because at the end it doesn't matter whether you have to remind people to add to the Changelog or whether to remind them to change the commit messages. Manually editing the Changelog also give you a better chance for rewording, as the changelog is more target to an end user than to a developer (like commit messages do).

tbh, I wouldn't bother yet with auto generating of changelogs yet.

Checking license files is something different and we should add the license plugin again.

core/src/main/resources/fabric8/default-kind-filename-mapping.adoc
```diff
@@ -0,0 +1,138 @@
+|===
+|Kind |Filename
```
rhuss: The first column is ist the filename part indicating the kind, the second then is the kind.

Maybe `Filename` is also a bit misleading as one could think about this as a whole file name but its only the last part before the extension. Hard to find a good name here, as the extension is obviously ".yml", so it something like a 'file indicator' or so. wdyt ?

core/src/test/java/io/fabric8/maven/core/util/AsciiDocParserTest.java
```diff
@@ -0,0 +1,99 @@
...
+        + "|cm" + System.lineSeparator()
+        + "a|ConfigMap" + System.lineSeparator()
+        + System.lineSeparator()
+        + "|cronjob" + System.lineSeparator()
+        + "|CronJob" + System.lineSeparator()
+        + "|===";
+
+    private static final String NONE_END_VALID_TABLE = "cols=2*,options=\"header\"]" + System.lineSeparator()
+        + "|===" + System.lineSeparator()
+        + "|Kind" + System.lineSeparator()
```
rhuss: For consistencies sake, I would change this header, too (also it doesnt really matter in the test)

core/src/main/resources/fabric8/default-kind-filename-mapping.adoc
```diff
@@ -0,0 +1,138 @@
+|===
```
rhuss: As this file will be updated when a new resource is to be added, we should add an asciidoc comment here to explain what can be update how.

* don't add extra newlines before the new column
* E.g. no multiline columns are allowed (each column must be an extra line)
* Keep Kind, filenames in a sorted order

etc.

core/src/main/resources/fabric8/default-kind-filename-mapping.adoc
rhuss: Tbh, I wouldn't change the header but the order of the columns because I expect users to look up which file indicator is used for which Kind, so look up the Kind (e.g. 'ConfigMap') to find out what to use in the fragment file (`-cm.yml`). As the table is supposed to be ordered alphabetically in the first column, it would be awesome if the kind would be in the first column.

I know this required some code and test changes, but I believe it would be worth before go out in the wild with it.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KindFilenameMapperUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class KindFilenameMapperUtil {
+
+    public static String[] loadMappings() {
+
+        final String location =
+            EnvUtil.getEnvVarOrSystemProperty("fabric8.mapping", "/fabric8/default-kind-filename-mapping.adoc");
```
rhuss: We already store some meta data in the classpath, but in `META-INF/fabric8` (like the `enrichers-default`). Could we switch the location to `META-INF/fabric8`, too ? And maybe use the `kind-fileindicator-mapping-default.adoc` as name ? (to adhere to the current convention to add `-default` towards the end).

core/src/main/resources/fabric8/default-kind-filename-mapping.adoc
lordofthejars: In the original solution, https://github.com/fabric8io/fabric8-maven-plugin/blob/5e52eded75357bc8df3ac4ce93e931bd39c75e23/core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java#L198-L244 first was the "short" version and second the "long" version so what you suggest here is to leave the header and change the columns, ok, it should not be hard at all.

About the name, how about `filename-suffix`?

rhuss: > About the name, how about `filename-suffix`?

Sound goods. We could also call it 'filename type' (not sure if 'suffix' pronounces the location of this type a bit too much). But never mind, we can change that easily later when we found something better.

core/src/main/java/io/fabric8/maven/core/util/AsciiDocParser.java
```diff
@@ -0,0 +1,156 @@
...
+
+        if (column != null && column.startsWith(END_TABLE)) {
+            return END_TABLE;
+        }
+
+        if(column == null || !column.isEmpty()) {
+            throw new IllegalArgumentException(String.format("Trying to read an empty line for end of row, but content %s was found or EOF", column));
+        }
+
+        return column.trim();
```
rhuss: I guess you could return a plain "" directly, otherwise the line above would have thrown an exception anyway.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KindFilenameMapperUtil.java
```diff
@@ -0,0 +1,49 @@
...
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private static InputStream loadContent(String location) {
+        InputStream resourceAsStream = KindFilenameMapperUtil.class.getResourceAsStream(location);
+
+        if (resourceAsStream == null) {
+            throw new IllegalArgumentException(String.format("%s cannotbe foind in classpath", location));
```
rhuss: typo: "cannotbe foind"

core/src/main/java/io/fabric8/maven/core/util/AsciiDocParser.java
```diff
@@ -0,0 +1,156 @@
...
+            throw new IllegalArgumentException("Trying to read a column but white line or EOF was found.");
+        }
+
+        int separator;
+        if ((separator = column.indexOf("|")) < 0) {
+            throw new IllegalArgumentException(String.format("Expected the initial of a column with (|) but %s found.", column));
+        }
+
+        return column.trim().substring(separator + 1)
+            .replaceAll("[`_*]", "");
```
rhuss: I would put another trim after the `replaceAll` to catch also things like `* dc*, ..`. Not very likely but I think would be a quick win.

core/src/test/java/io/fabric8/maven/core/util/AsciiDocParserTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+public class AsciiDocParserTest {
+
+    private static final String VALID_TABLE = "cols=2*,options=\"header\"]" + System.lineSeparator()
+        + "|===" + System.lineSeparator()
+        + "|Filename" + System.lineSeparator()
+        + "|Kind" + System.lineSeparator()
```
rhuss: would exchange the headers here.

core/src/test/java/io/fabric8/maven/core/util/AsciiDocParserTest.java
rhuss: obsolete as we decided to go for Kind, Filename.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
```diff
@@ -182,64 +182,29 @@ public boolean accept(File dir, String name) {
...
+
+            final List<String> filenameTypes = entry.getValue();
+            final String kind = entry.getKey();
+            for (String filenameType : filenameTypes) {
+                FILENAME_TO_KIND_MAPPER.put(filenameType, kind);
+            }
+
+            // In previous version, last one wins, so we do the same.
+            KIND_TO_FILENAME_MAPPER.put(kind, filenameTypes.get(filenameTypes.size() - 1));
+
```
rhuss: yeah, we should also document this.

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
+|===
```
rhuss: We should add a documentation in the header (as mentioned in a previous comment) and we should add that the canonical name should be the last part of the mapping (which I would say is the full kind in all lowercase).

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+|Kind |Filename Type
+
+|ConfigMap
+a|`cm`, `configmap`
+
+|CronJob
+a|`cronjob`, `cj`
```
rhuss: "cronjob" should go last

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|ClusterRole
+a|`cr`, `crole`, `clusterrole`
+
+|CustomResourceDefinition
+a|`crd`
```
rhuss: would add a "customresourcedefinition" although its quite lengthy.

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|ClusterRoleBinding
+a|`crb`, `clusterrb`
+
+|Deployment
+a|`deployment`
+
+|ImageStream
+a|`is`
```
rhuss: \+ "imagestream"

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|ImageStreamTag
+a|`istag`
```
rhuss: + "imagestreamtag"

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|Job
+a|`job`
+
+|LimitRange
+a|`lr`, `limitrange`
+
+|Namespace
+a|`ns`, `clusterrb`
```
rhuss: hmm. not really sure what "clusterrb" should be and why we allow a namespace resoure at all. I know its not part of the PR, I just thinking out loud ;-)

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|OAuthClient
+a|`oauthclient`
+
+|PolicyBinding
+a|`pb`
```
rhuss: + "policybinding"

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|PersistentVolume
+a|`pv`
```
rhuss: + "persistenvolume"

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|PersistentVolumeClaim
+a|`pvc`
```
rhuss: + "persistenvolumeclaim"

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|Project
+a|`project`
```
rhuss: same comment as for "namespace": Should this be really part of an application descriptor? These descriptors are stored along with other in the jar file, but the "project" or "namespace" is clearly depending on the environment where the application is installed.

core/src/main/resources/META-INF/fabric8/kind-fileindicator-mapping-default.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+|ProjectRequest
+a|`pr`
```
rhuss: + "projectrequest". I stop here, I think you get the idea for the other rows, too ;-)

rhuss: Again, not sure what a projectrequest is at all.

rhuss:
I squashed it to a single commit so that its easier to cherry pick to the 3.5.x branch (just in case)

PR #1384, opened 2018-09-14
Fix #1383  check if casting to OpenshiftClient is possible.

### LINKED ISSUE #1383: Unsafe casts to OpenshiftClient
There are a couple of places where we straight cast to OpenshiftClient without checking if this is possible.

Its possible these casts to produce ClassCastException if for any reason communication with the api server fails.

We neeed to check before we cast and properlly handle the error if cast is not possible.

iocanel:


rohanKanojia:
I've cherry-picked on 3.5.x

PR #1385, opened 2018-09-16
chore(framents): Some doc and tests update

rhuss:


rhuss:
I did some renaming and rearranging. @lordofthejars would be awesome if you could give it a quick look.

Issue #1382, opened 2018-09-14
Allow to provide additional fragment filename mappings


rhuss:
### Description
Currently, the mapping from filename suffixes to resource kinds is predefined within the plugin (now via an internal configuration implemented for #1372). 

To be more future proof for new kinds it would be nice to allow the user to specify additional mappings.

This could be done in several ways:

* Providing a single property pointing to a properties file holding the mapping (e.g. with `suffix=kind` properties. Maybe with a sensible default for the filename which, when it exists, is automatically picked up (like the fragments)
* Allow a plugin configuration to specify the mapping in XML style.

lordofthejars:
@rhuss 
>  Maybe with a sensible default for the filename

Should be the default disk-based or classpath based? Not sure about the best usage for default.

If disk-based, which would be the best default folder location for a Maven plugin?

rhuss:
Maybe we could allow both:

* Disk-based for a quick and explicit way to specify the mapping. As a default path we unfortunately currently can't use `src/main/fabric8` as everything within this folder is considered to be reserved for profiles and fragments. But maybe we should also might reconsider this, too for 4.0. I have to revisit the current external configuration that we allow and the files involved to come to a more streamlined directory layout.
* Classpath based for e.g more persistent extensions points. That's how we allow for custom enricher and generator, too. The dependencies should be added on the plugin level. Here I would put it again below `META-INF/fabric8`

lordofthejars:
Ok then for now just classpath in `META-INF/fabric8`. 

Now checking how to make to read custom ones from Maven `configuration` tag

PR #1387, opened 2018-09-18
feat(#1382): Allow to provide additional fragment filename mappings

lordofthejars:


plugin/src/test/java/io/fabric8/maven/plugin/mojo/build/MappingConfigTest.java
```diff
@@ -0,0 +1,36 @@
+package io.fabric8.maven.plugin.mojo.build;
+
+import java.io.File;
+import org.apache.maven.plugin.testing.MojoRule;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+
+@Ignore("I have not been able to make it run due classpath problems. "
```
lordofthejars: If you agree, I will open a new issue so we can investigate later when we do the deep refactor to see if we can clarify all dependencies and make test run.

core/src/main/java/io/fabric8/maven/core/config/MappingConfig.java
```diff
@@ -0,0 +1,24 @@
+package io.fabric8.maven.core.config;
+
+import org.apache.maven.plugins.annotations.Parameter;
+
+public class MappingConfig {
+
+    @Parameter(required = true)
+    private String kind;
```
rhuss: Just as a side note: For deeper objects in a plugin configuration, `@Parameter` is not evaluated but just reflection is used. `@Parameter` works only on top-level properties in a Mojo, so they are not evaluated here and should probably be removed.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KindFilenameMapperUtil.java
```diff
@@ -27,10 +32,20 @@
     public static Map<String, List<String>> loadMappings() {
 
         final String location = "/META-INF/fabric8/kind-filename-type-mapping-default.adoc";
+        final String locationMappingProperties =
+            EnvUtil.getEnvVarOrSystemProperty("fabric8.mapping", "/META-INF/fabric8/kind-filename-type-mapping-default.properties");
```
rhuss: 👍

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KindFilenameMapperUtil.java
```diff
@@ -27,10 +32,20 @@
...
 
-        try (final InputStream mappingFile = loadContent(location)) {
+        try (final InputStream mappingFile = loadContent(location); final InputStream mappingPropertiesFile = loadContent(locationMappingProperties)) {
```
rhuss: just a question: do you know how try-catch for resources deal with null elements. I mean, is this then ignore or does it result in an NPE when the jvm tries to close() it ?

plugin/src/test/java/io/fabric8/maven/plugin/mojo/build/MappingConfigTest.java
rhuss: I think you could just add it as an it test into below `it/` where you provide a fragment with an new type, add this type to the pom.xml and then assert on the generated resources whether your given fragment has been added. 'guess this is simpler.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
```diff
@@ -191,6 +191,15 @@ public boolean accept(File dir, String name) {
 
     protected final static void initializeKindFilenameMapper() {
         final Map<String, List<String>> mappings = KindFilenameMapperUtil.loadMappings();
+        updateKindFilenameMapper(mappings);
+    }
+
+    protected final static void remove(String kind, String filename) {
```
rhuss: not sure where this is use, but could miss it. If not necessary I would avoid it, so making the mappings read-only.

core/src/main/java/io/fabric8/maven/core/config/MappingConfig.java
lordofthejars: But I have seen that in all other examples they are doing in this way, I am not saying that you are not right, I'll check, and keep you updated.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KindFilenameMapperUtil.java
lordofthejars: Ignores.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
lordofthejars: It is protected because it is needed in a test, since it is static class we need to have a way to clean it after run the test.

core/src/main/java/io/fabric8/maven/core/config/MappingConfig.java
lordofthejars: It seems you are right (https://maven.apache.org/plugin-tools/maven-plugin-annotations/apidocs/org/apache/maven/plugins/annotations/Parameter.html) then let me write an it test and check if it works or not.

rhuss: Yeah, I used that `@Parameter` also blindly before I found out that it doesn't change anything. And then forgot to clean up (as it doesn't hurt either)

rhuss:
@lordofthejars please let me know whether you can add a Changelog and documentation. would like to merge this PR early this week.

lordofthejars:
HI @rhuss I have done these two things but now I left writing an IT test to validate that it is working. Sadly this week I am out recording some video sessions and in the apartment there is no wifi so I cannot work on it, I'll try to have this IT test for Monday so we can merge then.

rhuss:
@lordofthejars np, next week is perfectly fine!

it/src/it/custom-raw-resources/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+
+      <plugin>
+        <groupId>io.fabric8</groupId>
+        <artifactId>fabric8-maven-plugin</artifactId>
+        <version>@fmp.version@</version>
+        <configuration>
+          <mappings>
+            <mapping>
+              <kind>Service</kind>
+              <filenames>foo</filenames>
```
rhuss: what's about calling it 'filenameTypes' to be consistent with e.g. https://maven.fabric8.io/#kind-filename-type-mapping ?

core/src/main/java/io/fabric8/maven/core/config/MappingConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Parameter(required = true)
+    private String filenames;
```
rhuss: Maybe `filenameKinds` (see https://maven.fabric8.io/#kind-filename-type-mapping) ? And just remove the @Parameter annotation.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -420,6 +427,16 @@ public void executeInternal() throws MojoExecutionException, MojoFailureExceptio
         }
     }
 
+    private void updateKindFilenameMappings() {
+        if (mappings != null) {
+            final Map<String, List<String>> mappingKindFilename = new HashMap<>();
+            for(MappingConfig mappingConfig : this.mappings) {
+                mappingKindFilename.put(mappingConfig.getKind(), Arrays.asList(mappingConfig.getFilenamesAsArray()));
```
rhuss: You should do a null check here and throw an appropriate error. Remember @Parameter(requied=true) is not evaluated, and hence the kind can be null.

core/src/main/java/io/fabric8/maven/core/config/MappingConfig.java
```diff
@@ -0,0 +1,24 @@
...
+    public String getKind() {
+        return kind;
+    }
+
+    public String getFilenames() {
+        return filenames;
+    }
+
+    public String[] getFilenamesAsArray() {
+        return filenames.split(",\\s*");
```
rhuss: filenames can be null, so should either return an empty string array or null. The caller should throw an exception if this return value then is null. That's the only way to ensure that a mandatory value is provided.

plugin/src/test/java/io/fabric8/maven/plugin/mojo/build/MappingConfigTest.java
rhuss: Maybe we should remove the test then for now as we have an IT now ? Don't think we need to keep it for fixing it later.

core/src/main/java/io/fabric8/maven/core/config/MappingConfig.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String[] getFilenamesAsArray() {
+        if (this.filenameTypes == null) {
+            return new String[0];
+        }
+        return filenameTypes.split(",\\s*");
+    }
+
+    public boolean isValid() {
+        return kind != null &&  filenameTypes.length() > 0;
```
rhuss: Final remark, promised ;):

I think this can still result in a NPE. Better getFilenamesAsArray() > 0.

BTW, I think it would be better to throw an error if an invalid mapping is given instead of just ignoring it. So maybe rename this to `void validate()` and thrown an `IllegalArgumentException` if invalid ?

lordofthejars: Oh sorry true.

Issue #1390, opened 2018-09-18
java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to java.lang.String


KurtStam:
When the application.yml can't be parsed it throws a class-cast exception:

java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to java.lang.String

[application.yml.txt](https://github.com/fabric8io/fabric8-maven-plugin/files/2394631/application.yml.txt)

Full log:

https://gist.github.com/KurtStam/d9b5d0eb456b0145b60239a64765d87f

rhuss:
C&P for better accessibility and durability:

* `application.yml`

```yaml
camel:
  springboot:
    # The Camel context name
    name: {{title}}
  component:
    servlet:
      mapping:
        contextPath: /*

# Binding health checks to a different port
management:
  port: 8081

# disable all management enpoints except health
endpoints:
  enabled: false
  health:
    enabled: true
```
* Log

```
[INFO] F8: Running in OpenShift mode
[INFO] F8: Using docker image name of namespace: myproject
[INFO] F8: Generators:
[INFO] F8:  - spring-boot
[INFO] F8: Running generator spring-boot
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 19.594 s
[INFO] Finished at: 2018-09-18T14:02:56-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.fabric8:fabric8-maven-plugin:3.5.41:resource (default) on project rest-dsl: Execution default of goal io.fabric8:fabric8-maven-plugin:3.5.41:resource failed: Cannot extract generator: java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to java.lang.String -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginExecutionException
```

PR #1391, opened 2018-09-19
fix(#1390): Provides better exception message in parsing error

lordofthejars:
#1390 Provides better exception message in case of invalid  in Spring Boot

### DIFF
--- a/core/src/test/resources/util/invalid-application.yml
+++ b/core/src/test/resources/util/invalid-application.yml
@@ -1,0 +1,18 @@
+camel:
+  springboot:
+    # The Camel context name
+    name: {{title}}
+  component:
+    servlet:
+      mapping:
+        contextPath: /*
+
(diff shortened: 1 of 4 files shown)


core/src/test/resources/util/invalid-application.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: {{title}}
+  component:
+    servlet:
+      mapping:
+        contextPath: /*
+
+# Binding health checks to a different port
+management:
+  port: 8081
+
```
rohanKanojia: typo ;-)

lordofthejars: where? I cannot see hahahaha

rohanKanojia: s/enpoints/endpoints

PR #1394, opened 2018-09-24
feat: Adding Service Annotations for Service Discovery

KurtStam:
Initially just for Camel Rest DSL Projects. Support for other types of projects will follow later.

See also: https://issues.jboss.org/browse/ENTESB-9252

### DIFF
--- a/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
+++ b/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package io.fabric8.maven.enricher.fabric8;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathFactory;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+import io.fabric8.maven.core.util.Configs;
+import io.fabric8.maven.core.util.MapUtil;
+import io.fabric8.maven.enricher.api.BaseEnricher;
+import io.fabric8.maven.enricher.api.EnricherContext;
+import io.fabric8.maven.enricher.api.Kind;
+
+public class ServiceDiscoveryEnricher extends BaseEnricher {
+    static final String ENRICHER_NAME        = "f8-service-discovery";
+
+    //Default Prefix
+    static final String PREFIX    = "discovery.3scale.net/";
+    //Service Annotations
+    static final String DISCOVERY_VERSION = "discovery-version";
+    static final String SCHEME            = "scheme";
+    static final String PATH              = "path";
+    static final String DESCRIPTION_PATH  = "description-path";
+    //Service Labels
+    static final String DISCOVERABLE      = "discoverable";
+
+    private File springConfigDir;
+    private String path             = null;
+    private String scheme           = "http";
+    private String descriptionPath  = null;
+    private String discoverable     = null;
+    private String discoveryVersion = "v1";
+
+    private enum Config implements Configs.Key {
+        descriptionPath,
+        discoverable,
+        discoveryVersion,
+        path,
+        scheme,
+        springDir;
+
+        public String def() { return d; } protected String d;
+    }
+    
+    public ServiceDiscoveryEnricher(EnricherContext buildContext) {
+        super(buildContext, ENRICHER_NAME);
+
+        String baseDir  = getProject().getBasedir().getAbsolutePath();
+        springConfigDir = new File(getConfig(Config.springDir, baseDir + "/src/main/resources/spring"));
+        discoverable    = getConfig(Config.discoverable, null);
+    }
+
+    @Override
+    public Map<String, String> getAnnotations(Kind kind) {
+
+        if (kind == Kind.SERVICE) {
+
+            tryCamelDSLProject();
+
+            if (discoverable != null) {
+                Map<String, String> annotations = new HashMap<>();
+                annotations.put(PREFIX + DISCOVERY_VERSION, getConfig(Config.discoveryVersion, discoveryVersion));
+                annotations.put(PREFIX + SCHEME           , getConfig(Config.scheme, scheme));
+                if (getConfig(Config.path, path) != null) {
+                    annotations.put(PREFIX + PATH             , getConfig(Config.path, path));
+                }
+                if (getConfig(Config.descriptionPath, descriptionPath) != null) {
+                    annotations.put(PREFIX + DESCRIPTION_PATH , getConfig(Config.descriptionPath, descriptionPath));
+                }
+                if (log.isVerboseEnabled()) {
+                    for (String annotationName : annotations.keySet()) {
+                        log.verbose("Add %s annotation: %s=%s", PREFIX, 
+                                annotationName, annotations.get(annotationName));
+                    }
+                }
+                return annotations;
+            }
+        }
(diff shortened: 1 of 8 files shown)


enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+import io.fabric8.maven.core.util.Configs;
+import io.fabric8.maven.core.util.MapUtil;
+import io.fabric8.maven.enricher.api.BaseEnricher;
+import io.fabric8.maven.enricher.api.EnricherContext;
+import io.fabric8.maven.enricher.api.Kind;
+
+public class ServiceDiscoveryEnricher extends BaseEnricher {
```
rhuss: Is it about a general service discovery enricher, or only for camel-rest-dsl applications ? AFAIS you can use it only if a bean context file with /beans/camelContext/restConfiguration is found (only the 'discoverable' = true is set).

Wouldn't it make also sense to allow discoverable to be specified ? Or, when description-path and path is explicitely configured don't even check for camel rest dsl but just take this over.

But even then, what about calling it '3ScaleServiceDiscoveryEnricher' ?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+    //Service Annotations
+    static final String DISCOVERY_VERSION = "discovery-version";
+    static final String SCHEME            = "scheme";
+    static final String PATH              = "path";
+    static final String DESCRIPTION_PATH  = "description-path";
+    //Service Labels
+    static final String DISCOVERABLE      = "discoverable";
+
+    private File springConfigDir;
+    private String prefix;
```
rhuss: shouldn't be configurable as it's about this very specific specification we agreed on. It doesn't make sense to make parts of the key configurable (as it open also the danger that someone uses it, but the consumer still expect 3scale). The less moving parts here, the better.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+    @Override
+    public Map<String, String> getAnnotations(Kind kind) {
+
+        if (kind == Kind.SERVICE) {
+
+            tryCamelDSLProject();
+
+            if (discoverable != null) {
+                Map<String, String> annotations = new HashMap<>();
+                MapUtil.putIfAbsent(annotations, prefix + DISCOVERY_VERSION, getConfig(Config.discoveryVersion, discoveryVersion));
```
rhuss: why putIfAbsent ? How could a key already be included if you create the map just one line before ?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+                MapUtil.putIfAbsent(annotations, prefix + SCHEME           , getConfig(Config.scheme, scheme));
+                if (getConfig(Config.path, path) != null) {
+                    MapUtil.putIfAbsent(annotations, prefix + PATH             , getConfig(Config.path, path));
+                }
+                if (getConfig(Config.descriptionPath, descriptionPath) != null) {
```
rhuss: Does it make sense to add the annotations at all if the descriptionPath is empty ? I thought this would be a mandatory parameter. if so, i would throw an exception if null, not moving on.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+    @Override
+    public Map<String, String> getLabels(Kind kind) {
+
+        if (kind == Kind.SERVICE) {
+
+            tryCamelDSLProject();
+
+            if (discoverable != null) {
+                Map<String, String> labels = new HashMap<>();
+                MapUtil.putIfAbsent(labels, prefix + DISCOVERABLE, getConfig(Config.discoverable, discoverable));
```
rhuss: Just `labels.put()` ?

KurtStam:
@rhuss not sure why travis has issues. Have you seen this failure before, or does it simply need a rerun?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+                if (log.isVerboseEnabled()) {
+                    for (String labelName : labels.keySet()) {
```
rhuss: `labels` has exactly one element. No need to iterate over it.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+                        descriptionPath = nl.getAttributes().getNamedItem("apiContextPath").getNodeValue();
+                    }
+                }
+            } catch (Exception e) {
+                log.warn("Failed to load camel context file: %s", e);
+            }
+        }
+    }
+
+    public String getPrefix() {
```
rhuss: where is this method used ?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -0,0 +1,161 @@
...
+        springConfigDir = new File(getConfig(Config.springDir, baseDir + "/src/main/resources/spring"));
+        prefix          = getConfig(Config.prefix, DEFAULT_PREFIX) + "/";
+    }
+
+    @Override
+    public Map<String, String> getAnnotations(Kind kind) {
+
+        if (kind == Kind.SERVICE) {
+
+            tryCamelDSLProject();
```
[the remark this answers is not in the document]
KurtStam: We will probably add support for other types of projects (cxf?) in the api-provider, and then we will add support for that here too. So we'd be 'trying' to recognize based on other project features.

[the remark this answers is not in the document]
rhuss: How do you then know that a "try" was successful ? Is it about the last one wins ?

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: I think it makes sense to view it as a general service discovery enricher, it for now it only supports Camel DSL.

Yeah I think one should be able to supply the 'discoverable', will add that.

Classes can't start with a '3', so I went for the more generic.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: Debatable, but I don't feel very strongly about it, will remove it if that's a deal breaker for you.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: the spec says it's optional: Use "/" if it's absent.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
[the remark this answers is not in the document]
KurtStam: last one wins.

heiko-braun:
@rhuss Who does the backport the Fuse prod branches for FMP?

rhuss:
@heiko-braun not sure which branch, nor how it was handled in the past. I was a bit disconnected of the latest FIS releases wrt fmp.

heiko-braun:
Can we merge this or is anything holding it back?

rhuss:
@KurtStam Some documentation would be nice as an extra adoc in doc/src/main/asciidoc/inc/enricher + an entry in the changelog.

Otherwise, I'm fine with this PR.

KurtStam:
@rhuss Added documentation, and squashed.

KurtStam:
oops forgot the changelog entry hold on.

PR #1395, opened 2018-09-24
Fix #1365: Generated image reference in *-deployment.yml contains duplicate registry name

### LINKED ISSUE #1365: Generated image reference in *-deployment.yml contains duplicate registry name
### Description
When using the following configuration:
```
<images>
  <image>
    <name>${docker.registry.host}/my-project-name/%a:%l</name>
   ...
  </image>
</images>
```
and the property ${docker.registry.host} contains a dot (.) in it's value, the image name appearing in the generated *-deployment.yml for kubernetes contains a duplicate registry name:

```
image: foo.bar/foo.bar/my-project-name/some-service:latest
```

When not having a dot in this name (which is impossible since it refers to a docker registry DNS name... the result is as follows:

```
image: foo/my-project-name/some-service:latest
```
### Info
Going back to f-m-p version 3.5.40 fixes this

* f-m-p version : 3.5.41
* Maven version (`mvn -v`) :
```

```
* Kubernetes / OpenShift setup and version :

rohanKanojia:
#1365

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
+++ b/core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
@@ -97,14 +97,15 @@ class ContainerHandler {
         if (StringUtils.isBlank(imageConfiguration.getName())) {
             return null;
         }
+        ImageName imageName = new ImageName(imageConfiguration.getName());
         String configuredRegistry = EnvUtil.fistRegistryOf(
-            StringUtils.isBlank(imageConfiguration.getName()) ? null : new ImageName(imageConfiguration.getName()).getRegistry(),
+            StringUtils.isBlank(imageConfiguration.getName()) ? null : imageName.getRegistry(),
             imageConfiguration.getRegistry(),
             project.getProperties().getProperty("docker.pull.registry"),
             project.getProperties().getProperty("docker.registry"));
 
         String prefix = "";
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/fabric8/maven/core/handler/ContainerHandler.java
```diff
@@ -104,7 +104,7 @@ private String getImageName(ImageConfiguration imageConfiguration) {
             project.getProperties().getProperty("docker.registry"));
 
         String prefix = "";
-        if (StringUtils.isNotBlank(configuredRegistry)) {
+        if (StringUtils.isNotBlank(configuredRegistry) && !imageConfiguration.getName().contains(configuredRegistry)) {
```
rhuss: What is if `configuredRegistry` is not blank, but imageConfiguration's name contains also a registry *but* different to configuredRegistry ? I would the expect that the ImageConfiguration's registry would "wind".

rohanKanojia: Sorry, but what does "wind" mean in this context? You mean we should give preference to configuredRegistry?

rhuss: sorry, meant "win". Fat fingers ;-) 

Yes, the registry closes to the specified image name should take precedence. This is what we use as a rule of thumb also in d-m-p.

PR #1397, opened 2018-09-26
Fix #1327 : Upgrade d-m-p

### LINKED ISSUE #1327: Upgrade d-m-p
### Description

Upgrade to the latest version of [docker-maven-plugin](https://github.com/fabric8io/docker-maven-plugin)

rohanKanojia:
Fixes #1327

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -66,7 +66,7 @@
     <version.kubernetes-client>4.0.4</version.kubernetes-client>
     <version.openshift-client>${version.kubernetes-client}</version.openshift-client>
     <version.mockwebserver>0.0.13</version.mockwebserver>
-    <version.docker-maven-plugin>0.22.1</version.docker-maven-plugin>
+    <version.docker-maven-plugin>0.26.1</version.docker-maven-plugin>
     <version.networknt.validator>0.1.7</version.networknt.validator>
     <version.jgit>5.1.1.201809181055-r</version.jgit>
     <version.hamcrest-library>1.3</version.hamcrest-library>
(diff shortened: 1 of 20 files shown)


parent/pom.xml
```diff
@@ -351,6 +357,12 @@
         <version>2.8.7</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.codehaus.plexus</groupId>
+        <artifactId>plexus-interpolation</artifactId>
```
rhuss: Do you know why this dependency is required ?

parent/pom.xml
```diff
@@ -197,6 +197,12 @@
         <version>${project.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>com.google.code.findbugs</groupId>
+        <artifactId>jsr305</artifactId>
+        <version>3.0.2</version>
```
rhuss: where is this used ?

parent/pom.xml
rohanKanojia: It's being used here: https://github.com/fabric8io/fabric8-maven-plugin/blob/7d7b13e3f675543d09e56eccc71f20abd42be0ae/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java#L426

parent/pom.xml
rohanKanojia: It is used for Nullable annotation here : https://github.com/fabric8io/fabric8-maven-plugin/blob/7d7b13e3f675543d09e56eccc71f20abd42be0ae/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/VertxHealthCheckEnricher.java#L66

PR #1398, opened 2018-09-30
Fix #1365: Remove duplicate image reference in *-deployment.yml.

### LINKED ISSUE #1365: Generated image reference in *-deployment.yml contains duplicate registry name

rhuss:
Fix #1365: Generated image reference in *-deployment.yml contains duplicate registry name (#1395)

* fix: Generated image reference in *-deployment.yml contains duplicate registry name. Fixes #1365.

This is a fix for #1365 for the 3.5. line. It differs from the master fix,
as that it *does* not add always a tag (`:latest` when not given),
but leaves it untouched for best backwards compatibility,

### DIFF
--- a/samples/xml-config/pom.xml
+++ b/samples/xml-config/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>fabric8-maven-parent</artifactId>
-    <version>3.2-SNAPSHOT</version>
+    <version>4.0-SNAPSHOT</version>
     <relativePath>../../parent/pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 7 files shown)


samples/xml-config/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>fabric8-maven-parent</artifactId>
-    <version>3.2-SNAPSHOT</version>
+    <version>4.0-SNAPSHOT</version>
```
rohanKanojia: Should this be 4.0-SNAPSHOT?

rhuss: good point. creeped in with the cherry pick. I'm going to fix this right on the branch.

PR #1399, opened 2018-10-01
Update dmp to 0.27.2

rohanKanojia:


it/src/it/env-metadata/expected/kubernetes.yml
```diff
@@ -87,12 +87,12 @@ items:
           ports:
           - containerPort: 8080
             name: http
-            protocol: TCP
+            protocol: tcp
```
rhuss: Not sure where this change comes from, but the protocol has to be UPPERCASE according to the k8s spec at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#containerport-v1-core

So instead of changing the test we should find out why it is lower case now.

core/pom.xml
```diff
@@ -136,6 +136,11 @@
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-interpolation</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
```
rhuss: why is this needed ? Actually we are trying to move away from org.json (we did in dmp already) as there are licensing issues with this library. We should at least avoid to add it, instead add the gson library as it is currently only included via a transitive dependency on d-m-p. This dependency should be made explicit.

rhuss:
Thanks, @rohanKanojia. However two other regressions have been found lately so I will do a 0.27.2 this evening. Although it only affects running containers (which is not used by fmp) we should just wait a second and update this PR to point to 0.27.2.

rohanKanojia:
@rhuss : Very cool, Thanks :-) . We do need integration tests for dmp

it/src/it/env-metadata/expected/kubernetes.yml
rohanKanojia: Sure, will check by the end of this day ;-)

rohanKanojia: @rhuss : Extremely sorry. I missed a toUpperCase() call while refactoring ;-)

core/src/main/java/io/fabric8/maven/core/util/validator/ResourceValidator.java
```diff
@@ -31,16 +31,18 @@
 import javax.validation.Path;
 import javax.validation.metadata.ConstraintDescriptor;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
```
pepperbot: MINOR Remove this unused import 'com.fasterxml.jackson.core.JsonProcessingException'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java
```diff
@@ -341,7 +342,7 @@ private Boolean checkOrCreatePullSecret(BuildServiceConfig config, OpenShiftClie
             fromImage = extractBaseFromConfiguration(buildConfig);
         }
 
-        String pullRegistry = EnvUtil.fistRegistryOf(new ImageName(fromImage).getRegistry(), dockerBuildContext.getRegistryConfig().getRegistry(), dockerBuildContext.getRegistryConfig().getRegistry());;
+        String pullRegistry = EnvUtil.firstRegistryOf(new ImageName(fromImage).getRegistry(), dockerBuildContext.getRegistryConfig().getRegistry(), dockerBuildContext.getRegistryConfig().getRegistry());;
```
pepperbot: MINOR Remove this empty statement. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)

rhuss:
@rohanKanojia are we good to merge now ?

rohanKanojia:
@rhuss : I'm concerned about this: https://github.com/fabric8io/fabric8-maven-plugin/pull/1399/files#diff-d50a80a0e33f15e85e1c51cf1956ce49L27 . Is this okay?

rhuss:
@rohanKanojia as @lordofthejars fixed his integration test, he can give probably a qualified answer here ;-)

Issue #1402, opened 2018-10-01
Apply updates to  the new discovery.3scale.net labels/annotations.


chirino:
### Description

FMP is now adding discovery.3scale.net annotations to the Services.  I see it generating services like: 

```
apiVersion: v1
kind: List
items:
- apiVersion: v1
  kind: Service
  metadata:
    annotations:
      fabric8.io/git-commit: 6f270919931c01502be53e4036d7bfa940bf357b
      discovery.3scale.net/discovery-version: v1
      prometheus.io/port: "9779"
      fabric8.io/iconUrl: img/icons/camel.svg
      fabric8.io/git-branch: master
      prometheus.io/scrape: "true"
      discovery.3scale.net/scheme: http
    labels:
      expose: "true"
      discovery.3scale.net/discoverable: "true"
      app: swagger-petstore
      provider: fabric8
      version: 1.0-SNAPSHOT
      group: io.example
    name: swagger-petstore
  spec:
    ports:
    - name: http
      port: 8080
      protocol: TCP
      targetPort: 8080
    selector:
      app: swagger-petstore
      provider: fabric8
      group: io.example
```

But it's not setting `discovery.3scale.net/port` which is now required per an update to 
https://docs.google.com/document/d/1QsunYWA9jy1VRjkIeWFUwEScSXsmaAYmuwEF1Zs5SoU/edit#
Also the label should be `discovery.3scale.net: "true"` instead of `discovery.3scale.net/discoverable: "true"`. And the `discovery.3scale.net/description-path` was also not auto set for my project.

KurtStam:
Leaving open until PR is merged.

PR #1405, opened 2018-10-03
3.5.x - [fix] Apply updates to the new discovery.3scale.net labels/annotations

KurtStam:
#1402

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ velocity.log
 /.metadata/
 /.recommenders/
 .factorypath
+bin/
--- a/doc/src/main/asciidoc/inc/_enricher.adoc
+++ b/doc/src/main/asciidoc/inc/_enricher.adoc
@@ -938,9 +938,9 @@ label and annotations:
 
 * The following annotations:
   - `discovery.3scale.net/discovery-version`: the version of the 3scale discovery process. By default its "v1"
-  - `discovery.3scale.net/scheme`: this can be "http" or "https". "http" by default.
-  - `discovery.3scale.net/path`: the contextPath of the service if it's not at the root.
-  - `discovery.3scale.net/port`: the contextPath of the service if it's not at the root.
+  - `discovery.3scale.net/scheme`: the scheme of the service, this can be "http" or "https". "http" by default.
+  - `discovery.3scale.net/path`: the contextPath of the service, defaults to "/".
+  - `discovery.3scale.net/port`: the port of the service. The enricher obtains the port from the first service definition it encounters. Please override in the plugin configuration if a different port is desired.
   - `discovery.3scale.net/description-path`: the path to the service description document (OpenAPI/Swagger). The path can either be relative or an external full URL. By default the description is picked up from the root context (/)
 
 The following configuration parameters can be used to override the behavior of this enricher:
@@ -961,9 +961,9 @@ The following configuration parameters can be used to override the behavior of t
 
 | *path*
 | The `path` part of the URL where the service is hosted.
-| `/`
+| 
 
-| *path*
+| *port*
 | The `port` part of the URL where the service is hosted.
 | `80`
 
@@ -1006,7 +1006,7 @@ kind: Service
 name:
 metadata:
   labels:
-    discovery.3scale.net/discoverable : "true"
+    discovery.3scale.net : "true"
   annotations:
     discovery.3scale.net/discovery-version : "v1"
     discovery.3scale.net/scheme : "https"
--- a/doc/src/main/asciidoc/inc/_enricher.adoc
+++ b/doc/src/main/asciidoc/inc/_enricher.adoc
@@ -58,6 +58,9 @@ fabric8-maven-plugin comes with a set of enrichers which are enabled by default.
 | <<f8-prometheus>>
 | Add Prometheus annotations.
 
+| <<f8-service-discovery>>
+| Add 3scale service discovery annotations.
+
 | <<f8-maven-scm-enricher>>
 | Add Maven SCM information as annotations to the kubernetes/openshift resources
 
@@ -931,13 +934,14 @@ The enricher extracts the needed information from the camel-context.xml, and the
 The enricher looks for the `scheme`, `contextPath` and `apiContextPath` attributes, and it will add the following
 label and annotations:
 
-* The label `discovery.3scale.net/discoverable`. The value of the label can be set to "true" or "false", and was added to take part in the selector definition executed by 3scale to find all services that need discovery. Also it can act as a switch as well, to temporary turn off 3scale discovery integration by setting it to "false".
+* The label `discovery.3scale.net`. The value of the label can be set to "true" or "false", and was added to take part in the selector definition executed by 3scale to find all services that need discovery. Also it can act as a switch as well, to temporary turn off 3scale discovery integration by setting it to "false".
 
 * The following annotations:
-  - `discovery.3scale.net/discovery-version`: (optional) the version of the 3scale discovery process. By default its "v1"
-  - `discovery.3scale.net/scheme`: (optional) this can be "http" or "https". "http" by default.
-  - `discovery.3scale.net/path`: (optional) the contextPath of the service if it's not at the root.
-  - `discovery.3scale.net/description-path`: (optional) the path to the service description document (OpenAPI/Swagger). The path can either be relative or an external full URL. By default the description is picked up from the root context (/)
+  - `discovery.3scale.net/discovery-version`: the version of the 3scale discovery process. By default its "v1"
+  - `discovery.3scale.net/scheme`: this can be "http" or "https". "http" by default.
+  - `discovery.3scale.net/path`: the contextPath of the service if it's not at the root.
+  - `discovery.3scale.net/port`: the contextPath of the service if it's not at the root.
+  - `discovery.3scale.net/description-path`: the path to the service description document (OpenAPI/Swagger). The path can either be relative or an external full URL. By default the description is picked up from the root context (/)
 
 The following configuration parameters can be used to override the behavior of this enricher:
 
@@ -959,6 +963,10 @@ The following configuration parameters can be used to override the behavior of t
 | The `path` part of the URL where the service is hosted.
 | `/`
 
+| *path*
+| The `port` part of the URL where the service is hosted.
+| `80`
+
 | *descriptionPath*
 | The path to a location where the service description document is hosted. This path can be either a relative path if the document is self-hosted, or a full URL if the document is hosted externally.
 |
@@ -967,8 +975,8 @@ The following configuration parameters can be used to override the behavior of t
 | The version of the 3scale discovery implementation.
 | `v1`
 
-| *discoverable*
-| Sets the discoverable label to either true or false. If it's set to "false" 3scale will not try to discover this service.
+| *discovery.3scale.net*
+| Sets the discovery.3scale.net label to either "true" or "false". If set to "true" 3scale will try to discover this service.
 | `true`
 
 |===
@@ -1003,6 +1011,7 @@ metadata:
     discovery.3scale.net/discovery-version : "v1"
     discovery.3scale.net/scheme : "https"
     discovery.3scale.net/path : "/api"
+    discovery.3scale.net/port : "443"
     discovery.3scale.net/description-path : "/api/openapi.json"
 spec:
   type: LoadBalancer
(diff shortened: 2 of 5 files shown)


KurtStam:
@rhuss ugh. After fixing that plugin dependency I'm running into the next issue:

[ERROR] Failed to execute goal on project fabric8-maven-enricher-api: Could not resolve dependencies for project io.fabric8:fabric8-maven-enricher-api:jar:3.5-SNAPSHOT: Could not find artifact io.fabric8:fabric8-maven-core:jar:3.5-SNAPSHOT in maven-central-plugin-snapshots (https://oss.sonatype.org/content/repositories/snapshots) -> [Help 1]

due to running `maven-dependency-plugin:2.8:go-offline`.. it can't download the things it's building.

.gitignore
```diff
@@ -12,3 +12,4 @@ velocity.log
 /.metadata/
 /.recommenders/
 .factorypath
+bin/
```
rhuss: why do you want us to ignore bin/ ? Its not so uncommon to add some tooling into a bin directory.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -107,22 +118,42 @@ public ServiceDiscoveryEnricher(EnricherContext buildContext) {
...
+            if (!tried) {
+                tryCamelDSLProject();
+                //try more project types in future, last match wins
+                tried = true;
+            }
 
             if (discoverable != null) {
                 Map<String, String> labels = new HashMap<>();
-                String labelName = PREFIX + DISCOVERABLE;
+                String labelName = PREFIX;
```
rhuss: is it really a PREFIX still ?

.gitignore
KurtStam: eclipse uses it for some state

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: We kept the annotation prefix the same as the label name on purpose so it's obvious they go together.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -107,22 +118,42 @@ public ServiceDiscoveryEnricher(EnricherContext buildContext) {
...
+    public void addMissingResources(final KubernetesListBuilder listBuilder) {
+        listBuilder.accept(new TypedVisitor<ServiceBuilder>() {
+
+            @Override
+            public void visit(ServiceBuilder serviceBuilder) {
+                
+                List<ServicePort> ports = serviceBuilder.buildSpec().getPorts();
+                if (! ports.isEmpty()) {
+                    ServicePort firstServicePort = ports.iterator().next();
+                    port = firstServicePort.getPort().toString();
```
rhuss: Shouldn't a user configured port always win ? That way a user can override the auto detection which couldn't be done otherwise. I would set port here only when port is not set. So, I also wouldn't allow a default value of '80' for the port but leave it empty. Only if no port is detected you might consider to default it to 80. But tbh, I would throw an exception then as its very likely the wrong port.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -27,27 +28,33 @@
...
     static final String PATH              = "path";
+    static final String PORT              = "port";
     static final String DESCRIPTION_PATH  = "description-path";
     //Service Labels
     static final String DISCOVERABLE      = "discoverable";
 
     private File springConfigDir;
-    private String path             = null;
+    private String path             = "/";
+    private String port             = "80";
```
rhuss: I would leave this null, also to decide whether auto-detection is required or not.

doc/src/main/asciidoc/inc/_enricher.adoc
```diff
@@ -931,13 +934,14 @@ The enricher extracts the needed information from the camel-context.xml, and the
...
 
 * The following annotations:
-  - `discovery.3scale.net/discovery-version`: (optional) the version of the 3scale discovery process. By default its "v1"
-  - `discovery.3scale.net/scheme`: (optional) this can be "http" or "https". "http" by default.
-  - `discovery.3scale.net/path`: (optional) the contextPath of the service if it's not at the root.
-  - `discovery.3scale.net/description-path`: (optional) the path to the service description document (OpenAPI/Swagger). The path can either be relative or an external full URL. By default the description is picked up from the root context (/)
+  - `discovery.3scale.net/discovery-version`: the version of the 3scale discovery process. By default its "v1"
+  - `discovery.3scale.net/scheme`: this can be "http" or "https". "http" by default.
+  - `discovery.3scale.net/path`: the contextPath of the service if it's not at the root.
+  - `discovery.3scale.net/port`: the contextPath of the service if it's not at the root.
```
rhuss: why have you remove the optional part ? I found that quite useful.

parent/pom.xml
```diff
@@ -545,7 +545,7 @@
         <plugin>
           <groupId>org.codehaus.plexus</groupId>
           <artifactId>plexus-component-metadata</artifactId>
-          <version>1.6</version>
+          <version>1.7</version>
```
rhuss: I'm really not sure why this is required as we haven't any change anything wrt to the build setup. But its fine for me.

doc/src/main/asciidoc/inc/_enricher.adoc
rhuss: C&P Error: `/port` is *not* the "contextPath".

doc/src/main/asciidoc/inc/_enricher.adoc
```diff
@@ -959,6 +963,10 @@ The following configuration parameters can be used to override the behavior of t
 | The `path` part of the URL where the service is hosted.
 | `/`
 
+| *path*
```
rhuss: "port" ?

Also, I would explain the lookup mechanism here, e.g. that a service is inspected

doc/src/main/asciidoc/inc/_enricher.adoc
KurtStam: I think everything is optional from a config perspective at this point as the code will always create the annotations that are required by 3scale. So to me it got confusing if it was a 'config required' or a '3scale required'.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
```diff
@@ -107,22 +118,42 @@ public ServiceDiscoveryEnricher(EnricherContext buildContext) {
...
         return super.getLabels(kind);
     }
 
+    @Override
+    public void addMissingResources(final KubernetesListBuilder listBuilder) {
+        listBuilder.accept(new TypedVisitor<ServiceBuilder>() {
+
+            @Override
+            public void visit(ServiceBuilder serviceBuilder) {
+
```
rhuss: How are you sure that the service given here is the one with the annotations found ? You can configure many services, e.g. when configured manually via fragments. Not sure how this works with you stateful enricher.

You can easily put the wrong port in the annotation when e.g. an additional service with different ports comes last  in the visitor.

parent/pom.xml
KurtStam: @rhuss y no idea. Somehow it would reference the old group name of 'jdom' instead of the new 'org.jdom', which is fixed in 1.7.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: I'm taking the first port - which it logs out. TBH I was against specifying the port at all but here we are. The user can always override with the correct port, but 90% of the time this will probably work just fine. There is no way of knowing what the correct port is for the enricher.

rhuss: As already commented, it's currently the **other way round**: The autodetection always overrides a specified port.

Tbh, I would do *all* the handling (adding labels and annotations) in the visitor. That way you can avoid the intermediate state handover with enricher member variables and you can deal with multiple services.

rhuss: If you can't specify the port you can't annotate headless services, so adding a port makes sense for this case.

parent/pom.xml
rhuss: ok, thanks.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: Y I debated this, but @chirino seems to expect getting a value there regardless.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: @rhuss yes it should, so in line 101 we do: 
... getConfig(Config.port, port))
So the Config.port wins if specified.

3scale wants port to be 'required' value. So I have to specify it..

True that if no port is found an exception or warning is appropriate. I will add that.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: So you're saying that defaultValue is used over the one given by the config here?

getConfig(Configs.Key key, String defaultVal)

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
rhuss: @KurtStam so if port is a required value when the even examine the port mapping of the service as the required value will allways win ? (i would remove the 80 as default value and throw an exception if getConfig(port) is giving null.

But you are right, the config in your code override the default port, **but** are you sure that `addMissingResource()` is called *before* `getAnnotations()`. The order is not defined, so as mentioned, carrying over state via member variables is dangerous. Why not skipping `getAnnotations()` and just do everything in the visitor atomically.

Well, but as I would remove the autodecection anyway, this is probably not a big deal anymore.

KurtStam: Sure I can move everything into the addMissingResource, that makes sense. I do think we should prevent the user from forcing to specify a port as most services will only have 1 port.

KurtStam:
@rhuss I think it should be good to go now.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ServiceDiscoveryEnricher.java
KurtStam: I spoke with @chirino and changed it back

KurtStam:
@rhuss what's going on with circle-ci? I'd like to see this merged.

doc/src/main/asciidoc/inc/_enricher.adoc
```diff
@@ -967,8 +975,8 @@ The following configuration parameters can be used to override the behavior of t
 | The version of the 3scale discovery implementation.
 | `v1`
 
-| *discoverable*
-| Sets the discoverable label to either true or false. If it's set to "false" 3scale will not try to discover this service.
+| *discovery.3scale.net*
```
rhuss: sorry, again me: The configuration is called 'discoverable', isn't it ? It's only the label which is different.

KurtStam: Yes! Good catch I search/replaced too aggressively.

PR #1417, opened 2018-10-15
chore: Remove MergeEnricher as it is already broken ..

rhuss:
... and nobody recognized it which is a high indicator that nobody is using it.

We can re-introduce it later in case we would need some extra tuning
for the dependency enricher.


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/MergeEnricher.java
```diff
@@ -1,104 +0,0 @@
...
-            Map<String, HasMetadata> map = kindMaps.get(kind);
-            if (map == null) {
-                map = new HashMap<>();
-                kindMaps.put(kind, map);
-            }
-            HasMetadata old = map.get(name);
-            if (old != null) {
-                HasMetadata removeItem = mergeEntities(old, item);
-                if (removeItem != null) {
-                    removeList.add(removeItem);
```
rhuss: If two items have been merged successfully (return value != null) then it shoud be removed from the list ? Doesn't make sense at all...

PR #1423, opened 2018-10-23
Use the current 3.5.x last release version for docs purpose

oscerd:


rhuss:
@oscerd thanks ! But I'm afraid that the docs are generated from master only. Going to fix that for now, to more accurately reflect the current state.

PR #1424, opened 2018-10-24
feat(#977): Allows user to define ConfigMap from XML and annotation

### LINKED ISSUE #977: ConfigMaps content from external data (fragment/xml config)

lordofthejars:
Fixes #977

### DIFF
--- a/doc/src/main/asciidoc/inc/_enricher.adoc
+++ b/doc/src/main/asciidoc/inc/_enricher.adoc
@@ -87,6 +87,9 @@ fabric8-maven-plugin comes with a set of enrichers which are enabled by default.
 
 | <<fmp-triggers-annotation>>
 | Add ImageStreamTag change triggers on Kubernetes resources such as StatefulSets, ReplicaSets and DaemonSets using the `image.openshift.io/triggers` annotation.
+
+| <<fmp-configmap-file>>
+| Add ConfigMap elements defined as XML or as annotation.
 |===
 
 [[enrichers-standard]]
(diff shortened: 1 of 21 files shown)


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -978,6 +998,48 @@ private void addConfiguredResources(KubernetesListBuilder builder, List<ImageCon
...
+            element.withNewMetadata().withName("xmlconfig").endMetadata();
+            element.addToData(configMapFromConfiguration);
+
+            builder.addToConfigMapItems(element.build());
+        } catch (IOException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    private Map<String, String> createConfigMapFromConfiguration(io.fabric8.maven.core.config.ConfigMap configMap) throws IOException {
```
pepperbot: CRITICAL Refactor this method to reduce its Cognitive Complexity from 17 to the 15 allowed. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS3776)

doc/src/main/asciidoc/inc/_enricher.adoc
```diff
@@ -1017,6 +1020,63 @@ The trigger is added to all containers that apply, but can be restricted to a li
...
+This creates a ConfigMap data with key `A` and value `B`.
+
+You can also use `file` tag to refer to the content of a file.
+
+[source, xml]
+----
+<configuration>
+  <resources>
+    <configMap>
+      <elements>
```
rhuss: what about using `<entries>` and `<entry>` instead of `<element>` ? That would match nicely with the "map" semantics which also refers to entries.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ConfigMapEnricher.java
```diff
@@ -0,0 +1,85 @@
...
+                configMapFileLocations.put(getOutput(key), readContent(entry.getValue()));
+                it.remove();
+            }
+        }
+
+        return configMapFileLocations;
+    }
+
+    private String readContent(String location) throws IOException {
+        return new String(Files.readAllBytes(Paths.get(location)));
```
rhuss: cool, didn't know that ;-)

PR #1427, opened 2018-10-29
fix(#918): Adds Sanitize method to generate valid Kubernetes resources

### LINKED ISSUE #918: Sanitize Maven project names when artifactId starts with a number
to avoid this error that Max just saw
```
	
10:30 AM
is this expected: "Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://kubernetes.default/api/v1/namespaces/manderse-stage/services. Message: Service "2017-04-20a" is invalid: metadata.name: Invalid value: "2017-04-20a": must match the regex a-z? (e.g. 'my-name' or 'abc-123')."
```

we could just add a default "s" or something to the artifact id if it starts with a digit

lordofthejars:
Fix #918: Adds Sanitize method to generate valid Kubernetes resources names

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
+++ b/core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
@@ -49,7 +49,7 @@ public class GroupArtifactVersion {
      * ArtifactId is used for setting a resource name (service, pod,...) in Kubernetes resource.
      * The problem is that a Kubernetes resource name must start by a char.
      * This method returns a valid string to be used as Kubernetes name.
-     * @return Sanititzed Kubernetes name.
+     * @return Sanitized Kubernetes name.
      */
     public String getSanitizedArtifactId() {
         if (this.artifactId != null && !this.artifactId.isEmpty() && Character.isDigit(this.artifactId.charAt(0))) {
--- a/core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
+++ b/core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
@@ -17,6 +17,8 @@ package io.fabric8.maven.core.model;
 
 public class GroupArtifactVersion {
 
+    private static final String PREFIX = "s";
+
     private String groupId;
     private String artifactId;
     private String version;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
```diff
@@ -42,4 +44,18 @@ public String getVersion() {
...
         return getVersion() != null && getVersion().endsWith("SNAPSHOT");
     }
+
+    /**
+     * ArtifactId is used for setting a resource name (service, pod,...) in Kubernetes resource.
+     * The problem is that a Kubernetes resource name must start by a char.
+     * This method returns a valid string to be used as Kubernetes name.
+     * @return Sanititzed Kubernetes name.
+     */
+    public String getSanitizedArtifactId() {
```
rohanKanojia: I think apart from checking whether artifactId starts with a digit we should also sanitize it based on regex: `[a-z0-9]([-a-z0-9]*[a-z0-9])?` and max length(i.e 63 characters). WDYT?

lordofthejars: I thought the same but, look if image name starts with a number there is an easy fix (put a char), but in case of more than 63 chars (we cannot do anything, better to not substring a full name because probably is to avoid any duplicated). And again any other char, what would be the solution?

For this reason, I think that the best is to fix what we can and leave as an exception the other things.

But of course, it is an opinion.

rohanKanojia: ah, yeah. Makes sense... as always ;-)

core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
```diff
@@ -42,4 +44,18 @@ public String getVersion() {
     public boolean isSnapshot() {
         return getVersion() != null && getVersion().endsWith("SNAPSHOT");
     }
+
+    /**
+     * ArtifactId is used for setting a resource name (service, pod,...) in Kubernetes resource.
+     * The problem is that a Kubernetes resource name must start by a char.
+     * This method returns a valid string to be used as Kubernetes name.
+     * @return Sanititzed Kubernetes name.
```
rohanKanojia: Minor typo

core/src/main/java/io/fabric8/maven/core/model/GroupArtifactVersion.java
lordofthejars: Well most of the time I say things that has no sense at all :)

lordofthejars:
Thanks for the review.

rohanKanojia:
hehe, np. This PR was smaller than your previous ones. I'm not sure why CircleCI is failing. I think we can proceed with merge,

PR #1429, opened 2018-10-31
Fix CircleCI breaking build

rohanKanojia:
Use machine build instead of docker executor which is breaking right now


lordofthejars:
@rohanKanojia I merge it because I think is somehow urgent.

PR #1431, opened 2018-11-08
Fix #977 regression: Don't add ConfigMapEnricher to default profile

### LINKED ISSUE #977: ConfigMaps content from external data (fragment/xml config)

rohanKanojia:
Only run ConfigMapEnricher when ConfigMapEntries are present in pom


dev-gaur:
Build works fine on my local setup. LGTM.

PR #1433, opened 2018-11-13
Fix #222 : ServiceEnricher should evaluate Image labels for ports

### LINKED ISSUE #222: ServiceGenerator should evaluate Image labels for ports
The `SeviceEnricher` could check the Docker image configuration for specific labels (e.g. `fabric8.service.myservice=9090`) so that it could use this in a service's port specification. The last part of the label should be the service name.

That ways, generators like the spring-boot generator can transport such information to the enrichers in a transparent way. It would be also nice to have that sort of meta-data directly within the Docker image.

rohanKanojia:
Fixes #222 The SeviceEnricher could check the Docker image configuration for specific labels
(e.g. fabric8.service.myservice=9090) so that it could use this in a service's port
specification. The last part of the label should be the service name.


rohanKanojia:
@rhuss : ok, let me give it a try and update. I didn't think about sidecars while implementing this.

rhuss:
Let's get this merged now, we can refine in M3 then.

PR #1437, opened 2018-11-16
Java11

ffang:


ffang:
With this PR, "mvn clean install" are successful both on JDK8 and JDK11.

core/src/main/java/io/fabric8/maven/core/config/PlatformMode.java
```diff
@@ -66,6 +66,10 @@ public String getLabel() {
      * Returns true if the given maven properties indicate running in OpenShift platform mode
      */
     public static boolean isOpenShiftMode(Properties properties) {
-        return Objects.equal(openshift.toString(), properties.getProperty(FABRIC8_EFFECTIVE_PLATFORM_MODE, ""));
+        if (properties == null) {
```
lordofthejars: Maybe I will use a ternary in this case just for make code clean, but just my opinion of course.

generator/spring-boot/src/test/java/io/fabric8/maven/generator/springboot/SpringBootGeneratorTest.java
```diff
@@ -71,6 +69,9 @@ public void javaOptions() throws IOException, MojoExecutionException {
         assertEquals(1, configs.size());
         Map<String, String> env = configs.get(0).getBuildConfiguration().getEnv();
         assertNull(env.get("JAVA_OPTIONS"));
+        } catch (Exception ex) {
+            ex.printStackTrace();
```
lordofthejars: can you make the test make error by rethrowing the exception?
Also if there is an exception the test will fail automatically and print the stack trace.

ffang: Ah, that just for my debug and surely should be removed.
Another commit is coming soon...

core/src/main/java/io/fabric8/maven/core/config/PlatformMode.java
ffang: Yup, good point!

PR #1440, opened 2018-11-26
Fix #1019: Custom Liveness/Readiness probes are not being created

### LINKED ISSUE #1019: Custom liveness/readiness probes are not being created

rohanKanojia:
Fix #1019: Added CustomProbeEnricher for adding probes via XML config

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
+++ b/core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
@@ -154,6 +154,29 @@ public class ResourceConfig {
     public static class Builder {
         private ResourceConfig config = new ResourceConfig();
 
+        public Builder() { }
+
+        public Builder(ResourceConfig config) {
+            if(config != null) {
+                this.config.env = config.getEnv().orElse(null);
+                this.config.controllerName = config.getControllerName();
+                this.config.imagePullPolicy = config.getImagePullPolicy();
+                this.config.replicas = config.getReplicas();
+                this.config.liveness = config.getLiveness();
+                this.config.readiness = config.getReadiness();
+                this.config.annotations = config.getAnnotations();
+                this.config.serviceAccount = config.getServiceAccount();
+                this.config.configMap = config.getConfigMap();
+                this.config.volumes = config.getVolumes();
+                this.config.labels = config.getLabels();
+                this.config.annotations = config.getAnnotations();
+                this.config.secrets = config.getSecrets();
+                this.config.services = config.getServices();
+                this.config.metrics = config.getMetrics();
+                this.config.namespace = config.getNamespace();
+            }
+        }
+
         public Builder env(Map<String, String> env) {
             config.env = env;
             return this;
@@ -186,6 +209,16 @@ public class ResourceConfig {
 
         public Builder withConfigMap(ConfigMap configMap) {
             config.configMap = configMap;
+            return this;
+        }
+
+        public Builder withLiveness(ProbeConfig liveness) {
+            config.liveness = liveness;
+            return this;
+        }
+
+        public Builder withReadiness(ProbeConfig readiness) {
+            config.readiness = readiness;
             return this;
         }
 
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
@@ -97,11 +97,11 @@ public class DefaultControllerEnricher extends BaseEnricher {
     @Override
     public void addMissingResources(KubernetesListBuilder builder) {
         final String name = getConfig(Config.name, MavenUtil.createDefaultResourceName(getContext().getGav().getSanitizedArtifactId()));
-        final ResourceConfig config = new ResourceConfig.Builder()
-                    .controllerName(name)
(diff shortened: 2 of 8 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/CustomProbeEnricher.java
```diff
@@ -0,0 +1,66 @@
...
+                        }
+                    }
+
+                    ProbeConfig livenessProbeConfig = resourceConfig.getLiveness();
+                    if (livenessProbeConfig != null) {
+                        Probe livenessProbe = probeHandler.getProbe(livenessProbeConfig);
+                        if (livenessProbe != null) {
+                            containerBuilder.withLivenessProbe(livenessProbe);
+                        }
+                    }
```
lordofthejars: I think we could create a method which returns the `Probe`.

```java
getProbe(ProbeConfig pc) {
  if (pc != null) {
    return  probeHandler.getProbe(pc);
  }
  return null;
}
```

And then it is called twice:

```java
containerBuilder.withLivenesssProve(getProbe(resourceConfig.getLiveness())
```

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/CustomProbeEnricher.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public void addMissingResources(KubernetesListBuilder builder) {
+        builder.accept(new TypedVisitor<ContainerBuilder>() {
+            @Override
+            public void visit(ContainerBuilder containerBuilder) {
+                try {
+                    ResourceConfig resourceConfig = getConfiguration().getResource().get();
+
+                    Probe readinessProbe = getProbe(resourceConfig.getReadiness());
+                    if (readinessProbe != null) {
```
lordofthejars: Is this null check really necessary? I mean I think we can just call the `withXXXProbe` directly since by default its value is null, we are just setting again a null.

enricher/standard/src/test/java/io/fabric8/maven/enricher/standard/CustomProbeEnricherTest.java
```diff
@@ -0,0 +1,131 @@
...
+        // Enrich
+        CustomProbeEnricher enricher = new CustomProbeEnricher(context);
+        KubernetesListBuilder builder = getPodTemplateList();
+        enricher.addMissingResources(builder);
+
+        // Validate that the generated resource contains
+        KubernetesList list = builder.build();
+
+        String json = ResourceUtil.toJson(list.getItems().get(0));
+        System.out.println(json);
```
dev-gaur: forgot to omit this?

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
```diff
@@ -97,11 +97,26 @@ public DefaultControllerEnricher(MavenEnricherContext buildContext) {
...
+                    .withServiceAccount(resourceConfig.getServiceAccount())
                     .withReplicas(Configs.asInt(getConfig(Config.replicaCount)))
                     .build();
+        } else {
+            config = new ResourceConfig.Builder()
+                    .controllerName(name)
+                    .imagePullPolicy(getConfig(Config.pullPolicy))
+                    .withReplicas(Configs.asInt(getConfig(Config.replicaCount)))
+                    .build();
+        }
```
rhuss: would do sth like which makes it less redundant

```java
configBuilder = new ResourceConfig.Builder().controllerName(name)...;
if (resourceConfig != null) {
  configBuilder.volumes(...)...;

}

config = configBuilder.build();
```

rhuss: Sorry to bother you again, but what do you think about using a `ResourceConfig.Builder(ResourceConfig)` constructor which makes a deep copy of the given argument and then uses this as the internal resource object which is handed out later. That way, we can add new fields to ResourceConfig without bothering of enrichers using it (they get automatically the new fields initialized).
https://github.com/fabric8io/docker-maven-plugin/blob/38eba5ffa6920ea0bd24a1cc03c5c7224a26d4f4/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java#L323-L329 how this is done for the BuildConfiguration.


See e.g.

core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
```diff
@@ -154,6 +154,29 @@ public ConfigMap getConfigMap() {
     public static class Builder {
         private ResourceConfig config = new ResourceConfig();
 
+        public Builder() { }
+
+        public Builder(ResourceConfig config) {
+            if(config != null) {
+                this.config.env = config.getEnv().orElse(null);
+                this.config.controllerName = config.getControllerName();
```
rhuss: Thanks, I think this is good enough (although its not really a complete deep-copy, as you would have to create new collections, too)

PR #1438, opened 2018-11-16
Fixes overriding of SpecSelector MatchLabel entries by the enrichers entries.

dev-gaur:
Fixes #1326 .

PR #1442, opened 2018-11-27
Modified i-tests and docs as per PR #1438.

dev-gaur:


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ProjectEnricher.java
```diff
@@ -41,6 +41,8 @@
  *
  * The "app" label can be replaced with the (old) "project" label using the "useProjectLabel" configuraiton option.
  *
+ * The project labels which are already specified in the input fragments are not overriden by the enricher.
+ *
```
rohanKanojia: s/overriden/overridden ;-)

dev-gaur: Thanks!

Issue #712, opened 2016-12-06
Specify Kuberentes/OpenShift login data via configuration (XML/properties)


apupier:
_Please fill out the following information to help us in analyzing the issue, but feel free to skip it if you don't have the information at hand or if it does not apply. Please remove everything which does not apply to the issue. Including this info paragraph ;-)_

### Description

Maven properties cannot be used to configure the targeted OpenShift instance/application.
It currently requires to use System Environment variables or System Properties.

The Fabric8-maven-plugin relies exclusively on Fabric8 Kubernetes client which is not taking care of Maven properties.

See https://maven.apache.org/plugin-developers/common-bugs.html#Using_System_Properties for Maven plugin best practice
And the implementation currently used as it is delegated completely to another library: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java#L44

### Info

* f-m-p version : 3.2 and 3.1.80
* Maven version (`mvn -v`) :
```

```
* Kubernetes / OpenShift setup and version : CDK 2.3
* If it's a bug, how to reproduce :
* If it's a feature request, what is your use case :
* Sample project : *[GitHub Clone URL]*

jstrachan:
I don't quite follow what this issue means. What do you mean "targeted OpenShift instance/application."

In terms of connecting to a kubernetes cluster and namespace the fabric8-maven-plugin reuses the same `~/.kube/config` file - so it uses the same cluster + namespace as `kubectl` or `oc`. Or if you run maven jobs inside, say, Jenkins inside kubernetes it'll just reuse the local cluster and namespace OOTB.

What do you need?

jimmidyson:
I think @apupier is suggesting that the connection to OpenShift/Kubernetes should be configurable via Maven properties rather than relying on the kubernetes-client library.

apupier:
using ~/.kube/.config is a big issue in a multi-module, CI environment. This is fine only to do little demos.
I asked to have the possibility to use Maven Properties when using the fabric8-maven-plugin.
@nicolaferraro implemented back the way to use system variables instead of relying on .kube/config with https://github.com/fabric8io/fabric8-maven-plugin/pull/689 which is already a good step forward.
Using Maven properties as explained in https://maven.apache.org/plugin-developers/common-bugs.html#Using_System_Properties is the scalable and reliable way of implementing a maven plugin.

jstrachan:
[...]

PR #1448, opened 2018-12-10
fix(#712): Specify Kuberentes/OpenShift login data via configuration

lordofthejars:


plugin/src/main/java/io/fabric8/maven/plugin/mojo/AbstractFabric8Mojo.java
```diff
@@ -49,6 +50,97 @@
...
+    /**
+     * Master URL on which to operate
+     */
+    @Parameter(property = "fabric8.masterUrl")
+    private String masterUrl;
+
+    /**
+     * Api version on which to operate
+     */
+    @Parameter(property = "fabric8.apiVersion")
```
rhuss: Couldn't we use `ClusterConfiguration` directly as property so that we would have nested configs like

```xml
<configuration>
  <access>
    <apiVersion>...</apiVersion>
    <caCertData>....</caCertData>
    ....
  </access>
</configuration>
```

This would reduce the noise in the `AbstractDockerMojo`. The drawback would be that we would have to evaluate the properties on our own. However in this case we would be even able to use both Maven properties *and* system properties (which is not the case when using @Parameter)

core/src/main/java/io/fabric8/maven/core/access/ClusterConfiguration.java
```diff
@@ -0,0 +1,178 @@
...
+    private String clientKeyPassphrase;
+    private String trustStoreFile;
+    private String trustStorePassphrase;
+    private String keyStoreFile;
+    private String keyStorePassphrase;
+
+    public ClusterConfiguration() {
+    }
+
+    public void setMasterUrl(String masterUrl) {
```
rhuss: Actually we are using the builder pattern in our configuration classes (e.g. BuildImageConfiguration) without setters. 'would be awesome if we could stick to this pattern for consistencies sake.

core/src/main/java/io/fabric8/maven/core/access/ClusterConfiguration.java
```diff
@@ -0,0 +1,178 @@
...
+    private String clientKeyAlgo;
+    private String clientKeyPassphrase;
+    private String trustStoreFile;
+    private String trustStorePassphrase;
+    private String keyStoreFile;
+    private String keyStorePassphrase;
+
+    public ClusterConfiguration() {
+    }
+
```
rhuss: I would also add an `fromProps(Properties)` to a potential builder method to extract the values from system/maven properties. This would have the additional advantage that the property names are only used in such a method and would be duplicated over multiple mojos.

core/src/main/java/io/fabric8/maven/core/access/ClusterConfiguration.java
rhuss: what I mean:

```java
ClusterConfiguration cfg = 
     new ClusterConfiguration.Builder()
         .from(project.getProperties)
         .from(System.getProperties()
         .namespace(ns)
         .caCertFile(file)
         // ....
         .build();
```

core/src/main/java/io/fabric8/maven/core/access/ClusterAccess.java
```diff
@@ -15,50 +15,53 @@
...
     }
 
-    public ClusterAccess(String namespace, KubernetesClient client){
-        this.namespace = setNamespace(namespace);
+    public ClusterAccess(ClusterConfiguration clusterConfiguration, KubernetesClient client) {
+        this.clusterConfiguration = clusterConfiguration;
         this.client = client;
     }
 
+    /**public ClusterAccess(String namespace, KubernetesClient client) {
```
rohanKanojia: I think you forgot to remove this :-)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/AbstractFabric8Mojo.java
lordofthejars: I thought about it but notice that this will break back compatibility (`namespace` field) was already there before this change.

rhuss: We can keep the namespace field, mark it is deprecated and if set, used it to fill in the ClusterConfiguration's namespace field if this is set (but not the one in ClusterConfiguration). That we, still can support the old namespace part for some time and remove it eventually. 

Not sure so, as we are moving to 4.0 anyway, whether we just should do the switch and note this as a migration step. 

wdyt ?

lordofthejars: ok, then let's remove the field, I prefer to remove it right now instead of maintaining because if we maintain, we will never find the correct time to remove it.

PR #1450, opened 2018-12-12
Fix #1360: Namespace missing in generated DeploymentConfig

### LINKED ISSUE #1360: Namespace missing in generated DeploymentConfig
### Description
When I generate the openshift resources with `mvn clean package fabric8:build fabric8:resource` e.g. in namespace test, the generated DeploymentConfig does not contain the namespace in the image-ref.

### Info

* f-m-p version : 3.5.41
* Maven version (`mvn -v`) : 
```
Apache Maven 3.3.9
Maven home: /usr/share/maven
Java version: 1.8.0_171, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: de_DE, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-130-generic", arch: "amd64", family: "unix"
```
* Kubernetes / OpenShift setup and version : 1 master, 4 nodes
```
openshift v3.9.0+ee4d1e7-20
kubernetes v1.9.1+a0ce1bc657
```
* If it's a bug, how to reproduce :

** login to openshift project test
** run `mvn clean package farbic8:build fabric8:resource`
** openshift.yaml has a reference to the image in the DeploymentConfig def: $.spec.template.spec.containers[0].image
```
image: my-service:latest
```

IMHO this should also reference the namespace/project where the image has been pushed to otherwise when deploying the resource to another project it won't find the image in the current namespace.

```
image:test/my-service:latest
```

rohanKanojia:
Fix #1360

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/NamespaceEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/NamespaceEnricher.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package io.fabric8.maven.enricher.standard;
+
+import io.fabric8.kubernetes.api.builder.TypedVisitor;
+import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
+import io.fabric8.maven.core.util.Configs;
+import io.fabric8.maven.core.util.kubernetes.KubernetesHelper;
+import io.fabric8.maven.enricher.api.BaseEnricher;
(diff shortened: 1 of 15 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/NamespaceEnricher.java
```diff
@@ -0,0 +1,80 @@
...
+    // - plugin configuration
+    // - default name space from the kubernetes helper
+    // - "default"
+    private String getNamespace() {
+        String namespace = getConfig(Config.namespace);
+        if (StringUtils.isNotBlank(namespace)){
+            return namespace;
+        }
+
+        namespace = getContext().getConfiguration().getProperty("fabric8.namespace");
```
rhuss: With the changes from @lordofthejars the namespace also can come from the `<access>` configuration element. I'm not really sure how to deal with the namespace in generated resources. 

I mean IMO there should be no namespace included in resource descriptors which are supposed to be specific to the application and not to any target environment. Remember, the resource descriptors end up in the jar artefacts (below META-INF/) and distributed via Maven repositories. If a namespace is included here, it assumes that this namespace is available everywhere else. But this is obviously not the case.

So I really would hesitate to add the namespace directly to the resources. Instead a namespace makes sense, when _applying_ a descriptor. Or in other words, `fabric8:resource` should be namespace agnostric, and `fabric8:apply` should apply the resources in a given namespace (as it comes from the cluster access configuration). A user still can add manually a namespace to their resources if they want to (but I would highly discourage this).

We could add the enricher as optional, but would only respect the namespace given in the enricher configuration.

wdyth ?

rhuss: Ok, lets close that. I think the initial motivation coming from #1360 was that the connection to the imagestream is wrong. I thinl we could fix that differently.

rhuss:
We should think about a different solution for #1360 as adding the current namespace to all resources (which should be independent of the target environment) is not the proper way to go (see comments above).

PR #1451, opened 2018-12-12
Fix #401: Refactor DefaultControllerEnricher

### LINKED ISSUE #401: Refactor DefaultControlllerEnricher
Currently DefaultControllerEnricher has two task which should be separated into two different Enricher so that they can be switched on / off independently:
- Create a default controller (Deployment, DeploymentConfig, ReplicaSet, ReplicationController) if no such object exists for this project.
- Enhance a given controller with the configuration information presented in the plugin configuration.

rohanKanojia:
Fix #401

### DIFF
--- a/plugin/src/main/resources/META-INF/fabric8/profiles-default.yml
+++ b/plugin/src/main/resources/META-INF/fabric8/profiles-default.yml
@@ -24,6 +24,7 @@
     includes:
     - fmp-name
     - fmp-controller
+    - fmp-controller-from-configuration
     - fmp-service
     - fmp-image
     - fmp-portname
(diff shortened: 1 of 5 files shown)


plugin/src/main/resources/META-INF/fabric8/profiles-default.yml
```diff
@@ -24,6 +24,7 @@
     includes:
     - fmp-name
     - fmp-controller
+    - fmp-pluginConfiguration-controller
```
rhuss: can we avaoid to introduce camel case for enricher names ? I just would call it `fmp-controller-from-configuration` or so.

Issue #601, opened 2016-10-24
Add a HealthCheck Enricher for Webapps


jstrachan:
we should generate a readiness & liveness checks OOTB for war packaging apps.

If there is a service annotation for [api.service.kubernetes.io/path](https://github.com/fabric8-quickstarts/spring-boot-hystrix/blob/master/src/main/fabric8/service.yml#L3) we could use that if no explicit happy page URL is configured in the pom.xml; otherwise we could just default to using `/`

rhuss:
the health check enricher should work in the same spirit as the enrichers we have now for Vertx, Prometheus etc.

Alternatively, the webapp generator could also add just a Docker `HEALTHCHECK` we already can be reused by the `DockerHealthCheckEnricher`

PR #1456, opened 2018-12-17
feat(#601): Adds webapp healthckech

lordofthejars:


enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WebAppHealthCheckEnricher.java
```diff
@@ -0,0 +1,110 @@
...
+        String scheme = getScheme().toUpperCase();
+        String path = getPath();
+
+        return new ProbeBuilder().
+            withNewHttpGet().withNewPort(port).withPath(path).withScheme(scheme).endHttpGet().
+            withInitialDelaySeconds(getInitialDelay(readiness)).build();
+
+    }
+
+    private int getInitialDelay(boolean readness) {
```
dev-gaur: @lordofthejars You missed the 'i' in readiness . At line 89 and 93. It doesn't matter much.. But would be good to maintain naming consistency.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/WebAppHealthCheckEnricher.java
```diff
@@ -0,0 +1,110 @@
...
+        final int initialDelay = Configs.asInt(getConfig(Config.initialDelay));
+
+        if (initialDelay < 0) {
+            return readness ? DEFAULT_DELAY_READNESS : DEFAULT_DELAY_LIVENESS;
+        }
+
+        return  initialDelay;
```
dev-gaur: What if I wanted specify two different Initial Delay values for  Readiness and Liveness probes? Even the default values specified by you for Readiness and Liveness probes are different ( 100 and 1000). So, it should be allowed to specifiy two different Initial Delay values for the two probes.
For eg.,
```
<f8-healthcheck-webapp>
<config>
  <initialdelayReadiness>200</initialdelayReadiness>
  <initialDelayLiveness>400</initialDelayLiveness>
</config>
</f8-healthcheck-webapp>
```

dev-gaur: If we opt for this, then we can eliminate this getInitialDelay( ) method and change to
```
private Probe getProbe(INT InitialDelay) { ... }
```
and

Call it like
```
protected Probe getReadinessProbe() {
        int initialDelay = Configs.asInt(getConfig(Config.initialDelayReadiness));
        initialDelay = initialDelay < 0 ? DEFAULT_DELAY_READINESS : initialDelay;
        return getProbe(initialDelay);
}
```

dev-gaur: Correct me if I am wrong here.

lordofthejars: Yes, I agree with you, there is one enricher that is doing in the same way as I am doing here, but SpringBoot and thorntail are doing as you mention here, so I am changing.

PR #1463, opened 2018-12-19
feat(#729): Adds environment property 

### LINKED ISSUE #729: Environment specific OpenShift resources with some profile selection
### Description

I am trying to get a project set up where I can use the f-m-p to generate deployment controllers that are structurally different from environment to environment. I would like to have one layout in test without secrets in volumes, and then one prod with secrets mounted in volumes.

I am using resource fragments under src/main/fabric8 and am trying to use the fabric8 profiles and also some Maven profiles, but what ever I try of combinations I end up with getting resources generated from all my resource fragments regardless of how I structure them.

I am trying to place the different resource fragments under environment specific subdirectories:

src/main/fabric8/prod
src/main/fabric8/test

and was hoping I could then select one of these "profiles" with either a Maven profile pointing to only one fabric8 profile or some other trick.

Is this not something that is supported? And if not, then is there some other means where I can have structurally different resource fragments used for different environments.

Best regards
Lars Milland

### Info

* f-m-p version : 3.2.8

* Maven version (`mvn -v`) :
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T17:41:47+01:00)
Maven home: c:\maven
Java version: 1.8.0_60, vendor: Oracle Corporation
Java home: C:\Java\jdk1.8.0_60\jre
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 7", version: "6.1", arch: "amd64", family: "dos"

lordofthejars:
Fix #729 

Adds environment property so custom resources can be selected depending on the environment

### DIFF
--- a/it/src/it/custom-environment/verify.groovy
+++ b/it/src/it/custom-environment/verify.groovy
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+import io.fabric8.maven.it.Verify
+
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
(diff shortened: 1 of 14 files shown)


it/src/it/custom-environment/verify.groovy
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+import io.fabric8.maven.it.Verify
+
+/*
+ * Copyright 2016 Red Hat, Inc.
```
rohanKanojia: You might want to remove one of these ;-)

lordofthejars: Done

rohanKanojia:
@lordofthejars : I think you forgot to push :stuck_out_tongue_closed_eyes:

it/src/it/custom-environment/verify.groovy
dev-gaur: I can still see the copyright clause twice :sweat_smile: . Everything else LGTM.

bohmber:
This feature is more or less not very useful. 
For each env you would like to support you have to create a directory in src/main/fabric8 otherwise it will conflict with profiles. The most common use case, you would to override or add resources only for a few environments, is not working. 

If you change the ResourceDirCreator to check if the directory is available it will be a little bit better for the openshift overrides. But the openshift overrides are really limited. You can only add/replace something to existing files.  

```
public static File getFinalResourceDir(File resourceDir, String environment) {
    if (resourceDir != null && StringUtils.isNotEmpty(environment)) {
        final File file = new File(resourceDir, environment);
        if (file.exists()) {
            return file;
        }
    }
    return resourceDir;
}```

PR #1464, opened 2018-12-27
Fix #215: Change property prefixes from "docker." to "fabric8."

### LINKED ISSUE #215: Change property prefixes from "docker." to "fabric8."
This affects all Mojos which inherit from d-m-p Mojos (build, push, watch).
- Change in d-m-p from `@Parameter(property = ...)` to only `@Parameter`
- Add in AbstractDockerMojo a method `getProperty()` which in turn calls an overridable method `getPrefix()`
- Change from direct member field access to getter method which in turn use `getProperty()` from the abstract base class (+ lookup within the member variable of course)

rohanKanojia:
Fix #215 

We can merge this https://github.com/fabric8io/docker-maven-plugin/pull/1144 gets merged and dmp gets a new release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Feature 729: Select custom resources depending on environment.
 * Feature 601: Adds healthcheck for webapp
 * Fix 1460: Upgraded kubernetes client to 4.1.1
+* Fix 215: Change property prefixes from "docker." to "fabric8."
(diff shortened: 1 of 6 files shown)


manusa:
https://github.com/fabric8io/docker-maven-plugin/pull/1144 still not merged.
Is this still relevant @rohanKanojia ?

rohanKanojia:
I don't think it's relevant anymore since we have migrated docker-maven-plugin's BuildService to Jkube. In future, I would like to make DMP make use of JKube's BuildService.

PR #1474, opened 2019-01-04
Added 'fabric8.build.forcePull' parameter.

### LINKED ISSUE #255: Add option to forcefully reload a s2i builder image
This should be added to `fabric8:build`. See https://docs.openshift.com/enterprise/3.2/dev_guide/builds.html#s2i-force-pull for the option to set.

dev-gaur:
FIxes #255 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,11 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix 1460: Upgraded kubernetes client to 4.1.1
 * Fix 690: Removes deprecated _legacyPortMapping_ property.
 * Fix 1458: Support for from Image configuration in openshift docker build strategy
+
 * Fix 732: Added 'skip' options to goals.
 * Fix 1467: Wait timeout for build pod is too small
 * Fix 1466: Allow to configure noCache option for openshift docker build strategy
+* Fix 255: Add option to forcefully reload a s2i builder image.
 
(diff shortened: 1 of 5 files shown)


bohmber:
@dev-gaur  How this is related to ImagePullPolicy? If it's only openshift related the property name 'fabric8.build.forcePull' is misleading

rohanKanojia:
@dev-gaur : I agree with @bohmber  here, we should have some kind of more intuitive naming which tells that this property is for openshift builds only

dev-gaur:
@bohmber @rohanKanojia what name do you suggest? 

My thought process was :
"fabric8" - Mandatory prefix.
"build" - This parameter participates in the Build phase.
"forcePull" -  Its not just being used for ImageStream strategy.... This parameter is also being used in the Docker strategy! https://github.com/fabric8io/fabric8-maven-plugin/blob/be7376c3e078de404fe1621add5d3616fc41ef20/core/src/main/java/io/fabric8/maven/core/service/openshift/OpenshiftBuildService.java#L311

rohanKanojia:
Hm, I'm not so good with intuitive naming. If we name it like `fabric8.build.openshift.forcePull` it can become way too long. How about we add information about this flag in the documentation that it's only applicable on Openshift clusters?

dev-gaur:
Sounds fine

rohanKanojia:
@dev-gaur : There is a test failing. Could you please address failures there?

PR #1487, opened 2019-01-11
Fix #925: Unable to configure replicas count via property

### LINKED ISSUE #925: Unable to configure replicas count via property
### Description
When `replicas` is configured via fabric8 maven plugin `<resources>` configuration parameter, the plugin still sets the value as `1` which is the default value 

### Info

* f-m-p version : 3.3.5
* Maven version (`mvn -v`) : 3.5
* Kubernetes / OpenShift setup and version : 1.6/1.5

Define the following configuration in the maven project  $replicas: <any number>
```
 <plugin>
                <groupId>io.fabric8</groupId>
                <version>3.3.5</version>
                <executions>
                    <execution>
                        <id>fmp</id>
                        <goals>
                            <goal>resource</goal>
                            <goal>helm</goal>
                            <goal>build</goal>
                        </goals>
                        <configuration>
                            <resources>
                                <replicas>${replicas}</replicas>
                            </resources>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```
When you run the fabric8:resource, it still defaults to `1` as the value for replicas instead of what value provides via maven property `${replicas}`

rohanKanojia:
Fix #925 

XML properties(if set) should be given precedence over default ones.

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
@@ -97,10 +97,11 @@ public class DefaultControllerEnricher extends BaseEnricher {
     @Override
     public void addMissingResources(KubernetesListBuilder builder) {
         final String name = getConfig(Config.name, MavenUtil.createDefaultResourceName(getContext().getGav().getSanitizedArtifactId()));
-        ResourceConfig config = new ResourceConfig.Builder(getConfiguration().getResource().orElse(null))
+        ResourceConfig xmlResourceConfig = getConfiguration().getResource().orElse(null);
+        ResourceConfig config = new ResourceConfig.Builder(xmlResourceConfig)
                 .controllerName(name)
-                .imagePullPolicy(getConfig(Config.pullPolicy))
-                .withReplicas(Configs.asInt(getConfig(Config.replicaCount)))
+                .imagePullPolicy(getImagePullPolicy(xmlResourceConfig, getConfig(Config.pullPolicy)))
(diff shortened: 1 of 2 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
```diff
@@ -97,10 +97,11 @@ public DefaultControllerEnricher(MavenEnricherContext buildContext) {
     @Override
     public void addMissingResources(KubernetesListBuilder builder) {
         final String name = getConfig(Config.name, MavenUtil.createDefaultResourceName(getContext().getGav().getSanitizedArtifactId()));
-        ResourceConfig config = new ResourceConfig.Builder(getConfiguration().getResource().orElse(null))
+        ResourceConfig xmlResourceConfig = getConfiguration().getResource().orElse(null);
+        ResourceConfig config = new ResourceConfig.Builder(xmlResourceConfig)
```
lordofthejars: if `xmlResourceConfig` is null, is it protected in `ResourceConfig.Builder`?

rohanKanojia: umm, Could you please elaborate wdym by protected here?

rohanKanojia: here is `ResourceConfig` builder:
https://github.com/fabric8io/fabric8-maven-plugin/blob/21b78a929a64c21ccccee279feaef8c1dea4f318/core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java#L159-L180

lordofthejars: Yes there a null protection, so it is fine :)

PR #1488, opened 2019-01-14
Fix #1079: fabric8-includes does not seem to work with Karaf generator

### LINKED ISSUE #1079: fabric8-includes does not seem to work with Karaf generator
The JavaExecGenerator supports adding additional files to the docker image if they are put into src/main/fabric8-includes or src/main/fabric8-includes/bin  : 

https://github.com/fabric8io/fabric8-maven-plugin/blob/master/generator/java-exec/src/main/java/io/fabric8/maven/generator/javaexec/JavaExecGenerator.java#L196-L197

The KarafGenerator extends the BaseGenerator and not JavaExecGenerator and does not seem to support this.   

https://github.com/fabric8io/fabric8-maven-plugin/blob/master/generator/karaf/src/main/java/io/fabric8/maven/generator/karaf/KarafGenerator.java

The documentation is pretty clear that fabric8-includes is a feature, it should work for Karaf images as well.

### Info

* f-m-p version : 3.5.17
```

```

rohanKanojia:
Fix #1079

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix 732: Added 'skip' options to goals.
 * Fix 1467: Wait timeout for build pod is too small
 * Fix 1466: Allow to configure noCache option for openshift docker build strategy
+* Fix 1079: fabric8-includes does not seem to work with Karaf generator
 
 ### 4.0.0-M2 (2018-12-14)
 * Fix 10: Make VolumeConfiguration more flexible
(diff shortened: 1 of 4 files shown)


lordofthejars:
@rohanKanojia is there any easy way to test this integration so we can be sure all the time that we do not add any regression when changing anything either in include goal or karaf generator?

rohanKanojia:
@lordofthejars : ah, right. I couldn't find any sample or integration test regarding this generator. Would search on web

rohanKanojia:
Just tested it with  https://github.com/fabric8-quickstarts/karaf-camel-log, it's not working :disappointed: . Marking it WIP till I get it working

PR #1491, opened 2019-01-17
feat(#804): Allow set fragments as external references (URL)

### LINKED ISSUE #804: Allow for external resources reference; e.g. app tempplates
It would be useful to be able to include external k8s / OpenShift resources.
e.g. a SpringBoot app that requires RHT SSO could simply reference CE team's RHT SSO app template
* https://github.com/jboss-openshift/application-templates/blob/master/sso/sso70-https.json

(and not have to copy paste the resource file -- leave it to the original authors to be properly maintained)

@ro14nd ^^

lordofthejars:
Fix #804

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
+++ b/core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
@@ -15,6 +15,7 @@
  */
 package io.fabric8.maven.core.config;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -47,6 +48,9 @@ public class ResourceConfig {
 
     @Parameter
     private List<ServiceConfig> services;
+
+    @Parameter
+    private List<String> remotes;
 
     @Parameter
     private ConfigMap configMap;
@@ -149,6 +153,10 @@ public class ResourceConfig {
         return configMap;
     }
 
+    public List<String> getRemotes() {
+        return remotes;
+    }
+
     // =============================================================================================
 
     public static class Builder {
@@ -174,6 +182,7 @@ public class ResourceConfig {
                 this.config.services = config.getServices();
                 this.config.metrics = config.getMetrics();
                 this.config.namespace = config.getNamespace();
+                this.config.remotes = config.remotes;
             }
         }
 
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
```diff
@@ -15,6 +15,7 @@
  */
 package io.fabric8.maven.core.config;
 
+import java.util.ArrayList;
```
rohanKanojia: Unused import ;-)

it/src/it/remote-resources/verify.groovy
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+import io.fabric8.maven.it.Verify
+
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
```
rohanKanojia: I wish there license plugin could smartly detect and append existing headers rather than duplicating :thinking:

core/src/main/java/io/fabric8/maven/core/util/FileUtil.java
```diff
@@ -28,6 +30,14 @@
  */
 public class FileUtil {
 
+    public static File createTempDirectory() {
+        try {
+            return Files.createTempDirectory("fabric8").toFile();
```
rohanKanojia: It would be nice if we pass directory name as a parameter ;-)

lordofthejars: Currentñy this is not the directory, it is the prefix, Java creates a random directory in temporal space with `fabric8` prefix. So I think that we should not parametrize this method to maintain safely the temp directory to not populating with a different prefix.

core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
lordofthejars: Removed

core/src/main/java/io/fabric8/maven/core/util/FileUtil.java
rohanKanojia: Oh, are you creating file in `/tmp` directory?

PR #1494, opened 2019-01-21
Fix #1486: Remove Kompose Support.

### LINKED ISSUE #1486: Remove Kompose support
Remove Kompose support as it seems to be sunset now.

rohanKanojia:
Fix #1486 

Kompose has served it's purpose as a way to on-board new users from
Docker Compose to Kubernetes and OpenShift.


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -717,12 +702,9 @@ public boolean accept(File pathname) {
 
     private KubernetesListBuilder generateAppResources(List<ImageConfiguration> images, EnricherManager enricherManager)
         throws IOException, MojoExecutionException {
-        Path composeFilePath = checkComposeConfig();
-        ComposeService composeUtil = new ComposeService(komposeBinDir, composeFilePath, log);
         try {
             File[] resourceFiles = KubernetesResourceUtil.listResourceFragments(realResourceDir);
-            File[] composeResourceFiles = composeUtil.convertToKubeFragments();
-            File[] allResources = ArrayUtils.addAll(resourceFiles, composeResourceFiles);
+            File[] allResources = ArrayUtils.addAll(resourceFiles);
```
dev-gaur: Why not delete this line and replace allResources with resourceFiles ? This line is virtually redundant now.

lordofthejars: Yes this line should be deleted, although probably the compiler is already doing it for you.

Issue #718, opened 2016-12-07
WildFlySwarm - Detect port number from configuration file for health check


davsclaus:
Currently we use the default values of 8080. However they can be customized in the configuration.

But we should try to detect any of the various ways you can configure wildfly-swarm, and set its `swarm.http.port` configuration.

There is no consistent way like spring boot with application.properties so its a bit more tricky with swarm

rhuss:
This should be possible for Wildfly and Thorntail

lordofthejars:
The support has been implemented in Thorntail and not Swarm.

PR #1495, opened 2019-01-22
feat(#718): Implements Healthcheck port resolution for Thorntail

lordofthejars:
It has no sense to create healthcheck port for Swarm since the new platform is Thorntail.

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
@@ -47,7 +47,7 @@ public class SpringBootUtil {
         URL ymlResource = compileClassLoader.findResource("application.yml");
         URL propertiesResource = compileClassLoader.findResource("application.properties");
 
-        Properties props = getPropertiesFromYamlResource(ymlResource);
+        Properties props = YamlUtil.getPropertiesFromYamlResource(ymlResource);
         props.putAll(getPropertiesResource(propertiesResource));
         return props;
     }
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/fabric8/maven/core/util/SpringBootUtil.java
```diff
@@ -70,28 +70,7 @@ protected static Properties getPropertiesResource(URL resource) {
     /**
```
rohanKanojia: You should remove this comment too, since this method is removed ;-)

core/src/test/java/io/fabric8/maven/core/util/ThorntailUtilTest.java
```diff
@@ -0,0 +1,19 @@
+package io.fabric8.maven.core.util;
```
rohanKanojia: I think license plugin is crying because of this file

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/ThorntailV2HealthCheckEnricher.java
```diff
@@ -18,7 +18,10 @@
 import io.fabric8.kubernetes.api.model.Probe;
 import io.fabric8.kubernetes.api.model.ProbeBuilder;
 import io.fabric8.maven.core.util.Configs;
+import io.fabric8.maven.core.util.SpringBootUtil;
+import io.fabric8.maven.core.util.ThorntailUtil;
```
rohanKanojia: Not sure whether we need SpringBootUtil in this file

core/src/test/java/io/fabric8/maven/core/util/ThorntailUtilTest.java
lordofthejars: But what's happening?

lordofthejars: ok now :)

PR #1497, opened 2019-01-24
Refactor #828: Moved Metedata specific code into DefaultMetadataEnricher

### LINKED ISSUE #828: Move the MetadataEnricher to standard enrichers
We should move the code that adds selectors and labels to the generated resources from `EnricherManager` to a specific standard `MetadataEnricher`.

dev-gaur:
Fixes #828


dev-gaur:
> This looks good to me, I am curious why coverage is decreasing, if it is just a movement of code, but LGTM.

I think the movement of code has also changed the code flow a bit now. Maybe that's why the coverage has decreased. The part of code which was highly coupled to EnricherManager is now an independent enricher.

BTW, Are you fine with the fact that without including this particular enricher in the profile, user cannot use any of the metadata functions from other enrichers ?

PR #1498, opened 2019-01-24
Allows user to define Secret from annotation

bohmber:
Similar to fmp-configmap-file enricher for annotation

Signed-off-by: bohmber <bommel@apache.org>

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricher.java
@@ -1,0 +1,87 @@
+/**
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package io.fabric8.maven.enricher.standard;
+
+import io.fabric8.kubernetes.api.builder.TypedVisitor;
+import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
+import io.fabric8.kubernetes.api.model.SecretBuilder;
+import io.fabric8.maven.core.util.Base64Util;
+import io.fabric8.maven.enricher.api.BaseEnricher;
+import io.fabric8.maven.enricher.api.MavenEnricherContext;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+public class FileDataSecretEnricher extends BaseEnricher {
+
+    protected static final String PREFIX_ANNOTATION = "maven.fabric8.io/secret/";
+
+    public FileDataSecretEnricher(MavenEnricherContext buildContext) {
+        super(buildContext, "fmp-secret-file");
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricher.java
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+        });
+    }
+
+    private Map<String, String> createSecretFromAnnotations(final Map<String, String> annotations) throws IOException {
+        final Set<Map.Entry<String, String>> entries = annotations.entrySet();
+        final Map<String, String> secretFileLocations = new HashMap<>();
+
+        for(Iterator<Map.Entry<String, String>> it = entries.iterator(); it.hasNext(); ) {
```
dev-gaur: Hey @bohmber why not just use 

```
for (Map.Entry<String, String> entry : annotations.entrySet( ) )
```
?
The FMP code base uses this way of iteration everywhere, so it would look uniform with the codebase if you change it to this style.

enricher/standard/src/test/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricherTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class FileDataSecretEnricherTest {
+
+    private static final String TEST_APPLICATION_PROPERTIES_PATH = "src/test/resources/test-application.properties";
+    private static final String TEST_APPLICATION_PROPERTIES = "test-application.properties";
+    @Mocked
+    private MavenEnricherContext context;
+
+    @Test
+    public void should_materialize_file_content_from_annotation() throws IOException {
```
dev-gaur: please change the test name to camelCase.

dev-gaur:
Really appreciate you adding test and documentation with your code.

dev-gaur:
Shouldn't the annotation get deleted/removed after it is 'enriched' into a proper secret?
It won't look good if deploy with those annotations (plain text) in for eg.,a shared cluster maybe. Will defeat the purpose of that info being a secret.

I'm not a kubernetes/openshift expert in the matter though. Correct me if I'm wrong here.

Also, Changelog entry missing.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricher.java
bohmber: It's just a modified copy of the ConfigMapEnricher and i think the iterator is needed for the remove a few lines later 

`it.remove();`

enricher/standard/src/test/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricherTest.java
bohmber: I prefer camel case is well.  You should changed the method name in ConfigMapEnricherTest too.

bohmber:
@dev-gaur The annotation is deleted see my comment about the iterator and there is an assert in the unit test for it.

enricher/standard/src/test/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricherTest.java
bohmber: @dev-gaur done

bohmber:
@dev-gaur I can't add the change log entry before the pull request is created. I need the number and adding the change log entry early is causing more work on my side because i have to fix merge conflicts in the change log file. I can add the change log entry before the pull request will be merged

bohmber:
Is `FileDataSecretEnricher` a good class name?

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/FileDataSecretEnricher.java
dev-gaur: oh i see

dev-gaur:
> @dev-gaur The annotation is deleted see my comment about the iterator and there is an assert in the unit test for it.

aah missed that.


We'll handle the merge conflicts before merging. You can go ahead with the changelog entry.
No need for a number. You can just do with "Feature: allow users to define secrets from annotations."

bohmber:
@rohanKanojia changelog entry added

PR #1517, opened 2019-02-07
 update vmp groupid and vert.x version

rdruss:


rdruss:
I'll get that in today

Issue #802, opened 2017-02-03
Remove resource object post-processing


rhuss:
Currently in `ResourcesMojo` a lot of post processing happens with 

* Converting Kubernetes objects to OpenShift objects
* Converting OpenShift to Kubernetes objects
* Filtering out Kubernetes objects (for openshift.yml)
* Filtering out OpenShift objects (for kubernetes.yml)

All of this is not only complicated and lead to a lot of special conditional cases (`if instanceof ...`), but also hard to debug e.g. **why** a certain objects exists in the generated descriptors.

It is also the natural entry point for **quick hacks** which quickly pleases a certain use case but is pain and harmful in the long run when other use cases (which even might not be envisioned) happens. The _volume permission init container_ was a good example when it was implemented as a post processing step. Also, such special behaviour is not easy to disable, except with a new configuration parameter (which leads to configurtion fragmentation, blows up the mojo, needs to be documented, needs to be added to the various way to configure stuff). Adding to the configuration should be really the last option.

Instead all _functional processing_ should go into the `Enrichers`. See also #678 for more details and a solution to this problem.

rohanKanojia:
Closed via #1521

PR #1521, opened 2019-02-13
Refactor #802: Remove resource object post processing

rohanKanojia:
Refactor #678 #802


core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
```diff
@@ -257,6 +259,8 @@ public final static void updateKindFilenameMapper(final Map<String, List<String>
             apiVersion = apiVersions.getAppsVersion();
         } else if (Objects.equals(kind, "Job")) {
             apiVersion = apiVersions.getJobVersion();
+        } else if(Objects.equals(kind, "DeploymentConfig") && platformMode == PlatformMode.openshift) {
```
lordofthejars: Can you add an space before `(`

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
```diff
@@ -17,28 +17,18 @@
...
 import io.fabric8.maven.core.config.ResourceConfig;
-import io.fabric8.maven.core.handler.DaemonSetHandler;
-import io.fabric8.maven.core.handler.DeploymentHandler;
-import io.fabric8.maven.core.handler.HandlerHub;
-import io.fabric8.maven.core.handler.JobHandler;
-import io.fabric8.maven.core.handler.ReplicaSetHandler;
-import io.fabric8.maven.core.handler.ReplicationControllerHandler;
-import io.fabric8.maven.core.handler.StatefulSetHandler;
+import io.fabric8.maven.core.config.RuntimeMode;
+import io.fabric8.maven.core.handler.*;
```
lordofthejars: Can we avoid star operator?

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultControllerEnricher.java
```diff
@@ -181,4 +204,20 @@ private String getImagePullPolicy(ResourceConfig xmlResourceConfig, String defau
...
+    private Boolean isAutomaticTriggerEnabled(Boolean defaultValue) {
+        if(((MavenEnricherContext)getContext()).getProperty("fabric8.openshift.enableAutomaticTrigger") != null) {
+            return Boolean.parseBoolean(((MavenEnricherContext) getContext()).getProperty("fabric8.openshift.enableAutomaticTrigger").toString());
+        } else {
+            return defaultValue;
+        }
+    }
+
+    private Long getOpenshiftDeployTimeoutInSeconds(Long defaultValue) {
+        if (((MavenEnricherContext)getContext()).getProperty("fabric8.openshift.deployTimeoutSeconds") != null) {
```
lordofthejars: Do we really need this cast? Maybe the `getProperty` method should be in base class so we do not need to cast to `MavenEnricherContext`. At least we are not tighting to Maven classes. Same in line 210.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/PodAnnotationEnricher.java
```diff
@@ -57,6 +60,21 @@ public void adapt(KubernetesListBuilder builder) {
                         templateMetadata.setAnnotations(MapUtil.mergeMaps(templateMetadata.getAnnotations(), metadata.getAnnotations()));
                     }
                 }
+            } else if(platformMode == PlatformMode.openshift && item instanceof DeploymentConfig) {
```
lordofthejars: Space after `if`

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/PodAnnotationEnricher.java
```diff
@@ -57,6 +60,21 @@ public void adapt(KubernetesListBuilder builder) {
...
+                DeploymentConfig deploymentConfig = (DeploymentConfig)item;
+                ObjectMeta metadata = deploymentConfig.getMetadata();
+                DeploymentConfigSpec spec = deploymentConfig.getSpec();
+                if(metadata != null && spec != null) {
```
lordofthejars: Spaces after `ifs`

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/openshift/DeploymentConfigEnricher.java
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+
+            specBuilder.endSpec();
+        }
+        return builder.build();
+    }
+
+    private Boolean isAutomaticTriggerEnabled(MavenEnricherContext enricherContext, Boolean defaultValue) {
+        if(enricherContext.getProperty("fabric8.openshift.enableAutomaticTrigger") != null) {
```
lordofthejars: If `getProperty` can be called without casting then we will not have a Maven dependency anymore.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/Enricher.java
```diff
@@ -63,14 +64,14 @@
      *
      * @param builder the build to examine and add to
      */
-    void addMissingResources(KubernetesListBuilder builder);
+    void addMissingResources(PlatformMode platformMode, KubernetesListBuilder builder);
```
rhuss: The idea is the right one, but I'm missing the full implementation of #678 , i.e. removal of getLabels(), getAnnotations() and getSelector(), and moving to a create() and adapt() phase (as  described in #678). Is this something planned for later ?

core/src/main/java/io/fabric8/maven/core/config/RuntimeMode.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Use special OpenShift features like BuildConfigs
+     */
+    openshift(false, "OpenShift"),
+
+    /**
+     * Detect automatically whether running on OpenShift or Kuberentes.
+     * This is done by contacting an API server
+     */
+    auto(true, "Auto");
```
rhuss: Iiur then "PlatformMode" is used for creating the resources (i.e. for doing the two passes through the enricher chain), and "RuntimeMode" is for deployment (i.e. to select which client to use for applying resources).

What does then "auto" mean here ? I would also recommend to adapt the documentation to better reflect this distinction.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -293,10 +278,8 @@
     // Access for creating OpenShift binary builds
     private ClusterAccess clusterAccess;
 
-    private PlatformMode platformMode;
+    private RuntimeMode runtimeMode;
```
rhuss: There must be no reference to the runtimeMode in ResourceMojo. Do you know why this is needed ?

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -416,18 +399,17 @@ public void executeInternal() throws MojoExecutionException, MojoFailureExceptio
...
-                validateIfRequired(openShiftResourceDir, ResourceClassifier.OPENSHIFT);
-
-                // Remove OpenShift specific stuff provided by fragments
-                KubernetesList kubernetesResources = convertToKubernetesResources(resources, openShiftResources);
-                writeResources(kubernetesResources, ResourceClassifier.KUBERNETES, generateRoute);
-                File kubernetesResourceDir = new File(this.targetDir, ResourceClassifier.KUBERNETES.getValue());
-                validateIfRequired(kubernetesResourceDir, ResourceClassifier.KUBERNETES);
+
+                KubernetesList resources;
+                for(PlatformMode platformMode : new PlatformMode[] { PlatformMode.kubernetes, PlatformMode.openshift }) {
```
rhuss: 👍

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -416,18 +399,17 @@ public void executeInternal() throws MojoExecutionException, MojoFailureExceptio
             resolvedImages = getResolvedImages(images, log);
             if (!skip && (!isPomProject() || hasFabric8Dir())) {
                 // Extract and generate resources which can be a mix of Kubernetes and OpenShift resources
-                KubernetesList resources = generateResources(resolvedImages, remoteResources);
-                // Adapt list to use OpenShift specific resource objects
-                KubernetesList openShiftResources = convertToOpenShiftResources(resources);
```
rhuss: 👍 yeah, finally ;-)

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -497,14 +479,14 @@ private void lateInit() {
                 properties.setProperty(DOCKER_IMAGE_USER, namespace);
             }
             if (!properties.contains(PlatformMode.FABRIC8_EFFECTIVE_PLATFORM_MODE)) {
-                properties.setProperty(PlatformMode.FABRIC8_EFFECTIVE_PLATFORM_MODE, platformMode.toString());
+                properties.setProperty(PlatformMode.FABRIC8_EFFECTIVE_PLATFORM_MODE, runtimeMode.toString());
```
rhuss: That should not be needed.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -497,14 +479,14 @@ private void lateInit() {
...
             }
         }
 
         openShiftConverters = new HashMap<>();
         openShiftConverters.put("ReplicaSet", new ReplicSetOpenShiftConverter());
         openShiftConverters.put("Deployment",
-            new DeploymentOpenShiftConverter(platformMode, getOpenshiftDeployTimeoutSeconds()));
+            new DeploymentOpenShiftConverter(runtimeMode, getOpenshiftDeployTimeoutSeconds()));
```
rhuss: Do we still need this converters ? I though the PR was all about getting rid of them

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
rohanKanojia: It's used to find out which runtime we're using here and setting some properties(Maybe they were relevant in fabric8 context): https://github.com/fabric8io/fabric8-maven-plugin/blob/19c368936e53a88d402ff24c553855add5313a85/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java#L489

rohanKanojia: Plus we add ImageChange triggers in deployment config based on runtime mode: https://github.com/fabric8io/fabric8-maven-plugin/blob/19c368936e53a88d402ff24c553855add5313a85/plugin/src/main/java/io/fabric8/maven/plugin/converter/DeploymentOpenShiftConverter.java#L117-L127

core/src/main/java/io/fabric8/maven/core/handler/DeploymentConfigHandler.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import io.fabric8.kubernetes.api.model.*;
+import io.fabric8.maven.core.config.ResourceConfig;
+import io.fabric8.maven.core.config.RuntimeMode;
+import io.fabric8.maven.core.util.kubernetes.KubernetesHelper;
+import io.fabric8.maven.docker.config.ImageConfiguration;
+import io.fabric8.maven.docker.util.ImageName;
+import io.fabric8.openshift.api.model.*;
```
lordofthejars: Avoid star import.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
rhuss: We should really try hard to get rid of this. I.e. we should check who is using this property and seek for a solution which doesn't involve this runtime mode.

The reason why I'm arguing so hard for removing it is that it can't be that the resources are created differently depending on whether you are connected to a Kuberntes or OpenShift cluster. This totally defeats repeatable builds.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/Enricher.java
```diff
@@ -63,14 +37,12 @@
      *
      * @param builder the build to examine and add to
      */
-    void addMissingResources(KubernetesListBuilder builder);
+    void enrich(PlatformMode platformMode, KubernetesListBuilder builder);
```
rhuss: Looks very nice now ! It has now the two inner passes from the initial suggestion.

Wdyt about renaming these two methods to "create" (was "enrich") and to "enrich" (was "adapt") as suggested in #678 ?

That way we can clearly indicate the intension of this API, too.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/EnricherContext.java
```diff
@@ -109,8 +111,12 @@ default boolean hasDependency(String groupId, String artifactId) {
         return Optional.empty();
     }
 
-    // ===========================================================================================
-    // To be removed:
 
-    OpenShiftDependencyResources getOpenshiftDependencyResources();
+    PlatformMode getPlatformMode();
+
+    RuntimeMode getRuntimeMode();
```
rhuss: This must not be there. (see other comments)

enricher/api/src/main/java/io/fabric8/maven/enricher/api/EnricherContext.java
```diff
@@ -109,8 +111,12 @@ default boolean hasDependency(String groupId, String artifactId) {
...
+
+    Object getProperty(String key);
+
+    Properties getProperties();
```
rhuss: What's about having only one of these methods to keep the API slim? 

Alternatively, you could make `getProperty()` a "default" interface method, too.

And don't forget to add JavaDoc. While I wouldn't insist on having docs everywhere, for public API which are used e.g. for implementing Enrichers, good documentation is mandatory.

plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
rhuss: IMO the DeploymentOpenShiftConverter should either use the new PlatformMode or we should question why it adds the image trigger here. If this is because e.g. needed for 'fabric8:build' triggering a redeployment, then this functionality should go to 'fabric8:build' to patch existing deployment configs, or even trigger a redeployment on their own.

rohanKanojia:
@rhuss : I've filed an issue related to refactoring GeneratorContext out of ResourceMojo which can be done later: https://github.com/fabric8io/fabric8-maven-plugin/issues/1528

PR #1576, opened 2019-03-16
feature(resource): Remove runtimeMode from fabric8:resource

rhuss:
Also bumped up snapshot version to 4.1-SNAPSHOT

PR #1577, opened 2019-03-17
feature: Add Quarkus generator and health check enricher

rhuss:


rhuss:
These are my changes which I added to make Quarkus run out of the box with 4.1

Actually its a bit more (sorry, there has been some changes slipped in, I'm a bit in a hurry):

* Includes also #1576
* Switched the order for picking up properties for generators: Now system properties (provided from the outside) override project properties from the pom. Actually I consider the former behaviour a bug as it was not possible to ovveride a property with -D when set in <properties>.
This Quarkus support uses either the java mode or, when "native" config is set, the native GraalVM mode for building the image. Also, if smallrye-healthcheck is added as a dependency, liveness and readiness checks are added.

See also the sample I added.

The generator could be improved: Currently, it detects only works when it detects the appropriate jars in target/ to find out the file names. However, since the generators are also run for enriching, and enriching typically happens before compile/package (so no jars are there already), the generator should get the name of the generated jars from the Maven pom. Just had no time to find out, how Quarkus names these jars and how to get it from the pom.

enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/QuarkusHealthCheckEnricher.java
```diff
@@ -0,0 +1,86 @@
...
+ * permissions and limitations under the License.
+ */
+package io.fabric8.maven.enricher.fabric8;
+
+import io.fabric8.kubernetes.api.model.Probe;
+import io.fabric8.kubernetes.api.model.ProbeBuilder;
+import io.fabric8.maven.core.util.Configs;
+import io.fabric8.maven.enricher.api.MavenEnricherContext;
+
+import static io.fabric8.maven.core.util.Configs.asBoolean;
```
rohanKanojia: unused import :-)

generator/java-exec/src/main/java/io/fabric8/maven/generator/javaexec/JavaExecGenerator.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Map;
 
 import io.fabric8.maven.core.util.Configs;
+import io.fabric8.maven.core.util.FileUtil;
```
rohanKanojia: here as well

generator/quarkus/src/main/java/io/fabric8/maven/generator/quarkus/QuarkusGenerator.java
```diff
@@ -0,0 +1,170 @@
...
+        boolean isNative = Boolean.parseBoolean(getConfig(Config.nativeImage, "false"));
+
+        if (isNative) {
+            buildBuilder.from("registry.fedoraproject.org/fedora-minimal")
+                        .entryPoint(new Arguments.Builder()
+                                        .withParam("./" + findSingleFileThatEndsWith("-runner"))
+                                        .withParam("-Dquarkus.http.host=0.0.0.0")
+                                        .build())
+                        .workdir("/");
+
```
rohanKanojia: Would be nice in my opinion if we keep these values in static constants..

core/src/main/java/io/fabric8/maven/core/util/BuildLabelUtil.java
```diff
@@ -66,13 +66,15 @@ public static void addSchemaLabels (BuildImageConfiguration.Builder buildBuilder
...
         try {
             Repository repository = GitUtil.getGitRepository(project.getBasedir());
-            String commitID = GitUtil.getGitCommitId(repository);
-            labels.put(BuildLabelAnnotations.VCS_REF.value(), commitID);
-            String gitRemoteUrl = repository.getConfig().getString("remote", GIT_REMOTE, "url");
-            if (gitRemoteUrl != null) {
-                labels.put(BuildLabelAnnotations.VCS_URL.value(), gitRemoteUrl);
-            } else {
-                log.warn("Could not detect any git remote");
+            if (repository != null) {
```
rohanKanojia: :+1:

Issue #1579, opened 2019-03-18
4.0.0 is generating an unwanted sidecar copy of application instance


mojsha:
### Description

It looks like 4.0.0 is "auto-generating" an unwanted sidecar copy of the application instance. 

It is easiest explained using screenshots:

Using 3.5.42, the application is running in container "demo":
![image](https://user-images.githubusercontent.com/27762179/54524741-02e2de80-4973-11e9-8382-3590a19f7208.png)

Using 4.0.0, two instance of the application now are running, demo (as defined in a DC) and "demo-sample-t" (a concatenation of the DC name "demo" and namespace name "sample-t"):

![image](https://user-images.githubusercontent.com/27762179/54524800-23129d80-4973-11e9-8f34-a4c3450e6060.png)

As you can see, the "sample-t-demo" (sample-t is the name of the namespace) has been generated without any resource fragment.

### Info

* f-m-p version :
4.0.0
* Kubernetes / OpenShift setup and version :
3.10
* If it's a bug, how to reproduce :
Use a DC and upgrade to 4.0.0 and do the regular fabric8:resource fabric8:build fabric8:apply cycle. It should generate the unwanted sidecar.

@rohanKanojia

rohanKanojia:
@mojsha : I can confirm that this is certainly a bug which needs to be fixed. This is a sideeffect of #1202 since fmp now considers container in fragment as a sidecar. We need to have some option in order to disable this autogeneration of the main container.

rohanKanojia:
Hm, this is being done via `ControllerViaPluginConfigurationEnricher` which was earlier a part of DefaultControllerEnricher( see #401 ). You can get rid of this sidecar by excluding this enricher in your plugin configuration.

```
          <enricher>
            <includes>
              <include>fmp-portname</include>
            </includes>
            <excludes>
              <exclude>build</exclude>
              <exclude>fmp-controller</exclude>
              <exclude>fmp-controller-from-configuration</exclude>
            </excludes>
            <config>
              <f8-service>
                <type>NodePort</type>
              </f8-service>
            </config>
          </enricher>

```

rohanKanojia:
@mojsha: Okay, so this enricher is creating two separate containers instead of one. So the idea of sidecar containers was to treat every container as sidecar whose name is not equal to default application container(which is generated by fmp). So if you take a resource fragment like this:
```
kind: "DeploymentConfig"
apiVersion: "v1"
spec:
  template: 
    spec:
[...]

PR #1610, opened 2019-04-02
Add ImageChange triggers to application container when resource fragments are used

rohanKanojia:
+ Actually we only mark container as main application container when it's generated using
  handlers, so we should also mark it after we use resource fragments. I've added a step
  in the step while merging default generated and resource fragment spec.

+ Don't mark containers from resource fragments as FABRIC8_GENERATED

+ Marking all containers as FABRIC8_GENERATED breaks functionality of some
  health check enrichers, like docker-health-check-enricher. So add them to
  a different key in pi so that imagechangeenricher can detect.

Related to #1579


core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
```diff
@@ -701,6 +701,7 @@ public static void mergePodSpec(PodSpecBuilder builder, PodSpec defaultPodSpec,
             }
             builder.withContainers(containers);
         }
+        return defaultApplicationContainerName; // Return the main application container's name.
```
lordofthejars: This can return a `null`, is it fine?

rohanKanojia: I am checking null return value in ControllerViaPluginConfigurationEnricher, null value would indicate that enricher hasn't processed(merging default generated container spec with spec provided in resource fragments) controller. It's just here to make us aware of which container is main application container.

If container is specified in resource fragment, this would be set. Otherwise it would be null, in that case FABRIC8_GENERATED processing instruction would be set via handlers

PR #1609, opened 2019-04-01
Refactor for Namespace and Project

dev-gaur:
Fixes #1608
Fixes #1611
Fixes #1470

PR #1614, opened 2019-04-08
Enable namespace metadata annotation when generating resource via fragments

dev-gaur:
in #1609 I forgot to deal with use case when namespace is generated via yaml fragment. In this case, the other  generated resources should be generated with respective namespace annotation as well, which is not the case, currently.

```
namespace.yml

metadata:
  name: fragment-ns

---------------------------------------
deployment.yml

metadata:
  namespace: fragment-ns
```

Also, the apply goal wasn't working as expected.

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultNamespaceEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultNamespaceEnricher.java
@@ -18,6 +18,7 @@ package io.fabric8.maven.enricher.standard;
 import io.fabric8.kubernetes.api.builder.TypedVisitor;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 import io.fabric8.kubernetes.api.model.Namespace;
+import io.fabric8.kubernetes.api.model.NamespaceBuilder;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
 import io.fabric8.maven.core.config.PlatformMode;
 import io.fabric8.maven.core.config.ResourceConfig;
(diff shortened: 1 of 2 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultNamespaceEnricher.java
```diff
@@ -95,13 +99,47 @@ public void create(PlatformMode platformMode, KubernetesListBuilder builder) {
...
         builder.accept(new TypedVisitor<ObjectMetaBuilder>() {
+            private String getNamespaceName() {
+                String name = null;
+                if (config.getNamespace() != null && !config.getNamespace().isEmpty()) {
+                    name = config.getNamespace();
+                }
+
+                name = builder.getItems().stream()
+                        .filter(item -> Arrays.asList(NAMESPACE_KINDS).contains(item.getKind()))
+                        .findFirst().get().getMetadata().getName();
```
lordofthejars: `findFirst` returns an `Optional` which might contain or not the element, so if it does not contain `getMetadata` will return a NPE.

dev-gaur: HI Alex, https://github.com/fabric8io/fabric8-maven-plugin/pull/1614/files#diff-6e4d1897bfc7a57cabed4541720c8ab0R117 I'm doing a check for Namespace and Project Kind here, before calling this private method. So null case will never occur.

lordofthejars: thanks for the clarification

PR #1625, opened 2019-04-22
fixing #issue872

### LINKED ISSUE #872: Using spring.app.name as servicename
The spring boot applications has the option of specifying the `spring.app.name` property, when thats being defined the f8-m-p can use that as service name instead of using `${project.artifactId}`

### Info

* f-m-p version : 3.2.30
```

```
* Kubernetes / OpenShift setup and version : 1.4.1/3.4.0

theexplorist:
fixes #872

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
@@ -156,8 +156,10 @@ public class DefaultServiceEnricher extends BaseEnricher {
     }
 
     private String getServiceName() {
-        if (getAppName() != null) {
-            return getAppName();
+
+        String appName = getAppName();
+        if (appName != null) {
+            return appName;
         } else {
             return getConfig(Config.name, MavenUtil.createDefaultResourceName(getContext().getGav().getSanitizedArtifactId()));
         }
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
@@ -39,7 +39,15 @@ import io.fabric8.maven.docker.config.ImageConfiguration;
 import io.fabric8.maven.enricher.api.BaseEnricher;
 import io.fabric8.maven.enricher.api.MavenEnricherContext;
 import java.io.IOException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -135,7 +143,7 @@ public class DefaultServiceEnricher extends BaseEnricher {
         }
     }
 
-    private String getAppName(){
+    private String getAppName() {
         try {
             if (getContext().getProjectClassLoaders().isClassInCompileClasspath(true)) {
                 Properties properties = SpringBootUtil.getSpringBootApplicationProperties(getContext().getProjectClassLoaders().getCompileClassLoader());
@@ -147,10 +155,10 @@ public class DefaultServiceEnricher extends BaseEnricher {
         return null;
     }
 
-    private String getServiceName(){
-        if(getAppName() != null){
+    private String getServiceName() {
+        if (getAppName() != null) {
             return getAppName();
-        }else{
+        } else {
             return getConfig(Config.name, MavenUtil.createDefaultResourceName(getContext().getGav().getSanitizedArtifactId()));
         }
     }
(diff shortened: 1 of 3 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -32,20 +32,14 @@
...
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
dev-gaur: No wildcard imports!

dev-gaur: You can disable the wildcard imports in your IDE so save yourself from the hassle next time.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -141,14 +135,34 @@ private void addServices(KubernetesListBuilder builder) {
...
+                Properties properties = SpringBootUtil.getSpringBootApplicationProperties(getContext().getProjectClassLoaders().getCompileClassLoader());
+                return properties.getProperty("spring.application.name");
+            }
+        } catch (Exception ex) {
+            log.error("Error while reading the spring-boot configuration", ex);
+        }
+        return null;
+    }
+
+    private String getServiceName(){
```
dev-gaur: space missing between ( ) and {

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -141,14 +135,34 @@ private void addServices(KubernetesListBuilder builder) {
...
+        if(getAppName() != null){
```
dev-gaur: spaces missing .. `if (getAppName( ) != null) {`

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -141,14 +135,34 @@ private void addServices(KubernetesListBuilder builder) {
...
+            return getAppName();
+        }else{
```
dev-gaur: `} else {`

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -141,14 +135,34 @@ private void addServices(KubernetesListBuilder builder) {
         }
     }
 
+    private String getAppName(){
+        try {
+            if (getContext().getProjectClassLoaders().isClassInCompileClasspath(true)) {
+                Properties properties = SpringBootUtil.getSpringBootApplicationProperties(getContext().getProjectClassLoaders().getCompileClassLoader());
+                return properties.getProperty("spring.application.name");
+            }
+        } catch (Exception ex) {
```
dev-gaur: Can you narrow down to a more specific Exception Class type ? Also, Just `e` suffices as the variable name here.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -141,14 +135,34 @@ private void addServices(KubernetesListBuilder builder) {
         }
     }
 
+    private String getAppName(){
+        try {
+            if (getContext().getProjectClassLoaders().isClassInCompileClasspath(true)) {
+                Properties properties = SpringBootUtil.getSpringBootApplicationProperties(getContext().getProjectClassLoaders().getCompileClassLoader());
+                return properties.getProperty("spring.application.name");
```
dev-gaur: @ro14nd @lordofthejars @rohanKanojia can you provide other property names that we should consider for the name ? 

* spring.app.name
* app.name
* spring.service.name ?

WDYT ?

rohanKanojia: we should stick to standard spring properties, which is `spring.application.name` in this case. If there are other ways there, then we can add them here as well.

dev-gaur: yeah, that's what I meant.. 

I think `spring.app.name` should be considered as well

rohanKanojia:
@dev-gaur : Can you please take a look if there is anything pending from your side? Otherwise, can we proceed to merge this PR?

theexplorist:
> @dev-gaur : Can you please take a look if there is anything pending from your side? Otherwise, can we proceed to merge this PR?

I will update the PR with requested changes by today so it can be good to merge.

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
dev-gaur: @theexplorist can you add a `getProperty("spring.app.name")` as well? Even in #872 its mentioned `spring.app.name` .

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -141,14 +143,34 @@ private void addServices(KubernetesListBuilder builder) {
...
+            }
+        } catch (Exception ex) {
+            log.error("Error while reading the spring-boot configuration", ex);
+        }
+        return null;
+    }
+
+    private String getServiceName() {
+        if (getAppName() != null) {
```
lordofthejars: Can you skip calling `getAppName` twice? Do something like:

```
String appName = getAppName();
if (appName!=null) {
return appName;
} else {
....
}
```

After that, I think we are ready to merge.

PR #1631, opened 2019-05-06
Added ingress support for fabric8:resource

dev-gaur:
Still WIP as I've not been able to test ingress properly on my minikube setup


plugin/src/main/resources/META-INF/fabric8/profiles-default.yml
```diff
@@ -27,6 +27,7 @@
     - fmp-controller
     - fmp-controller-from-configuration
     - fmp-service
+    - fmp-ingress
```
rohanKanojia: I would refrain from adding ingress to default profile since it needs controller setup also

dev-gaur: removed

rohanKanojia:
@dev-gaur : Were you able to test it properly on your minikube setup? Is it still a WIP?

core/src/main/java/io/fabric8/maven/core/config/ResourceConfig.java
```diff
@@ -91,6 +92,8 @@
     @Parameter
     private List<ServiceAccountConfig> serviceAccounts;
 
+    private List<IngressRule> ingressRules;
```
rohanKanojia: @dev-gaur : Why are you using a kuberenetes model class instead of a custom config class used in other xml config resources here?

PR #1633, opened 2019-05-11
fixing issue 1632

### LINKED ISSUE #1632: imagePullPolicy configuration not working
### Info
I have a following configuration as mentioned by documentation.
```
<plugin>
<groupId>io.fabric8</groupId>
<version>${fmp.version}</version>
<configuration>
    <images>
        <image>
            <name>username/repository:${project.build.finalName}</name>
            <build>
                <from>java:8</from>
                <maintainer>maintainer</maintainer>
                <assembly>
                    <name>assembly</name>
                    <targetDir>/user/local/projects</targetDir>
                    <descriptorRef>artifact</descriptorRef>
                </assembly>
                <entryPoint>
                    <exec>
                        <arg>java</arg>
                        <arg>-Xmx128m</arg>
                        <arg>-jar</arg>
                        <arg>/user/local/projects/${project.build.finalName}.jar</arg>
                    </exec>
                </entryPoint>
                <ports>
                    <port>${port}</port>
                </ports>
                <imagePullPolicy>Always</imagePullPolicy>
            </build>
        </image>
    </images>
</configuration>
</plugin>
```
Also my deployment.yml is minimal.

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ${project.artifactId}
  labels:
    app: ${project.artifactId}
spec:
  replicas: 1
  selector:
    matchLabels:
      app: ${project.artifactId}
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: ${project.artifactId}
```

Everything works fine, only imagePullPolicy is always ifNotPresent in generated deployment.yml file. I am trying to make it to Always as you can see in configuration file. I tried using enricher as well as mentioned in this [https://stackoverflow.com/questions/45746984/how-to-configure-imagepullpolicy-from-fabric8-maven-plugin](solution), but no success yet.

* f-m-p version : 4.1.0

**UPDATE**
The above mentioned stackoverflow solution works only in case we don't have explicitly written deployment.yml under fabric8 folder. In case we are enriching already existing deployment.yml file, we must use `fmp-controller-from-configuration`.

theexplorist:
Fix #1632

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ControllerViaPluginConfigurationEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ControllerViaPluginConfigurationEnricher.java
@@ -71,7 +71,7 @@ public class ControllerViaPluginConfigurationEnricher extends BaseEnricher {
         ResourceConfig xmlResourceConfig = getConfiguration().getResource().orElse(null);
         final ResourceConfig config = new ResourceConfig.Builder()
                 .controllerName(name)
-                .imagePullPolicy(getConfig(Config.pullPolicy))
+                .imagePullPolicy(getImagePullPolicy(xmlResourceConfig, getConfig(Config.pullPolicy)))
                 .withReplicas(getReplicaCount(builder, xmlResourceConfig, Configs.asInt(getConfig(Config.replicaCount))))
                 .build();
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 
 ### 4.1-SNAPSHOT
 * Update Docker Maven Plugin to 0.29.0
-
+* Fix 1632: imagePullPolicy configuration fixed as per documentation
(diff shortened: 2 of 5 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/ControllerViaPluginConfigurationEnricher.java
```diff
@@ -144,4 +144,18 @@ private void mergeStatefulSetSpec(StatefulSetBuilder builder, StatefulSetSpec sp
...
 
+    /**
+     * This method overrides the ImagePullPolicy value by the value provided in
+     * XML config.
+     *
+     * @param resourceConfig
+     * @param defaultValue
+     * @return
+     */
+    private String getImagePullPolicy(ResourceConfig resourceConfig, String defaultValue) {
```
rohanKanojia: I would suggest moving this function to it's parent class i.e `BaseEnricher` since both `DefaultControllerEnricher` and `ControllerViaPluginConfiguration` are using this method

PR #1638, opened 2019-05-16
Chore: Remove enrichAll flag from ImageChangeTriggerEnricher config

rohanKanojia:
After discussion with @mojsha about the problems he is facing, I'm removing this parameter from Enricher config and adding it like a mojo parameter. Somehow maven doesn't recognize it when parameter contains "-" in the name.


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -225,6 +225,9 @@
     @Parameter(property = "fabric8.namespace")
     private String namespace;
 
+    @Parameter(property = "fabric8.sidecar")
+    private String sidecar;
```
lordofthejars: It should be a boolean with default to false instead of String and parse it later.

PR #1643, opened 2019-05-24
added service port normalization

### LINKED ISSUE #73: default to port 80 for exposed service ports for targetPort 8080?
we currently generate services like this for spring boot:

``` yaml

---
apiVersion: "v1"
kind: "Service"
metadata:
  name: "spring-boot-web"
spec:
  ports:
  - port: 8080
    protocol: "TCP"
    targetPort: 8080
```

if we have only 1 port its > 1000 and there's no explicit configuration I wonder if we should assume port 80? Or at least we can assume 8080 should be exposed as port 80 by default

dev-gaur:
fixes #73

### DIFF
--- a/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
+++ b/enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
@@ -26,10 +26,11 @@ import io.fabric8.kubernetes.api.model.ServiceFluent;
 import io.fabric8.kubernetes.api.model.ServicePort;
 import io.fabric8.kubernetes.api.model.ServicePortBuilder;
 import io.fabric8.kubernetes.api.model.ServiceSpec;
+import io.fabric8.maven.core.config.PlatformMode;
 import io.fabric8.maven.core.config.ResourceConfig;
 import io.fabric8.maven.core.config.ServiceConfig;
+import io.fabric8.maven.core.handler.HandlerHub;
 import io.fabric8.maven.core.handler.ServiceHandler;
-import io.fabric8.maven.core.config.PlatformMode;
 import io.fabric8.maven.core.util.Configs;
 import io.fabric8.maven.core.util.MavenUtil;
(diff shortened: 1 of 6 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricher.java
```diff
@@ -101,30 +116,59 @@ public DefaultServiceEnricher(MavenEnricherContext buildContext) {
...
+            @Override
+            public void visit(ServicePortBuilder portBuilder) {
+
+
+                PORT_NORMALIZATION_MAPPING.keySet().forEach(key -> {
+                    if (key.intValue() == portBuilder.buildTargetPort().getIntVal()) {
+                        portBuilder.withPort(PORT_NORMALIZATION_MAPPING.get(key));
+                    }
+                });
+            /*
```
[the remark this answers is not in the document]
dev-gaur: Aah. sorry to forgot to remove that. Fixed now

enricher/standard/src/test/java/io/fabric8/maven/enricher/standard/DefaultServiceEnricherTest.java
```diff
@@ -97,6 +97,17 @@ public void portConfigWithMultipleMappings() throws Exception {
...
+    @Test
+    public void portConfigWithMultipleMapping1() throws JsonProcessingException {
+        setupExpectations("port", "8080:8081,8443:8443", "multiPort", "true", "normalizePort", "true");
+
+        String json = enrich();
+        assertPort(json, 0, 80, 8081, "http", "TCP");
+        assertPort(json, 1, 443, 8443, "https", "TCP");
+        assertThat(json, hasJsonPath("$.spec.ports[*]", hasSize(2)));
+    }
+
```
lordofthejars: Can you add a use case when normalize is `false`?

dev-gaur: The normalize is false by default... So all other test cases are normalize = false.

PR #1646, opened 2019-06-03
Fixing #issue1630

### LINKED ISSUE #1630: CronJob version in set to v1 instead of batch/v1beta1
When creating a cronJob file in src/main/fabric8, the generated file uses version `v1` instead of `batch/v1beta1`: 

```
---
apiVersion: v1
kind: CronJob
metadata:
  labels:
    app: fs-connector-a3i
    provider: fabric8
    version: 1.3.0.BUILD-SNAPSHOT
    group: com.airbusds.geo.agri.fs
  name: synchronize-results
spec:
  jobTemplate:
...
```

This is due to the class KubernetesResourceUtil that does not manage cronjob files, see https://github.com/fabric8io/fabric8-maven-plugin/blob/master/core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java#L254

```
if (Objects.equals(kind, "Ingress")) {
  apiVersion = apiVersions.getExtensionsVersion();
} else if (Objects.equals(kind, "StatefulSet") || Objects.equals(kind, "Deployment")) {
  apiVersion = apiVersions.getAppsVersion();
 } else if (Objects.equals(kind, "Job")) {
  apiVersion = apiVersions.getJobVersion();
} else if (Objects.equals(kind, "DeploymentConfig") && platformMode == PlatformMode.openshift) {
  apiVersion = apiVersions.getOpenshiftV1version();
}
```

I suggest adding a new use case:
```
} else if (Objects.equals(kind, "CronJob"){
  apiVersion = apiVersions.getCronJobVersion();
}
```
And adding a new property for CronJob in ResourceVersioning with value "ResourceVersioning".

theexplorist:
Fixes #1630 
Fixing CronJob version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix #1628: Adding back support for templates in kubernetes mode to HelmMojo; Fix for NPE in UndeployMojo
 * Fix 1591: Add support for custom resources creation via resource fragments
 * Fix 1636: Added support to choose configMap name
+* Fix 1630: Fix CronJob Version
 
 ### 4.1.0 (16-02-2019)
 * Fix 1578: Fmp is generating ImageChange triggers in the DeploymentConfig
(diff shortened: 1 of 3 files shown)


theexplorist:
> Looks good, Could you please make sure it's working by testing it locally?

Ya sure!

Issue #1478, opened 2019-01-09
Add Java 11 support


rhuss:

rohanKanojia:
Duplicate of #1472

rohanKanojia:
I think we should prioritize this @lordofthejars @dev-gaur WDYT??

dev-gaur:
Closing as fixed.

PR #1651, opened 2019-06-12
Java 11 compliance

dev-gaur:
#1478


.circleci/config.yml
```diff
@@ -42,6 +42,20 @@ jobs:
         paths:
         - ~/.m2
 
+  build-java11:
```
rohanKanojia: You need to add these jobs in workflows.all.jobs array also

.circleci/config.yml
```diff
@@ -59,6 +73,23 @@ jobs:
         paths:
         - ~/.m2
 
+  # Run Tests
+  tests-java11:
+    working_directory: ~/fabric8io/fabric8-maven-plugin
+    machine: true
```
rohanKanojia: How are we ensuring that we're building on Java 11? I don't see any jdk11 image being used.

dev-gaur: The -Pjava11 flag, this profile has configured maven-compiler-plugin to use java 11.

parent/pom.xml
```diff
@@ -934,6 +938,53 @@
       </dependencies>
     </profile>
 
+    <profile>
```
rohanKanojia: What difference does this profile make? I compiled project on jdk11; apart from `ThorntailV2HealthCheckEnricherTest.java` project build fine. Perhaps I missed something?

lordofthejars: J11 removes some API from J10 (ie XML) so you need to add them manually. I cannot see the file but I guess this is what you mean in your comment.

PR #1655, opened 2019-06-18
Openshift serviceaccount requires OAuth authentication

dantalian-pv:
When username/password is used for `fabric8:apply` or `fabric8:deploy` to apply DeploymentConfig to OpenShift cluster it fails for serviceaccount, because serviceaccount can only use OAth token. Here is a patch which adds `token` entry into `access` configuration to make authentication on OpenShift cluster possible.

Signed-off-by: dantalain-pv <dantalian.pv@gmail.com>


dantalian-pv:
Sure, done

PR #1660, opened 2019-06-24
Chore: Minor refactor to avoid code duplication.

rohanKanojia:
No need to specify specific methods for every flag we add to plugin.

### DIFF
--- a/enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
+++ b/enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
@@ -43,11 +43,17 @@ import java.util.Properties;
  */
 public class BaseEnricher implements Enricher {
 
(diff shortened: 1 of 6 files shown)


enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -228,4 +190,16 @@ protected void setProcessingInstruction(String key, List<String> containerNames)
         processingInstructionsMap.put(key, String.join(",", containerNames));
         enricherContext.setProcessingInstructions(processingInstructionsMap);
     }
+
+    protected Boolean getBooleanValueFromConfig(String propertyName, Boolean defaultValue) {
```
dev-gaur: @rohanKanojia why not let the name be `getValueFromConfig` and put function overloading to use ? @lordofthejars thoughts?

rohanKanojia: Nice catch! I would update this PR

PR #1667, opened 2019-07-11
Fix 1666: corrected back switchToDeployment flag functionality.

### LINKED ISSUE #1666: switchToDeployment does not work
### Info

* f-m-p version :4.1.0
* Maven version (`mvn -v`) :3.6.1
```
spring boot:2.0.6
label<switchToDeployment> It didn't work，The generated file contains parameters openshift.yaml “kind:DeploymentConfig”
```
* Kubernetes / OpenShift setup and version :v1.10/v3.10

dev-gaur:
fixes #1666

### DIFF
--- a/enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
+++ b/enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
@@ -47,13 +47,14 @@ public class BaseEnricher implements Enricher {
     private final EnricherConfig config;
     private final String name;
     protected EnricherContext enricherContext;
-    public static final String FABRIC8_GENERATED_CONTAINERS = "FABRIC8_GENERATED_CONTAINERS";
-    public static final String NEED_IMAGECHANGE_TRIGGERS = "IMAGECHANGE_TRIGGER";
-    public static final String IMAGE_CHANGE_TRIGGERS = "fabric8.openshift.imageChangeTriggers";
-    public static final String OPENSHIFT_TRIM_IMAGE_IN_CONTAINER_SPEC = "fabric8.openshift.trimImageInContainerSpec";
-    public static final String OPENSHIFT_ENABLE_AUTOMATIC_TRIGGER = "fabric8.openshift.enableAutomaticTrigger";
-    public static final String SIDECAR = "fabric8.sidecar";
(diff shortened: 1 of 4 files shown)


enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/openshift/DeploymentConfigEnricher.java
```diff
@@ -52,10 +53,17 @@ public DeploymentConfigEnricher(MavenEnricherContext context) {
 
     @Override
     public void create(PlatformMode platformMode, KubernetesListBuilder builder) {
+
         if(platformMode == PlatformMode.openshift) {
-            for(HasMetadata item : builder.buildItems()) {
 
-                if(item instanceof Deployment) {
+            boolean switchToDeployment = false;
```
rohanKanojia: I would extract this to a method in BaseEnricher and a static constant for flag name. Since there is already `getValueFromConfig()` method available; this can make your work a bit easier

enricher/api/src/main/java/io/fabric8/maven/enricher/api/BaseEnricher.java
```diff
@@ -59,6 +60,12 @@ public BaseEnricher(EnricherContext enricherContext, String name) {
         this.name = name;
     }
 
+    protected boolean useDeploymentforOpenShift() {
```
rohanKanojia: Do we really need this method? Can't we simply use getValueFromConfig directly?

dev-gaur: I checked for getValueFromConfig method. That method is specific is not a general utility, but a specific method used in VertxHealthCheckEnricher.

rohanKanojia: what? No. I think I wrote it to be used by all enrichers

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/openshift/DeploymentConfigEnricher.java
dev-gaur: Aah.. I made this changes in a bit of rush. Forgot about BaseEnricher and static constant.

rohanKanojia: np

PR #1673, opened 2019-07-18
Got rid of ExposeEnricher

### LINKED ISSUE #1374: Rethink Route/Ingress generation
### Description

Currently the configuration and generation of a Route/Ingress via automatic enrichment is a bit interwoven. For the target goal, to move all central logic into Maven independent parts (like enrichers), an option like `fabric8.openshift.generateRoute=false` which just filters out any routes during writing of the resource files, is problematic as a user might explicit´ly configure a route e.g. via a fragment might lead to conflicting situations.

Actually, every postprocessing of the generated resource files (like it happens also for converting Kubernetes to OpenShift objects) conflicts the way how the enrichment process is intended.

Ideally everything happens in the enrichers. So a good solution would be to move everything into `RouteIngressEnricher` and just disable this enricher if you don't want a route or ingress (which can be easily done with a profile, too). 

Another issue with the current RouteEnricher is that it depends on the ExposeEnricher that a certain label gets added to a service which is picked up by the route enricher. Instead, the RouteEnricher should examine itself all service and should be able to configure to select the service to expose. This is much easier to understand, and makes the ExposeEnricher obsolete.

dev-gaur:
fixes #1374


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ We use semantic versioning in some slight variation until our feature set has st
 * The `PATCH_LEVEL` is used for regular CD releases which add new features and bug fixes.
 
 After this we will switch probably to real [Semantic Versioning 2.0.0](http://semver.org/)
+### 4.2-SNAPSHOT
```
rohanKanojia: Rename `4.1-SNAPSHOT` and move items there

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/IngressEnricher.java
```diff
@@ -30,12 +30,12 @@
...
 import io.fabric8.maven.core.util.kubernetes.Fabric8Annotations;
+import io.fabric8.maven.core.util.kubernetes.KubernetesResourceUtil;
 import io.fabric8.maven.enricher.api.BaseEnricher;
+import io.fabric8.maven.enricher.api.Kind;
 import io.fabric8.maven.enricher.api.MavenEnricherContext;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
rohanKanojia: nit, You introduced wildcard imports here

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/openshift/RouteEnricher.java
```diff
@@ -68,9 +63,9 @@ public void visit(ServiceBuilder serviceBuilder) {
 
     private void addRoute(KubernetesListBuilder listBuilder, ServiceBuilder serviceBuilder, List<Route> routes) {
         ObjectMeta metadata = serviceBuilder.getMetadata();
-        if (metadata != null && isExposedService(serviceBuilder)) {
+        if (metadata != null) {
             String name = metadata.getName();
-            if (!hasRoute(listBuilder, name)) {
+            //if (!hasRoute(listBuilder, name)) {
```
rohanKanojia: remove this if not needed

enricher/standard/src/main/java/io/fabric8/maven/enricher/standard/openshift/RouteEnricher.java
```diff
@@ -105,6 +100,18 @@ private RoutePort createRoutePort(ServiceBuilder serviceBuilder) {
         return routePort;
     }
 
+    private boolean hasWebPort(List<ServicePort> ports) {
```
rohanKanojia: Why are we checking only a specific set of ports of web ports? What if user has supplied some other port rather than these ports? if he has, can we add that check here as well along with standard ports?

dev-gaur: I have followed what the expose enricher used to do.
Also, the purpose of a Route is basically to service incoming traffic..which comes through http and https only. So this set of ports looks legitimate to me.

 @lordofthejars thoughts ?

rohanKanojia: I'm okay with these ports, but I'm just concerned about the case when the port is not among one of these ports

dev-gaur: Then that simply violates http and https and it should not be accepted, imo

Issue #1678, opened 2019-07-23
fabric8:build crashes on xml-config sample


dev-gaur:
* f-m-p version : 4.2-SNAPSHOT
* Sample project : https://github.com/fabric8io/fabric8-maven-plugin/tree/master/samples/xml-config

rohanKanojia:
what error are you getting?

theexplorist:
There was error in <labels> for Pnew, now for Pexternal why we are using digest in from image, it throws image not found and it fails. So if you say I will correct that as well
 @dev-gaur @rohanKanojia

rohanKanojia:
@theexplorist : Is this issue fixed now? Or is there still something left to fix?

theexplorist:
It's fixed

PR #1679, opened 2019-07-24
Correction in labels in -Pnew configuration  

theexplorist:


dev-gaur:
fixes #1678

samples/xml-config/pom.xml
```diff
@@ -241,7 +243,7 @@
 		  <name>${image.user}/${project.artifactId}:${project.version}</name>
 		  <alias>camel-service</alias>
 		  <build>
-		    <from>fabric8/s2i-java@@sha256:a6ba8a6e76ebd75f99f2431fe0ec39a82985bc21c0f9bd56021deca1ffde520c</from>
+		    <from>fabric8/s2i-java</from>
```
rohanKanojia: Just need to remove one `@` here,

samples/xml-config/pom.xml
```diff
@@ -211,6 +212,7 @@
                   <volumes>
                   </volumes>
                 </deployment>
+                </labels>
```
rohanKanojia: wrong place to close label

samples/xml-config/pom.xml
```diff
@@ -182,8 +182,9 @@
               <resources>
                 <!-- Labels that are applied to all created objects -->
                 <labels>
-                  <group>quickstarts</group>
-                </labels>
+                    <all>
+                        <group>quickstarts</group>
+                    </all>
```
rohanKanojia: add `</label>` to close labels field here

theexplorist: deployments is one of the labels in metadataconfig. On the contrary there is no field deployment in ResourceConfig hence there build fails. And I have tested the build, now it passes.

rohanKanojia: oh, okay. Have you fixed this in external profile as well?

samples/xml-config/pom.xml
theexplorist: Oh there are 2 @s I had not took notice!

samples/xml-config/pom.xml
theexplorist: Yes everything is correct there, adding the digest again!

theexplorist: Fixed, now i guess it is good to merge @rohanKanojia

PR #1703, opened 2019-09-11
Fix #1699: Ability to specify object namespace in fragments

### LINKED ISSUE #1699: Ability to specify object namespace in fragments
As a developer, I want the ability to deploy different object into different namespaces. 
That means the plugin should apply objects according to namespace specified in fragments.

Currently what happens if fragment contains a namespace different than the specified working namespace is this error

```
Failed to execute goal io.fabric8:fabric8-maven-plugin:4.3-SNAPSHOT:resource-apply (default-cli) on project test-openshift: Failed to update ConfigMap from openshift.yml. io.fabric8.kubernetes.client.KubernetesClientException: Namespace mismatch. Item namespace:debezium-mysql. Operation namespace:debezium.. ConfigMap(apiVersion=v1, binaryData=null, data={log4j.properties=kafka.logs.dir=logs
```

rohanKanojia:
Fix #1699 

namespace inside resource fragments should be given more priority over
default namespace


rohanKanojia:
I've tested this on this sample project : https://github.com/r0haaaan/fabric8-maven-plugin-sample-with-crds

PR #1716, opened 2019-10-01
Fix #1714: mvn fabric8:resource converts types of kubernetes YAML annotations in fragment

### LINKED ISSUE #1714: mvn fabric8:resource converts types of kubernetes YAML annotations in fragment
### Info

* f-m-p version : 4.2.0
* Maven version (`mvn -v`) : 3.6.2
```
metadata:
  annotations:
    # Enable client certificate authentication
    nginx.ingress.kubernetes.io/auth-tls-verify-client: 'on'

```

After doing whatever it does to an ingress resource fragment (taken from the nginx ingress controller examples), the f-m-p will emit:

```
metadata:
  annotations:
    nginx.ingress.kubernetes.io/auth-tls-verify-client: on

```

Note the lack of quotes. This results in the warning:

`[message=.metadata.annotations.nginx.ingress.kubernetes.io/auth-tls-verify-client: boolean found, string expected, violation type=type]`

The difference matters quite a bit as the wrong type will stop the ingress controller from actually enforcing client certificate auth...

rohanKanojia:
Fixes #1714 

Modified Jackson ObjectMapper to disable minimizing quotes in case of yamls

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix #1699: Ability to specify object namespace in fragments 
 * Feature #1536: Java Image Builder Support
 * Fix #1704: fabric8-build failing on openshift
+* Fix #1714: resource-goal converts types of kubernetes YAML annotations in fragment
 * Feature #1706: Prometheus Enricher; Configuration support for Prometheus path
 * Added generator support for Open Liberty
 
--- a/core/src/main/java/io/fabric8/maven/core/util/ResourceUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/ResourceUtil.java
@@ -23,6 +23,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import com.google.gson.JsonObject;
 import org.apache.commons.io.FilenameUtils;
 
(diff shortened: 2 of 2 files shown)


rohanKanojia:
tested on https://github.com/r0haaaan/fmp-sample-external-resources with and without this patch.

erikgb:
@rohanKanojia : I can confirm that the PR actually fixes my conrete problem as reported in #1714. Well done and thanks!

erikgb:
@rohanKanojia : I am checking out the latest version, with this fix (4.3.0). The fix seems to have made the generated resources less readable, specifically when creating ConfigMap resources from file.

Given:
````yaml
metadata:
  annotations:
    maven.fabric8.io/cm/declarative.yaml: kong-config.yaml
````
````yaml
_format_version: "1.1"

services:
- name: my-service
  url: https://example.com
  plugins:
  - name: key-auth
  routes:
  - name: my-route
    paths:
    - /

consumers:
- username: my-user
  keyauth_credentials:
  - key: my-key

````

f-m-p 4.3.0:
````yaml
data:
  declarative.yaml: "_format_version: \"1.1\"\n\nservices:\n- name: my-service\n \
    \ url: https://example.com\n  plugins:\n  - name: key-auth\n  routes:\n  - name:\
    \ my-route\n    paths:\n    - /\n\nconsumers:\n- username: my-user\n  keyauth_credentials:\n\
    \  - key: my-key\n"

````
f-m-p 4.2.0:
````yaml
data:
  declarative.yaml: |
    _format_version: "1.1"

    services:
    - name: my-service
      url: https://example.com
      plugins:
      - name: key-auth
      routes:
      - name: my-route
        paths:
        - /

    consumers:
    - username: my-user
      keyauth_credentials:
      - key: my-key
````

It looks ok again in the cluster console, but I must admit that I would have preferred to have the generated resource more readable.
What do you think? Is there a (quick) fix?

rohanKanojia:
@erikgb : Not sure. Actually we were minimising quote in jackson while converting to yaml. For your env fix, I had to disable that:
https://github.com/fabric8io/fabric8-maven-plugin/blob/fa070a9fe6b6c4939ba7fc6e697429917821e19b/core/src/main/java/io/fabric8/maven/core/util/ResourceUtil.java#L90-L100

erikgb:
@rohanKanojia : Ok, I can live it. At least for now - lesser evil. ;-) I can try to dig in to the Jackson settings and experiment a bit....

rohanKanojia:
I also noticed that yamls are less readable with this, but I thought maybe we need to compromise over readability for fixing that bug. But it would be awesome if we can find some way to have both readability and selective quote addition(not sure about this though!)

rohanKanojia:
@erikgb : Anyway, I'm done for the week. I would try to look into this next week. Have a nice weekend!

erikgb:
@rohanKanojia : Same to you! Have a nive weekend!

PR #1717, opened 2019-10-01
Fix #1715: ApplyService#applyProjectRequest should be (truly) idempotent

### LINKED ISSUE #1715: ApplyService#applyProjectRequest should be (truly) idempotent
### Description

It seems like the method [ApplyService#applyProjectRequest](https://github.com/fabric8io/fabric8-maven-plugin/blob/721c960ffaf8e56fd713e170594f9b9ea352bed7/core/src/main/java/io/fabric8/maven/core/service/ApplyService.java#L1073) isn't truly idempotent - and I think the method name applies that it should be. It is close, but no cigar. If the method is invoked when a POST ProjectRequest is already in process by Openshift, but **before** the Project/Namespace actually exists, it will fail - because Openshift returns HTTP 409 when trying to POST ProjectRequest.

Somehow this bug is only noticeable when you are creating RoleBinding and BuildConfig resources. Why this happens becomes apparent if you look at the callers to the method:
````
ApplyService.applyProject(Project)  (io.fabric8.maven.core.service)
    ApplyMojo.executeInternal()  (io.fabric8.maven.plugin.mojo.build)
ApplyService.applyNamespace(String, Map<String, String>)  (io.fabric8.maven.core.service)
    ApplyService.applyNamespace(String)  (io.fabric8.maven.core.service)
        ApplyMojo.executeInternal()  (io.fabric8.maven.plugin.mojo.build)
        ApplyService.getNamespace(HasMetadata)  (io.fabric8.maven.core.service)
        ApplyService.applyRoleBinding(RoleBinding, String)  (io.fabric8.maven.core.service)
        ApplyService.applyBuildConfig(BuildConfig, String)  (io.fabric8.maven.core.service)
````
### Info

* f-m-p version : 4.2.0 (and same behavior with HEAD of master; 4.3-SNAPSHOT)
* Maven version (`mvn -v`) : 3.5.3 and also 3.6.1 (but I think it is irrelevant)
* Kubernetes / OpenShift setup and version : Openshift 3.11

### How to reproduce

* Prerequisite: Openshift mode

1. Create a simple Maven project with a single RoleBinding resource
1. Run mvn fabric8:resource-apply

### PR DESCRIPTION
**Pull request by rohanKanojia**
Fix #1715

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/service/ApplyService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/ApplyService.java
@@ -70,9 +70,11 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 import static io.fabric8.maven.core.util.kubernetes.KubernetesHelper.getKind;
 import static io.fabric8.maven.core.util.kubernetes.KubernetesHelper.getName;
@@ -102,6 +104,8 @@ public class ApplyService {
     private boolean rollingUpgradePreserveScale = true;
     private boolean recreateMode;
     private PatchService patchService;
+    // This map is to track projects created.
+    private static Set<String> projectsCreated = new HashSet<>();
 
     public ApplyService(KubernetesClient kubernetesClient, Logger log) {
         this.kubernetesClient = kubernetesClient;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/src/main/java/io/fabric8/maven/core/service/ApplyService.java [hunk 1]
```diff
@@ -1071,6 +1075,9 @@ public boolean applyProject(Project project) {
      * Returns true if the ProjectRequest is created
      */
     public boolean applyProjectRequest(ProjectRequest entity) {
+        // Check whether project creation attempted before
+        if (projectsCreated.contains(getName(entity)))
+            return true;
```
erikgb: According to the method signature, I think the return value should be false if projectsCreated contains getName(entity)

rohanKanojia: Nice catch! Thanks :+1:

core/src/main/java/io/fabric8/maven/core/service/ApplyService.java
```diff
@@ -1071,6 +1075,9 @@ public boolean applyProject(Project project) {
      * Returns true if the ProjectRequest is created
      */
     public boolean applyProjectRequest(ProjectRequest entity) {
+        // Check whether project creation attempted before
+        if (projectsCreated.contains(getName(entity)))
```
dev-gaur: @rohanKanojia curly brackets, please!

PR #1723, opened 2019-10-04
Resolve #1718: Support for binary files in fmp-configmap-file

### LINKED ISSUE #1718: Support for binary files in fmp-configmap-file
ConfigMapEnricher (fmp-configmap-file) was a very nice addition to f-m-p, and we are using it already.

But when I try to use it for files with binary content, i.e. a Java truststore or a (binary) license file, this does not work.

In order to make ConfigMapEnricher work for binary data, the ConfigMapEnricher has to store the file contents in the binaryData field - instead of the standard data field. And this should ideally be done transparently - with the same logic as oc/kubectl uses when creating ConfigMaps with the '--from-file' option: non-UTF8 data should use the binaryData field, otherwise use the data field (as now).

I had a look at the source code, and I think this can be fixed without too much effort. I could prepare a PR?

erikgb:


erikgb:
@rohanKanojia : Here are my suggested code changes. I will append an entry to the changelog and add some documentation, but it would be nice if you could start reviewing the code changes. I had to do more changes than I expected.... And the XML-configuration didn't have any tests at all! Tests could always be improved, but I do not have time to catch up the backlog on this now. Of course I added tests for new feature.

erikgb:
Another thing: I which Kubernetes/Openshift version was the binaryData field introduced for ConfigMap resources? I just had a look at the [compatability matrix](https://github.com/fabric8io/fabric8-maven-plugin), and I do not want to impact compatabilities with this PR....

rohanKanojia:
I think they were inroduced in 1.14.2

rohanKanojia:
Actually for configmap we had an integration test, there wasn't any unit test for that. Thanks for adding unit tests too

rohanKanojia:
@erikgb : I'm going to merge this and release 4.3.0

erikgb:
I just added a line to the changelog. Do you want me to add something about this in the documentation? I gave it a try, but lost inspiration. ;-) As a user, and noob on Openshift/Kubernetes, I would expect f-m-p to handle binary files correctly. So I think no additional documentation is required in this case.

rohanKanojia:
Yeah, documentation only says how to add configmap. Ideally it should work for both kinds of configmaps

PR #1725, opened 2019-10-05
Upgrade to latest version of Jackson (2.10.0). Better fix for #1714

### LINKED ISSUE #1714: mvn fabric8:resource converts types of kubernetes YAML annotations in fragment

erikgb:
https://github.com/fabric8io/fabric8-maven-plugin/pull/1716 made the generated YAML files a lot less readable, but fixed #1714. I did some investigation, and it seems like our issue is caused by a bug in Jackson. The bug was still present in version 2.9.10, but fixed in 2.10.0.

This PR contains the following:
1. Revert the temporary fix for #1714 prepared by @rohanKanojia.
1. Introduce Jackson BOM to get dependency management on Jackson dependencies
1. Upgrade Jackson to the latest release (2.10.0)
1. Add dependency management for snakeyaml. Jackson requires a newer version than the default managed version by Maven (because of docker-maven-plugin).

There are several ways to fix the last issue/change. Please advice if you want me to do something else. I.e. exclude the transitive dependency pulled in by the dependency to docker-maven-plugin.

### DIFF
--- a/core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
@@ -387,7 +387,7 @@ public class KubernetesResourceUtil {
             Map<String, Object> ret = mapper.readValue(file, typeRef);
             return ret != null ? ret : new HashMap<String, Object>();
         } catch (JsonProcessingException e) {
-            throw new JsonMappingException(String.format("[%s] %s", file, e.getMessage()), e.getLocation(), e);
+            throw new JsonProcessingException(String.format("[%s] %s", file, e.getMessage()), e.getLocation(), e) {};
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,6 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix #1699: Ability to specify object namespace in fragments 
 * Feature #1536: Java Image Builder Support
 * Fix #1704: fabric8-build failing on openshift
-* Fix #1714: resource-goal converts types of kubernetes YAML annotations in fragment
 * Feature #1706: Prometheus Enricher; Configuration support for Prometheus path
 * Fix #1715: ApplyService#applyProjectRequest should be (truly) idempotent
 * Added generator support for Open Liberty
--- a/core/src/main/java/io/fabric8/maven/core/util/ResourceUtil.java
+++ b/core/src/main/java/io/fabric8/maven/core/util/ResourceUtil.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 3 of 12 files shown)


erikgb:
While I was in "Jackson land", also fixed (most of) the Jackson related deprecation warnings in the code - in a separate commit. The last remaining deprecation warning (SerializationFeature.WRITE_EMPTY_JSON_ARRAYS) doesn't have a good replacement in our case, but we are luckily not alone. See https://github.com/FasterXML/jackson-databind/issues/1547 for additional details.

erikgb:
@rohanKanojia : Can you please advice me on how to fix the "broken" CI license build? I've tried a couple of different things, but no luck so far. In order to add a test for invalid content (giving JsonProcessingException) in resource fragments, I need a file matching the resource fragment name pattern **and** containing some non UTF-8 characters.

core/src/main/java/io/fabric8/maven/core/util/kubernetes/KubernetesResourceUtil.java
```diff
@@ -387,7 +387,7 @@ private static void addIfNotExistent(Map<String, Object> fragment, String key, S
             Map<String, Object> ret = mapper.readValue(file, typeRef);
             return ret != null ? ret : new HashMap<String, Object>();
         } catch (JsonProcessingException e) {
-            throw new JsonMappingException(String.format("[%s] %s", file, e.getMessage()), e.getLocation(), e);
+            throw new JsonProcessingException(String.format("[%s] %s", file, e.getMessage()), e.getLocation(), e) {};
```
dev-gaur: just curious. how do curly braces help here? I see that you're trying to invoke a protected constructor. Does the curly brace turn it into an anonymous class object extending the `JsonProcessingException` (just a wild guess!). I'm learning Java :)

erikgb: Yes, the curely braces creates a anonymous class extending `JsonProcessingException`. I chose this solution for two reasons:
1. The deprecated constructor on `JsonMappingException` suggests a replacement using a factory with `JsonParser` as one of it's parameters. I didn't figure out how to get access to it from the `ObjectMapper`.
1. The purpose of this catch block is just to enrich the error message presented to the user with the source file information. So I thought it was best not to transform a generic exception (`JsonProcessingException`) into a more specific (`JsonMappingException`). It could end up being wrong/misleading.

erikgb: A protected constructor is not nessacarily considered part of an API, and I think Jackson fairly can make this constructor package-protected (or even remove it) at some point - without a new major version, ref. SemVer. I tend to use just plain `RuntimeException` to wrap exceptions like this one. I am not a big fan of checked exceptions in general....

dev-gaur:
@erikgb An integration test can help here

erikgb:
@dev-gaur Agree, do I have to add a new integration test, or do you think I can extend an existing one? Please advise.

erikgb:
I've added an integration test to verify that environment variables with YAML boolean values are working as expceted.

PR #1727, opened 2019-10-09
fabric8

cmadunezim:
master/compare:3.5x


rohanKanojia:
@cmadunezim : Could you please elaborate the use case you're trying to solve with this PR?

PR #1733, opened 2019-10-14
#1732 - Resolving template placeholders only in kubernetes.yaml but not individual template files (which will be resolved by helm mojo)

### LINKED ISSUE #1732: Helm Mojo not replacing template placeholders with corresponding `.Values.xxx` placeholders
# Description
Helm Mojo not replacing template placeholders with corresponding `.Values.xxx` placeholders.
Instead deployment.yml contains resolved strings from template.yaml.

* f-m-p version : 4.3.0

djotanov:
Fix #1732

### DIFF
--- a/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
+++ b/plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
@@ -656,13 +656,6 @@ public class ResourceMojo extends AbstractFabric8Mojo {
             }
             File kubernetesYaml = new File(this.targetDir, "kubernetes.yml");
             resolveTemplateVariables(parameters, kubernetesYaml);
-            File kubernetesResourceDir = new File(this.targetDir, "kubernetes");
-            File[] kubernetesResources = kubernetesResourceDir.listFiles((dir, filename) -> filename.endsWith(".yml"));
-            if (kubernetesResources != null) {
-                for (File kubernetesResource : kubernetesResources) {
-                    resolveTemplateVariables(parameters, kubernetesResource);
-                }
-            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/io/fabric8/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -656,13 +656,6 @@ private void resolveTemplateVariablesIfAny(KubernetesList resources) throws Mojo
             }
             File kubernetesYaml = new File(this.targetDir, "kubernetes.yml");
             resolveTemplateVariables(parameters, kubernetesYaml);
-            File kubernetesResourceDir = new File(this.targetDir, "kubernetes");
```
rohanKanojia: Could you please elaborate on how this fixes #1732

rohanKanojia: oh, so we were resolving placeholders in template files also.

djotanov: Those lines were replacing template placeholders in kubernetes resource files with concrete values specified in template.yaml file. The placeholders need to be resolved only in kubernetes.yml in order to be able to do fabric8:deploy or fabric8:apply, but they don't need to be resolved in individual kubernetes resource files as they will be needed in HelmMojo execution.
The HelmMojo is using those same kubernetes resources to create chart, but since the placeholders were already replaced with concrete values from template.yaml, it can not replace them with corresponding `.Values.xxx` placeholders.

djotanov: Easiest is to build spring-boot-with-yaml sample on this branch with fabric:helm goal and inspect generated helm chart

PR #1740, opened 2019-10-17
Fix #1738: JibBuildConfiguration ignores BuildImageConfiguration.registry

### LINKED ISSUE #1738: JibBuildConfiguration ignores BuildImageConfiguration.registry
### Description
JibBuildConfiguration MUST take into account all relevant properties from BuildImageConfiguration.

We do not have access to Docker Hub (registry-1.docker.io) in our CI pipelines, but have to go through a corporate Docker image "proxy". Normally this should be doable with the correct registry configuration in f-m-p. But currently this is not possible for a Jib build because the JibBuildConfiguration created from BuildImageConfiguration seems to ignore a lot of the image configuration in this method: https://github.com/fabric8io/fabric8-maven-plugin/blob/f75d09b1466c5e005a455b381151bdab7b2398eb/core/src/main/java/io/fabric8/maven/core/util/JibBuildServiceUtil.java#L218

This is a blocker that needs to be fixed.

@rohanKanojia @dev-gaur A general (critical) question: Why are you accepting pull request with no, or very little test coverage? In the long run this will definitely discourage developers from using f-m-p. :smirk:
I feel a bit sad now, as I was looking forward to use the new Java Image Builder support in f-m-p, just realizing that it seems to be full of basic bugs....

rohanKanojia:
Fix #1738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ After this we will switch probably to real [Semantic Versioning 2.0.0](http://se
 * Fix #1732: Correctly replacing template placeholders in helm chart
 * Fix #1734: Add logging during jib builds
 * Fix #1696: Fix put operations for ImageStreams in ApplyService
+* Fix #1738: JibBuildConfiguration ignores BuildImageConfiguration.registry
+* Fix #1737: Wrong Jib output directory in Maven multi-module build
 
 ### 4.3.0 (04-10-2019)
 * Updated custom-enricher sample
--- a/core/src/main/java/io/fabric8/maven/core/service/kubernetes/JibBuildService.java
+++ b/core/src/main/java/io/fabric8/maven/core/service/kubernetes/JibBuildService.java
@@ -16,19 +16,20 @@
 package io.fabric8.maven.core.service.kubernetes;
 
 import com.google.cloud.tools.jib.api.Credential;
+import com.google.cloud.tools.jib.api.ImageFormat;
+import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
+import com.google.cloud.tools.jib.api.JibContainer;
+import com.google.cloud.tools.jib.api.RegistryException;
 import io.fabric8.maven.core.service.BuildService;
 import io.fabric8.maven.core.util.JibBuildServiceUtil;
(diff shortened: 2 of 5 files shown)


rohanKanojia:
@erikgb : Could you please try it out and see if it's working for you use case ?

erikgb:
@rohanKanojia : This PR does not seem to fix my use case, but it is very hard to verify it 100%. To "integration test" it, I need to push/deploy it to our corporate Maven repository.... But I know how to (Java) debug, and it seems like the registry handling is improved - compared to HEAD of master. But still not bug free. The problem is that there are so many ways to configure registry handling in f-m-p, and when transforming to Jib "dialect" you need to care for all of them. 😄 

What I am trying to do at the moment is quite simple. I set the Maven property fabric8.generator.registry to our corporate Docker registry proxy. The expectation is that the base image, fabric8/java-centos-openjdk11-jdk:1.6.3 in my case, is pulled from this proxy Docker registry and not registry-1.docker.io (which is blocked). This does not currently seem to happen:
```
baseImageConfiguration = {ImageConfiguration@9222} 
 image = {ImageReference@9229} "fabric8/java-centos-openjdk11-jdk:1.6.3"
  registry = "registry-1.docker.io"
  repository = "fabric8/java-centos-openjdk11-jdk"
  tag = "1.6.3"
 credentialRetrievers = {RegularImmutableList@9224}  size = 0
```
**I think we need tests to do changes to this functionality**. It seems like very basic functionality might work, but if you try to use more advanced functionality in f-m-p it just doesn't work. I will try to find some temporary workaround for my issues for now. Probably by avoiding use of the registry handling in f-m-p and only use absolute Docker image references....

So my suggestion is to postpone doing any more related to this, and release f-m-p 4.3.1. I will have more time next week, and can try to do some work on the it - at least add some tests. I am not comfortable at all doing changes without tests....

rohanKanojia:
@erikgb : Could you please give it a try again? I've added code to handle generator parameters as well.

rohanKanojia:
@erikgb : I'm merging this and releasing 4.3.1 .

erikgb:
@rohanKanojia : Sorry for the late response. I was off for the weekend. I will check if version 4.3.1 works in my use case and let you know....

PR #1748, opened 2019-10-26
QuarkusGenerator: Extract base image from config or fallback to default.

juazugas:
Allow specifying in the configuration for the Quarkus generator the base image of the container. 

This change extracts the configured key `from` for the generator or reads the property `fabric8.generator.from` to use it as a base image for the container.

It also fallbacks to the default values in case any is provided.


rohanKanojia:
@juazugas : Thanks a lot for this PR :+1:  Could you please run a `mvn -N license:format` to fix license issues. Also, please add a line to CHANGELOG regarding this change.

PR #1774, opened 2020-01-09
Chore: Update Fabric8 Kubernetes Client 4.7.0

rohanKanojia:


core/src/test/java/io/fabric8/maven/core/service/openshift/PatchServiceTest.java
```diff
@@ -74,16 +75,18 @@ public void testResourcePatching() {
     }
 
     @Test
+    @Ignore
```
manusa: Why?

rohanKanojia: ah, sorry. I forgot to remove this. I remember I had fixed this test.

enricher/api/src/main/java/io/fabric8/maven/enricher/api/visitor/MetadataVisitor.java
```diff
@@ -166,7 +167,20 @@ protected Kind getKind() {
 
         @Override
         protected ObjectMeta getOrCreateMetadata(PodTemplateSpecBuilder item) {
-            return item.hasMetadata() ? item.buildMetadata() : item.withNewMetadata().endMetadata().buildMetadata();
+            if (Boolean.TRUE.equals(item.hasMetadata())) {
+                if (item.buildMetadata().getLabels() == null) {
+                    item.editMetadata().withLabels(Collections.EMPTY_MAP).endMetadata();
```
manusa: Please, use `Collections.emptyMap()` (returns a generic non-raw type) instead of `Collections.EMPTY_MAP`

manusa: There are too many duplicate lines in the class, find a proposal to refactor here:

https://github.com/rohanKanojia/fabric8-maven-plugin/pull/4/

